Issue #269, opened 2026-04-15
E2E smoke test failures on v9.22.0


nyldn:
## E2E Smoke Test Failure Report

**Version:** v9.22.0
**Date:** 2026-04-15T06:06:58Z
**Claude Code:** 2.1.97 (Claude Code)

### Results
| Phase | Passed | Failed |
|-------|--------|--------|
| A: Install/Uninstall (Docker) | 14 | 1 |
| B: Commands (Native) | 12 | 4 |
| **Total** | **26** | **5** |

### Failures
```
  ❌ A15 Unit tests — 69/137 passed, 68 failed
  ❌ B2 /octo:setup — Pattern 'octopus|setup|provider|codex|gemini|install' not found in output
  ❌ B7 orchestrate.sh doctor — No output
  ❌ B13 Unit tests — 99/137 passed, 38 failed
  ❌ B17 probe-single — No output
```

<details>
<summary>Full log</summary>

```
═══════════════════════════════════════════════════════════════
Phase A: Install/Uninstall Lifecycle
Claude Code: 2.1.87 (Claude Code)
═══════════════════════════════════════════════════════════════
  ✅ A1 Marketplace add
  ✅ A2 Plugin install
  ✅ A3 Plugin in list
    Installed version: 9.22.0
  ✅ A4 Plugin directory exists
  ✅ A5 orchestrate.sh executable
  ✅ A6 orchestrate.sh syntax valid
  ✅ A7 Commands: 95 (≥40)
  ✅ A8 Skills: 51 (≥45)
  ✅ A9 Personas: 32 (≥30)
  ✅ A10 hooks.json valid JSON (.claude-plugin/)
  ✅ A11 check-providers.sh
  ✅ A12 build-fleet.sh
  ✅ A13 build-fleet.sh output valid (2 agents)
  ✅ A14 All lib/*.sh syntax valid
  ❌ A15 Unit tests — 69/137 passed, 68 failed
═══════════════════════════════════════════════════════════════
Phase B: Live Command Tests
Claude Code: 2.1.97 (Claude Code)
═══════════════════════════════════════════════════════════════
  ✅ B0 Plugin version matches latest (9.22.0)
  ✅ B1 /octo:doctor
  ❌ B2 /octo:setup — Pattern 'octopus|setup|provider|codex|gemini|install' not found in output
    First 3 lines: 
[...]

PR #267, opened 2026-04-14
feat(memory): provider contract — mcp-memory-service alongside claude-mem (#220)

michaelheichler:
## Context

Issue [#220](https://github.com/nyldn/claude-octopus/issues/220) surfaced two asks in its later thread:

1. **@nyldn** (maintainer): *\"detect mcp-memory-service in mcpServers config and query it during Discover phase alongside claude-mem\"* — put on the P2 backlog.
2. **@wazionapps**: *\"normalize the memory **contract**, not just the retrieval call\"* — expose the same primitives for scoped namespaces, freshness metadata, and typed writes so backends can *degrade gracefully instead of all pretending to be the same store.*\"

This PR lands the contract layer and a working second backend. The Discover-phase prompt integration is noted as a clean follow-up.

### `scripts/lib/memory.sh` — single façade
Callers never touch a backend directly. Primitives:

| Function | Purpose |
|---|---|
| `memory_available` | Any reachable backend present? |
| `memory_search <query> [limit] [scope]` | Ordered query, first non-empty wins |
| `memory_observe <type> <title> <text> [scope]` | First-reachable backend only, no double-record |
| `memory_context [scope] [limit]` | Formatted recent-activity summary |
| `memory_backends` | Ordered backend list resolver |
| `memory_scope` | Scope label (git-root basename by default) |

### Backend selection

| Env var | Effect |
|---|---|
| `OCTOPUS_MEMORY_BACKEND=auto` *(default)* | mcp-memory-service first when registered in Claude Code `mcpServers`, else claude-mem |
| `OCTOPUS_MEMORY_BACKEND=claude-mem,mcp-memory-service` | Explicit preference, queried in order |
| `OCTOPUS_MEMORY_SEARCH_MERGE=1` | Aggregate across backends instead of first-win |
| `OCTOPUS_MEMORY_SCOPE=<label>` | Override scope |

Detection matches by **command/package string** in `mcpServers` so the user's key name (`memory`, `mcp-memory-service`, anything) doesn't matter — we look for the `uvx mcp-memory-service` signature.

### `scripts/mcp-memory-bridge.sh` — CLI adapter

Claude's MCP tool layer stays the primary path (that's how the service is designed to run). This bridge is a fallback so the bash side can detect availability and do best-effort reads without a live MCP session.

[...]

PR #272, opened 2026-04-15
test: stop cascading silent aborts (postfix increment + SIGPIPE/pipefail)

michaelheichler:
## Summary

Two latent bugs in the test suite were hidden by #269 — once #270 stopped \`orchestrate.sh\` from silently aborting, these tests started reporting their own assertion failures, and we discovered the assertion code itself was unreliable.

### 1. \`((VAR++))\` exit-1 under \`set -e\`

When \`VAR=0\`, postfix increment evaluates the expression to \`0\` *before* incrementing — bash treats arithmetic exit \`0\` as a failed command, and \`set -e\` aborts the test script mid-run.

Affected files were either silently passing (only counting tests run before the first \`pass()\` call) or silently aborting on Test 1.

Bulk-converted to \`((++VAR))\` (pre-increment returns the new value, never \`0\`).

**Files:** 15 test files across \`tests/\`, \`tests/unit/\`, \`tests/integration/\`.

### 2. SIGPIPE killing pipelines under \`set -o pipefail\`

\`grep -A 100 PATTERN file | grep -q OTHER\` early-exits on first match. Upstream \`grep\` gets SIGPIPE (exit 141). With \`set -o pipefail\`, the pipeline's exit becomes 141, the \`if\` condition reads false, and the assertion is reported as **failed** even when the pattern *was* present.

This only fired on test files where the search target had grown past the pipe-buffer size (~64 KiB) — \`test-model-config-v849.sh\` and \`test-provider-activation.sh\`. Disabled \`pipefail\` in those two files (\`set -uo pipefail; set +o pipefail\`) with a one-line WHY comment.

### 3. test-claude-mem-integration: stale grep pattern

The grep was checking for legacy \`bridge_script.*observe\` / \`claude-mem-bridge.*observe\` patterns. PR #267 routes \`save_session_checkpoint\` through the new \`memory_observe\` contract — extended the regex to match either form.

## Result

**137 / 137 test suites pass** (was 100 / 137 on v9.22.0 main).

## Test plan

- [x] \`bash tests/run-all-tests.sh\` → 137 passed, 0 failed
- [x] Each individually fixed file runs to completion and reports correct counts
- [x] Pre-increment is functionally equivalent for tally counters

## Dependencies

**Depends on #270** (the memory.sh pipefail leak). Without that fix, \`orchestrate.sh\` still aborts and the resulting cascade dwarfs these test-side fixes — CI on this branch alone (against unfixed main) would still fail. Branch is based on \`fix/memory-pipefail-leak\` to make CI representative.

## Summary by CodeRabbit

* **Chores**
  * Standardized counter increment patterns across many test scripts for consistency.
  * Relaxed strict error handling in a helper script to reduce false-positive failures.

* **Tests**
  * Updated test harnesses to delegate bookkeeping and preserve existing outputs while stabilizing evaluation.
  * Hardened pattern checks and result accounting to make tests more robust and reliable.

### DIFF
--- a/scripts/lib/memory.sh
+++ b/scripts/lib/memory.sh
@@ -6,8 +6,6 @@
 #   OCTOPUS_MEMORY_BACKEND  "auto" (default) | comma-separated list
 #   OCTOPUS_MEMORY_SCOPE    override scope label (default: repo basename)
 
-set -euo pipefail
-
 _MEMORY_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 _MEMORY_PLUGIN_DIR="$(cd "${_MEMORY_LIB_DIR}/../.." && pwd)"
 _MEMORY_BRIDGE_DIR="${_MEMORY_PLUGIN_DIR}/scripts"
--- a/tests/integration/test-value-proposition.sh
+++ b/tests/integration/test-value-proposition.sh
@@ -43,15 +43,15 @@ assert_true() {
     local condition="$1"
     local description="$2"
 
-    ((TESTS_RUN++))
+    ((++TESTS_RUN))
 
     if eval "$condition"; then
         echo -e "${GREEN}✓${NC} $description"
-        ((TESTS_PASSED++))
+        ((++TESTS_PASSED))
         return 0
     else
         echo -e "${RED}✗${NC} $description"
-        ((TESTS_FAILED++))
+        ((++TESTS_FAILED))
         return 1
     fi
 }
@@ -60,15 +60,15 @@ assert_file_exists() {
     local file="$1"
     local description="$2"
 
-    ((TESTS_RUN++))
+    ((++TESTS_RUN))
 
     if [[ -f "$file" ]]; then
         echo -e "${GREEN}✓${NC} $description"
-        ((TESTS_PASSED++))
+        ((++TESTS_PASSED))
         return 0
     else
         echo -e "${RED}✗${NC} $description (file not found: $file)"
-        ((TESTS_FAILED++))
+        ((++TESTS_FAILED))
         return 1
     fi
 }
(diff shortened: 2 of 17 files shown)


michaelheichler:
Scaled this PR down. Force-pushed a minimal replacement (`4aca71f`) that drops the shared-framework migration and keeps only the three changes that are strictly required to clear the cascade:

1. `((VAR++))` → `((++VAR))` across 15 test files (postfix-increment silent-abort under `set -e`).
2. `grep -q PAT` → `grep PAT >/dev/null` in the two files hitting the SIGPIPE/pipefail path (`test-model-config-v849.sh`, `test-provider-activation.sh`). Keeps `set -euo pipefail` intact in both files, no strict-mode toggles.
3. `test-claude-mem-integration`: extract `save_session_checkpoint` body with awk before grepping, so the assertion can't be satisfied by the standalone `memory_observe` definition in `lib/memory.sh`. Uses a here-string (`<<<`) instead of a pipe to avoid any pipefail interaction.

**What I deliberately dropped** from the earlier round of CodeRabbit feedback:

- Sourcing `tests/helpers/test-framework.sh` in `test-docs-sync.sh` and `test-security-functions.sh`, and converting their local `pass`/`fail` helpers into shims around `test_case`/`test_pass`/`test_fail`. The local helpers have worked for a long time; the refactor was cosmetic and expanded the diff beyond what the #269 cascade requires.
- The `return 1` → `return 0` rewrites in `test-docs-sync.sh` — no longer needed once the file isn't sourcing the strict-mode framework.
- The `trap 'rm -f "$ALL_SRC"' EXIT` add in `test-security-functions.sh` — same reason; the existing `rm -f` in `main()` remains the only cleanup path. Happy to open a separate tiny PR for the trap if you'd prefer it landed.

Net diff is now 16 files, 165 insertions / 164 deletions — effectively 1 net line. 137/137 test suites pass locally on this branch.

If CodeRabbit re-flags the framework refactor as a blocker I'd push back: it's not load-bearing for #269 and the maintainer burden of a 170-line refactor on a bugfix PR is real. Happy to open it as a separate stylistic cleanup PR with its own CI run if you want that work done.

michaelheichler:
Closing this PR. The cascade fixes are correct but the file-by-file work here needs a serious rebase once the broader test-suite migration to `tests/helpers/test-framework.sh` is properly planned — doing it piecemeal in a bugfix PR is the wrong shape. The underlying bug classes (`((VAR++))` under `set -e`, SIGPIPE under pipefail, grep-based dump assertions) are documented in issue #276 for anyone picking this up with the scope it deserves. Not blocking the rest of the #269 stack.

PR #273, opened 2026-04-15
ci: fix Portability Lint + skip pr-review on fork PRs

michaelheichler:
## Summary

Two pre-existing CI failures that fire on every PR but aren't caused by the PR being reviewed. Surfaced while shepherding #270 / #271 / #272 — every PR's status checks were red on these two even when the actual changes were trivial.

## 1. Portability Lint — GNU sed range with command grouping

`scripts/build-factory-skills.sh:78` used:

```bash
sed -n '/^trigger:/,/^[a-z_]*:/{ /^trigger:/d; /^[a-z_]*:/d; p; }'
```

BSD sed (macOS) doesn't accept the `{ ... }` command-grouping inside an address range. Replaced with an awk in/out-of-block state machine — portable and arguably more legible:

```bash
awk '
  /^trigger:/  { in_block=1; next }
  /^[a-z_]+:/  { in_block=0 }
  in_block     { sub(/^  /, ""); print }
'
```

Verified locally with the exact regex from `.github/workflows/test.yml`'s lint step.

## 2. `pr-review` — "Resource not accessible by integration"

The `Post Review Comment` step was crashing with:

> `RequestError [HttpError]: Resource not accessible by integration`

Root cause: this PR is from a fork. Two things bite simultaneously:

1. **Fork PRs receive no secrets.** `OPENAI_API_KEY` / `GEMINI_API_KEY` are unset in the runner, so the upstream review step produces empty/garbage output.
2. **`GITHUB_TOKEN` is read-only on fork PRs**, regardless of the workflow's `permissions:` block. `github.rest.issues.createComment` therefore 403s.

Result: the job fails on every fork PR — and right now every external contribution comes from a fork, so it fails 100 % of the time.

### Fix

- Guard the job with `head.repo.full_name == github.repository` so it only runs on PRs from branches in the base repo (where secrets and write tokens are available).
- Add `continue-on-error: true` on the Post Review Comment step as a belt-and-suspenders for future API hiccups (rate limits, transient 5xx, etc.).

### What I deliberately did NOT do

I did **not** switch the trigger to `pull_request_target`. That event runs in the base-repo context with full secrets and write token — exposing `OPENAI_API_KEY` / `GEMINI_API_KEY` to whatever code the fork PR ships. Standard GitHub Actions security guidance is to keep secret-using workflows on `pull_request` and accept that fork PRs won't get the bot review.

If you want fork PRs to receive a Claude Octopus review, the right pattern is a separate `workflow_run`-triggered job that runs after the unsafe checkout, but that's a bigger refactor — happy to do it as a follow-up if you'd like.

## Test plan

- [x] Portability Lint command from `test.yml` runs locally and reports `OK`
- [x] Workflow YAML parses (no syntax errors)
- [x] On the next merge to main, base-repo PRs still trigger pr-review; fork PRs cleanly skip

## Summary by CodeRabbit

* **Chores**
  * CI updated to skip review jobs for forked PRs and tolerate comment-posting failures.
  * Build tooling improved to extract and format skill metadata more robustly.
  * Relaxed strict error-handling in a build helper to allow continued execution on non-fatal script errors.
* **Tests**
  * Numerous test scripts standardized increment expressions and migrated several suites to a shared test framework for consistent reporting.

### DIFF
--- a/.github/workflows/claude-octopus.yml
+++ b/.github/workflows/claude-octopus.yml
@@ -112,8 +112,9 @@ jobs:
           if-no-files-found: ignore
 
   # Job for PR code review
+  # Fork PRs receive no secrets and a read-only GITHUB_TOKEN — the review can't run and the comment can't post.
   pr-review:
-    if: github.event_name == 'pull_request'
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
 
     permissions:
@@ -162,6 +163,7 @@ jobs:
           ./scripts/orchestrate.sh --ci code-review '{"target":"staged","publish":"never"}' 2>&1 | tee review-output.txt
 
       - name: Post Review Comment
+        continue-on-error: true
         uses: actions/github-script@v9
         with:
           script: |
--- a/scripts/build-factory-skills.sh
+++ b/scripts/build-factory-skills.sh
@@ -75,7 +75,11 @@ for src in "$SKILLS_SRC"/*.md; do
   # Extract trigger content to enrich description
   trigger=""
   if echo "$frontmatter" | grep -q "^trigger:"; then
-    trigger="$(echo "$frontmatter" | sed -n '/^trigger:/,/^[a-z_]*:/{ /^trigger:/d; /^[a-z_]*:/d; p; }' | sed 's/^  //' | head -5)"
+    trigger="$(echo "$frontmatter" | awk '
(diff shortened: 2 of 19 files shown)


michaelheichler:
Closing. The Portability Lint fix and fork-PR guard work, but this PR is stacked on #272 (test-claude-mem-integration is needed for its Unit Tests to pass) and #272 is being closed for scope reasons. The underlying CI hygiene and workflow-permission issues are better handled as part of a dedicated CI cleanup pass — flagged in issue #276.

Issue #276, opened 2026-04-15
Latent bugs surfaced while fixing #269 — triage & help wanted


michaelheichler:
While shepherding the #269 fix (PR #270, `memory.sh` pipefail leak) I reached for the thread on several pre-existing bugs that were hidden by the cascade. Some of them I've fixed in stacked PRs (#271, #272, #273, #275). Others I'm flagging here because they're real and I don't want them to get lost, but I can't reasonably own the whole backlog. Happy to help with any of them — just want to write them down first.

### 1. `PROGRESS_FILE` is referenced but never defined — **fixed in #271**
`lib/session.sh` and `lib/agents.sh` both use `$PROGRESS_FILE`. No file in `scripts/` assigns it. The v9.7.x extraction that moved `init_progress_tracking()` out of `orchestrate.sh` left the variable behind. Once `check_ux_dependencies()` flips `PROGRESS_TRACKING_ENABLED=true` (jq present), `init_progress_tracking` does `mv .tmp.NNNN ''` and crashes. `discover` and `embrace` died deterministically for users with jq installed.

### 2. `score_result_file` returns multi-line numbers — **fixed in #275**
`lib/heuristics.sh` uses `headers=$(grep -c '^#' <<< "$content" || echo "0")`. `grep -c` exits 1 on no-match *and* prints `0` to stdout, so the `|| echo "0"` fallback appends a second `0`, yielding `"0\n0"`. Next arithmetic expansion dies with `[[: 0\n0: arithmetic syntax error`. Any probe whose synthesis reaches `rank_results_by_signals` hangs the orchestrator indefinitely (no `probe-synthesis-*.md` is ever written — agents finish, progress stalls, user sees a silent 30+ minute hang).

### 3. `((VAR++))` + `set -e` silently aborts 15 test files — **fixed in #272**
[...]

PR #274, opened 2026-04-15
feat: add Cursor Agent CLI as provider (Grok 4.20)

ImL1s:
## Summary
- Add Cursor Agent CLI as a new AI provider, enabling access to Grok 4.20 (xAI) via `cursor-agent.sh`
- Wire into all 9 required orchestration files: dispatch, providers, doctor, models, orchestrate, check-providers, install-deps, model-config
- Fix `qwen.sh` bug: remove `--no-ask-user` flag that doesn't exist in Qwen CLI v0.14.4

## Motivation

This PR adds official support for Cursor Agent CLI (`agent -p`) as a first-class provider. It uniquely provides access to Grok 4.20 (xAI), a model unavailable through existing providers (Codex=OpenAI, Gemini=Google, Qwen=Alibaba, Claude=Anthropic). Following the established headless `-p` pattern, it incurs zero extra API cost for Cursor subscribers and strengthens multi-model consensus by adding a 5th independent model family.

## Changes

| File | Description |
|------|-------------|
| `scripts/lib/cursor-agent.sh` | **New**: 3 functions + `_is_cursor_agent_binary()` identity check (97 lines) |
| `scripts/lib/dispatch.sh` | Add `cursor-agent` command routing + provider mapping + model allowlist |
| `scripts/lib/providers.sh` | Add `cursor-agent` health check + include in `check_all_providers()` loop |
| `scripts/lib/doctor.sh` | Add `cursor-agent` diagnostic with auth detection |
| `scripts/lib/models.sh` | Add `grok-4-20`, `grok-4-20-thinking`, `composer-2-fast`, `composer-2` to catalog |
| `scripts/orchestrate.sh` | Source `cursor-agent.sh` + add to `AVAILABLE_AGENTS` |
| `scripts/helpers/check-providers.sh` | Add `cursor-agent` availability check |
| `scripts/helpers/octo-model-config.sh` | Add to `KNOWN_PROVIDERS` |
| `scripts/install-deps.sh` | Add `cursor-agent` dependency check |
| `scripts/lib/qwen.sh` | Remove invalid `--no-ask-user` flag (Qwen CLI v0.14.4 compat) |
| `CHANGELOG.md` | Add entries for both changes |

## Design Decisions

- **Binary identity check**: `agent` is a generic binary name. `_is_cursor_agent_binary()` verifies the version string matches Cursor's `2026.xx.xx-hash` format to prevent collision with other tools.
[...]

PR #278, opened 2026-04-16
fix: stop silent test aborts (((VAR++)) + set -e) and fix BSD sed portability

nyldn:
## Summary

Fixes the two remaining unfixed bugs from #276 (items 3 and 5, originally in closed PRs #272 and #273).

### Bug 1: `((VAR++))` + `set -e` silently aborts 13 test files

`((VAR++))` evaluates to `0` on the first call (postfix returns old value before incrementing). Bash treats exit code 0-from-arithmetic as success, but the *expression value* `0` makes `((...))` return exit code 1. With `set -e`, the script aborts before any test actually runs. The test appears to pass (exit 0 from the header echo) with 0 failures — because nothing ran.

**Fix**: `((++VAR))` — pre-increment returns the new value, never `0`.

**Files fixed**: 13 test files across `tests/` and `tests/unit/`.

### Bug 2: GNU-only `sed` in `build-factory-skills.sh:78`

BSD sed (macOS) rejects command grouping inside address ranges (`sed -n "/start/,/end/{ ... }"`). This causes the Portability Lint CI check to fail on every PR.

**Fix**: Replaced with a portable `awk` in/out-of-block state machine.

Ref: #276 items 3 and 5

## Summary by CodeRabbit

* **Refactor**
  * Improved build-time parsing for more reliable trigger extraction.
  * Standardized increment semantics across test scripts for consistency.
* **Tests**
  * Updated test helpers and suites to use uniform counter behavior, improving test-reporting consistency and reliability.

### DIFF
--- a/scripts/build-factory-skills.sh
+++ b/scripts/build-factory-skills.sh
@@ -75,7 +75,7 @@ for src in "$SKILLS_SRC"/*.md; do
   # Extract trigger content to enrich description
   trigger=""
   if echo "$frontmatter" | grep -q "^trigger:"; then
-    trigger="$(echo "$frontmatter" | sed -n '/^trigger:/,/^[a-z_]*:/{ /^trigger:/d; /^[a-z_]*:/d; p; }' | sed 's/^  //' | head -5)"
+    trigger="$(echo "$frontmatter" | awk '/^trigger:/{found=1;next} /^[a-z_]*:/{found=0} found{sub(/^  /,"");print}' | head -5)"
   fi
 
   # Build Factory-compatible description
--- a/scripts/build-factory-skills.sh
+++ b/scripts/build-factory-skills.sh
@@ -72,10 +72,20 @@ for src in "$SKILLS_SRC"/*.md; do
   description="$(echo "$frontmatter" | grep "^description:" | head -1 | sed 's/^description: *//' | sed 's/^"//' | sed 's/"$//')"
   description="$(normalize_single_line "$description")"
 
-  # Extract trigger content to enrich description
+  # Extract trigger content to enrich description.
+  # Uses <<< "$frontmatter" (no pipelines) and awk's internal `exit` after 5 lines
+  # so neither grep -q nor head -n can SIGPIPE the pipeline under `set -euo pipefail`.
   trigger=""
-  if echo "$frontmatter" | grep -q "^trigger:"; then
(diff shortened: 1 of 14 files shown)


nyldn:
Thanks for the thorough review, @coderabbitai. Two substantive responses:

## ✅ Critical: SIGPIPE risk in trigger extraction — fixed

Applied the suggested `<<< "$frontmatter"` + awk-internal-exit approach in 4b3ec97. Removes both the `grep -q` and `head -5` pipeline hazards that would have caused the same SIGPIPE-under-pipefail bug this PR is designed to eliminate elsewhere. Good catch — fixing this here rather than leaving it as a latent instance was the right call.

## 🟠 7 test files not using shared test-framework.sh

These 7 comments all flag the same pattern: test files using local `pass()`/`fail()` helpers instead of `test_suite`/`test_case`/`test_pass`/`test_fail` from `tests/test-framework.sh`.

The concern is legitimate — consistent test harness usage is a repo coding guideline — but **every flagged file already had this pattern before this PR**. This PR's only change to those files is the mechanical `((++VAR))` fix to stop silent aborts. Migrating to the shared harness is a separate refactor that would:

- Change test semantics (the shared harness has different reporting, timing, skip behavior)
- Touch 7+ test files each with their own assertion patterns
- Increase the diff from +72/-72 to several hundred lines
- Create merge-conflict surface area for PRs in flight (#274 in particular)

The alternative — land this PR as a focused fix for the silent-abort + BSD sed portability bugs, then open a follow-up that migrates the bespoke helpers to the shared framework — is cleaner both for review and for bisect. I'm going to leave the helper style as-is here. If the follow-up migration is wanted, I'm happy to open it as a separate PR.

CI is green across all suites (smoke, unit, integration, portability, pr-review). Ready for merge decision.

nyldn:
Thanks for flagging this! These fixes have been applied directly to main as part of a comprehensive QA pass:

- `((VAR++)) || true` added to all 40+ instances across the codebase
- BSD sed compatibility fixed (`sed -E` for ERE, removed `\|` alternation in basic mode)
- `grep -oP` replaced with portable `grep -oE` + `sed` throughout
- `date -Iseconds` replaced with `date -u +%Y-%m-%dT%H:%M:%SZ`
- `mapfile` replaced with `while read` loop

Closing as superseded — all the same bugs and more are covered. 🐙

PR #279, opened 2026-04-18
fix(qwen): drop --no-ask-user (invalid Qwen flag — Copilot CLI cross-contamination)

ImL1s:
## Summary

`scripts/lib/qwen.sh` invokes `qwen` with `--no-ask-user`, but that flag is not a valid Qwen Code CLI option (verified against the [official Qwen Code headless docs](https://qwenlm.github.io/qwen-code-docs/en/users/features/headless): valid flags are `-p`, `-o`, `--yolo`/`-y`, `--approval-mode`, `--system-prompt`, `--continue`, `--resume`). Every Qwen invocation was failing on flag-parse.

## Root cause

`--no-ask-user` is the **Copilot CLI**'s flag — see `scripts/lib/copilot.sh:79,81` and `scripts/lib/dispatch.sh:103`. It looks like the flag got cross-contaminated into `qwen.sh` when the Qwen provider was first added. This is a strict CLI-arg-parser hard-fail (yargs/commander style), so the bug couldn't slip past silently.

## Fix
```diff
-response=\$(timeout \"\$timeout\" qwen -p \"\$prompt\" --approval-mode yolo -o text --no-ask-user 2>&1) ...
+response=\$(timeout \"\$timeout\" qwen -p \"\$prompt\" --approval-mode yolo -o text 2>&1) ...
```

`--approval-mode yolo` is already passed and provides the needed non-interactive behavior.

## Back-compat

No regression risk. `--approval-mode yolo` has been present in Qwen CLI since v0.1.x (Gemini CLI fork inheritance — Gemini's own provider in `dispatch.sh:73` only uses `-o text --approval-mode yolo`, never `--no-ask-user`). Removing the invalid flag fixes all Qwen versions, breaks none.

## Verification

- \`bash -n scripts/lib/qwen.sh\` — passes
- \`qwen --help | grep -E 'no-ask-user|approval-mode'\` — confirms `--no-ask-user` not present, `--approval-mode` present
- Confirmed `upstream/main:scripts/lib/qwen.sh:59` still has the broken flag (this fix is still needed)
- Spot-checked Qwen Code 0.14.x — same flag set throughout

## Why standalone

Per @nyldn's review on #274 (2026-04-17): *\"consider landing [the qwen fix] as a standalone PR first.\"*

## Note on previous force-push

This PR initially also bundled a `memory.sh` change. I noticed that conflicted with `cfaf6871 fix(#269)` (which correctly removes the top-level `set` directive entirely), so I dropped that commit and force-pushed. PR is now strictly the qwen one-liner.

## Suggested follow-up (NOT in this PR)

Worth considering a tiny test (e.g. \`tests/unit/test-adapter-flags.sh\` style) that asserts \`scripts/lib/qwen.sh\` does not contain Copilot-only flags, to catch future cross-contamination. Happy to open that as a separate PR if you'd like.

## Relation to #274

#274 will be closed and re-submitted as a focused Cursor Agent provider PR after this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Qwen CLI now prompts for user approval during execution instead of operating in non-interactive mode.

### DIFF
--- a/scripts/lib/qwen.sh
+++ b/scripts/lib/qwen.sh
@@ -56,7 +56,7 @@ qwen_execute() {
     [[ "${VERBOSE:-}" == "true" ]] && log DEBUG "qwen_execute: type=$agent_type, timeout=${timeout}s, auth=$(qwen_auth_method)" || true
 
     local response exit_code
-    response=$(timeout "$timeout" qwen -p "$prompt" --approval-mode yolo -o text --no-ask-user 2>&1) && exit_code=0 || exit_code=$?
+    response=$(timeout "$timeout" qwen -p "$prompt" --approval-mode yolo -o text 2>&1) && exit_code=0 || exit_code=$?
 
     # Handle errors
     if [[ $exit_code -ne 0 ]]; then


ImL1s:
@nyldn — Quick correction. I just noticed that the second commit in this PR (`fix(memory): use set -eo pipefail`) **conflicts with your earlier fix** in `cfaf6871 fix(#269): drop set -euo pipefail leak from sourced lib/memory.sh`. Your fix completely drops the top-level `set` directive (the right call); my commit was adding back `set -eo pipefail`, which is a partial revert.

I've force-pushed to drop that commit. **The PR is now just the qwen `--no-ask-user` fix**, which is still needed (verified `upstream/main:scripts/lib/qwen.sh:59` still has the flag). Sorry for the noise.

ImL1s:
@coderabbitai review

(The walkthrough is stale — it still mentions the `memory.sh` change that was force-pushed out. Please re-analyze the current diff which is just the qwen one-liner.)

ImL1s:
FYI — ran end-to-end verification on this PR before letting you merge:

**Test 1** — direct CLI: `qwen -p "test" --approval-mode yolo -o text` → CLI accepts the command (no flag-parse error). Confirms the new command line is valid.

**Test 2** — confirm bug: `qwen ... --no-ask-user` → CLI prints help (rejects unknown flag). Confirms upstream `qwen.sh:59` still has the broken flag (this fix is still needed).

**Test 3** — wrapping function: sourced `scripts/lib/qwen.sh` from `fix/shell-compat` branch and invoked `qwen_execute("qwen", "reply: e2e wrapping ok", 30)`. Returned exit 1 with `qwen: Auth failure — run: qwen (to trigger OAuth) or set QWEN_API_KEY`. The wrapper **runs to completion through the new command line** and exits cleanly via the existing auth-failure path. Fix verified end-to-end.

**Honest scope of "verified"**: Test 3 is a targeted integration check (sourced lib + stub `log()`), not a full live happy-path E2E (qwen OAuth on this machine is currently in an error state — see separate comment below). The deterministic claim is: the wrapper invokes `qwen` without `--no-ask-user` and propagates errors correctly. A stronger check would be a fake-`qwen` on `PATH` asserting argv directly — happy to add a `tests/unit/test-qwen-provider.sh` as a follow-up if useful.

Tested with `qwen` v0.14.5 on macOS.

ImL1s:
**Separate / unrelated to this PR** — heads-up only, not blocking:

While running the verification above I noticed the Qwen CLI now reports:

> `Qwen OAuth free tier was discontinued on 2026-04-15. Run /auth to switch to Coding Plan, OpenRouter, Fireworks AI, or another provider.`

This is a **provider-side change**, not a code issue. Existing users who relied on the free OAuth tier for Qwen will see this message even after this PR lands (the flag fix is still needed; the auth concern is independent). Might be worth a follow-up README/docs note in a separate PR — happy to draft one if you'd like.

nyldn:
@ImL1s thanks for the thorough verification. The `--no-ask-user` fix is applied directly in v9.26.0 (in `scripts/lib/qwen.sh:59`). Your end-to-end test approach was exemplary — the 3-test matrix was more rigorous than what we typically get.

Re: the Qwen OAuth free tier discontinuation — noted. We'll add a docs note about alternative auth methods (Coding Plan, OpenRouter, Fireworks AI) in the provider section. Thanks for the heads-up.

PR #308, opened 2026-04-22
fix: stdin inheritance in run_with_timeout + nested json_extract for OpenRouter/Perplexity (closes #307)

### LINKED ISSUE #307: PR #304 stdin fallback defeated by bash job-control + json_extract nested-path bug — perplexity still non-functional
# PR #304 stdin fallback defeated by bash job-control + pre-existing json_extract bug — perplexity still non-functional

## Summary

PR #304's stdin fallback for `perplexity_execute` / `openrouter_execute` is necessary but not sufficient. Testing PR #304's patch locally reveals two additional blockers — one is a bash semantic I missed in the original #305 investigation, the other is a pre-existing JSON-parsing bug in `perplexity_execute`. Both need fixing together for perplexity probes to actually produce output.

## Environment

- **Plugin:** octo 9.27.0 + PR #304 stdin-fallback patch applied locally
- **Claude Code:** 2.1.87
- **OS:** macOS Darwin 25.3.0 (arm64)
- **Shell:** zsh → bash (non-interactive, job control off)

### Evidence

```bash
# Minimal repro — background cmd inside pipe cannot read stdin
$ bash -c '
  read_stdin() { echo "stdin: [$(cat)]" >&2; }
  printf "hello world" | ( read_stdin & wait )
'
stdin: []   # ← nothing, despite printf writing "hello world"
```

Bash manual confirms: "If job control is not in effect (as when the shell is not interactive) ... the standard input for an asynchronous command is redirected from `/dev/null`."

`run_with_timeout` is called inside a non-interactive bash pipeline (`printf | run_with_timeout | tee`). The `"$@" &` inside it creates a background job with stdin redirected to `/dev/null`, so perplexity_execute's new `$(cat)` fallback sees an empty stdin.

### Evidence from orchestrator

Running `orchestrate.sh probe-single perplexity 'What year?'` with PR #304's patch:

```
# .tmp-*.err log:
[ERROR]: Perplexity error: "message":"Message content was empty","type":"invalid_message","code":400
```

perplexity_execute called with empty `$2` AND empty stdin → empty JSON → Perplexity API 400.

### Proposed fix — `scripts/lib/heartbeat.sh`

```diff
     else
         # Fallback with proper cleanup (also used for shell functions)
         local cmd_pid monitor_pid
-
-        "$@" &
+        # Explicit <&0 so background shell-function jobs inherit stdin from the pipe.
+        # Non-interactive bash redirects stdin to /dev/null for `cmd &` without this.
+        "$@" <&0 &
         cmd_pid=$!
```

### Verification

```bash
$ bash -c '
  read_stdin() { echo "stdin: [$(cat)]" >&2; }
  printf "hello world" | ( read_stdin <&0 & wait )
'
stdin: [hello world]   # ← works
```

## Blocker 2 — `json_extract` called with wrong JSON path in `perplexity_execute`

Even with stdin flowing correctly (after applying fix above), perplexity_execute still fails because the response parser looks at the wrong path.

### Code in `scripts/lib/perplexity.sh` (both perplexity_execute and openrouter_execute)

```bash
local content=""
if json_extract "$response" "content"; then
    content="$REPLY"
fi
```

`json_extract` uses `jq -re --arg f "$field" '.[$f] // empty'` — **top-level field only**.

Perplexity API returns:

```json
{
  "choices": [
    { "message": { "content": "It is 2026." } }
  ]
}
```

`.content` at top level is empty, so `content=""`. The "Empty response" warning fires and the function echoes the raw JSON back — but the caller expects the extracted text.

### Proposed fix — use nested path with jq directly

```diff
 local content=""
-if json_extract "$response" "content"; then
-    content="$REPLY"
+if command -v jq &>/dev/null; then
+    content=$(printf '%s' "$response" | jq -re '.choices[0].message.content // empty' 2>/dev/null) || content=""
 fi
```

Same fix needed in `openrouter_execute` and `openrouter_execute_model` (same call pattern).

## Combined effect

With both fixes applied (on top of v9.27.0 + PR #304), perplexity probe produces the correct output:

```
$ bash orchestrate.sh probe-single perplexity 'What year is it?' test 'test'
# Result file: ## Output contains "It is 2026." with citations
# Status: SUCCESS
```

Without them: either 400 from empty content (Blocker 1) or "Empty response" warning + raw JSON dumped to result file (Blocker 2).

## Test coverage suggestion

`test-v8.24.0-perplexity-integration.sh` should exercise the full orchestrator path:

```bash
# Run through probe-single, not direct perplexity_execute call
result=$(bash orchestrate.sh probe-single perplexity 'test prompt' test-id 'test' 2>&1)
grep -q "Status: SUCCESS" "<path/to/result-file>"
```

Direct function call tests pass today (I verified); the orchestrator path is where the compound bug hides.

## Related

- [#300](https://github.com/nyldn/claude-octopus/issues/300) — original two-bug report, fixed in v9.27.0
- [#305](https://github.com/nyldn/claude-octopus/issues/305) — partial fix for #300 fix 2, PR #304 pending merge
- [#304](https://github.com/nyldn/claude-octopus/pull/304) — QA-hardening branch with the stdin fallback

Thanks for the fast iteration on #300 and #305. This follow-up is deeper than I realized in #305 — the stdin fallback alone isn't enough on non-interactive bash.

nyldn:
Fixes both blockers from #307:

**Blocker 1 — stdin lost in background jobs:** `run_with_timeout` fallback uses `"$@" &` which redirects stdin to /dev/null in non-interactive bash. Fixed with explicit `<&0`.

**Blocker 2 — json_extract wrong path:** OpenRouter/Perplexity APIs return content at `.choices[0].message.content`, not top-level `.content`. Added jq nested path extraction with fallback to `json_extract`.

Applied to `execute_openrouter` and `openrouter_execute_model` in orchestrate.sh.

Closes #307

### DIFF
--- a/scripts/orchestrate.sh
+++ b/scripts/orchestrate.sh
@@ -7,27 +7,6 @@ set -eo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PLUGIN_DIR="$(dirname "$SCRIPT_DIR")"
-# Cache platform detection — avoids repeated subprocess spawns (v8.33.0)
-OCTOPUS_PLATFORM="$(uname)"
-
-# v8.36.0: Host runtime detection — Claude Code vs Factory AI Droid vs Codex vs Gemini
-# Factory's plugin interop resolves ${CLAUDE_PLUGIN_ROOT} automatically,
-# but we detect the host for version checking and env var fallbacks.
-# v9.16.0: Extended for Codex CLI and Gemini CLI host detection (Direction A)
-if [[ -n "${DROID_PLUGIN_ROOT:-}" ]]; then
-    OCTOPUS_HOST="factory"
-    # Factory provides DROID_PLUGIN_ROOT; ensure CLAUDE_PLUGIN_ROOT is also set
-    export CLAUDE_PLUGIN_ROOT="${CLAUDE_PLUGIN_ROOT:-$DROID_PLUGIN_ROOT}"
-elif [[ -n "${CODEX_HOME:-}" || -n "${CODEX_SANDBOX:-}" || -n "${CODEX_PLUGIN_ROOT:-}" ]]; then
-    OCTOPUS_HOST="codex"  # HOST:codex — Codex CLI is the host runtime
-elif [[ -n "${GEMINI_HOME:-}" || -n "${GEMINI_PLUGIN_ROOT:-}" ]]; then
-    OCTOPUS_HOST="gemini"  # HOST:gemini — Gemini CLI is the host runtime
-elif [[ -n "${CLAUDE_PLUGIN_ROOT:-}" ]]; then
-    OCTOPUS_HOST="claude"
-else
-    OCTOPUS_HOST="standalone"
-fi
-
 # Keep debug flag defined even when nounset is enabled by sourced scripts.
 OCTOPUS_DEBUG="${OCTOPUS_DEBUG:-false}"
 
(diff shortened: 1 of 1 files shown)


nyldn:
Closing as duplicate — this fix shipped in #310 (merged earlier today) with a more complete solution:

- #310 fixes both blockers from #307 in their extracted lib locations (`scripts/lib/heartbeat.sh:156` and `scripts/lib/perplexity.sh:123,213`) rather than inside `orchestrate.sh`
- Adds regression tests (`tests/unit/test-perplexity-issue-307.sh`, 6 tests)
- Verified against live Perplexity API (now surfaces `insufficient_quota` instead of the old silent `Message content was empty, code 400`)

Thanks for the fast turnaround — #310 ended up being the landed version via the extracted-lib path.

Issue #38, opened 2026-02-20
[BUG] Skills simulate workflow instead of calling orchestrate.sh


moozgoos:
/octo:research runs all agents as Opus instead of spawning (After asking Claude what models were used to spawn agents, he confirms that Opus 4.6 was chosen). Other features may also malfunction (no consumption of Gemini tokens).
Codex/Gemini via orchestrate.sh. ~/.claude-octopus/results/ and 

/logs/ remain empty. 

Setup confirms Codex and Gemini are installed 
and authenticated (OAuth). Version 8.7.0 (latest from plugin market, but in Claude plugin browser it says 8.15.0), Claude Code v2.1.49.

//EDITIt seems that
`/octo research app ideas`
works on all providers, but at the same time
`/octo:research app ideas`
spawns agents with default Claude model only

PR #334, opened 2026-05-01
fix(develop): replace recursive Skill call with direct orchestrate.sh dispatch

SouroucheB:
## Problem

`/octo:develop` causes an infinite loop in Claude Code.

When the command file is loaded, it instructs Claude to call `Skill(skill: "octo:develop")`. In Claude Code's skill resolution, this resolves back to the same command file (not the internal `flow-develop` skill), creating a recursive loop that never executes `orchestrate.sh`.

Related to #38.

## Root Cause

Both the command (`commands/develop.md`, name: `develop`) and the skill (`skills/flow-develop.md`, alias: `develop`) resolve to `octo:develop` in the plugin namespace. Claude Code's `Skill` tool hits the command file instead of the skill, causing the loop.

## Fix

Replace `Skill(skill: "octo:develop")` with a direct `bash` call to `orchestrate.sh develop`, which is exactly what `flow-develop` would do if resolution worked correctly.

```bash
bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" develop "<user's arguments here>"
```

## Tested

Confirmed the loop is broken and `orchestrate.sh` is invoked directly.

## Summary by CodeRabbit

* **Documentation**
  * Require dispatching development via the CLI orchestrator; forbid alternate dispatch methods or tool substitutions
  * Updated examples and wording to reflect the mandated Bash dispatch, added a provider preflight step and required activation banner, clarified post-completion behavior to always ask the user what to do next
  * Added a “Model and Effort Policy” section covering Opus mode/defaults, effort override env vars, and durable memory recording guidance
* **Tests**
  * Added a unit check ensuring docs include the provider preflight, activation banner, correct dispatch order, and model/effort/memory policy markers

### DIFF
--- a/.claude/commands/develop.md
+++ b/.claude/commands/develop.md
@@ -84,6 +84,14 @@ AskUserQuestion({
 
 **Dispatches to the develop workflow via `orchestrate.sh` for the implementation phase.**
 
+### Model and Effort Policy
+
+- For develop/tangle work on Opus 4.7, use `xhigh` for complex implementation and `medium` for trivial work.
+- Fast Opus 4.6 mode is 6x more expensive ($30/$150 per MTok vs $5/$25 standard), uses extra-usage billing, applies only to Opus 4.6, and trades cost for lower latency with equivalent quality.
+- Default to Opus 4.7 standard mode for multi-phase workflows; use fast mode only for interactive single-shot requests when explicitly selected.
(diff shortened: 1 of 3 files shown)


nyldn:
Thanks for this. Replacing the recursive Skill call with a direct dispatch path removes a fragile layer from `/octo:develop`, and the change fit the direction we have been taking with command execution. Merged and included in the current release.

PR #341, opened 2026-05-03
fix(gemini-exec): stream stderr in real-time via tail -f

SouroucheB:
## Problem

Gemini CLI captures stderr internally and only flushes it on process exit — after 10 retries with exponential backoff (~4 min total). This prevented quota watchers in `agent-sync.sh` and `spawn.sh` from detecting exhaustion early: they were watching empty files for the full retry window.

## Fix

Stream the internal `$err_file` to the caller's stderr in real-time using `tail -f ... &` immediately after the file is created. Kill the tail process after Gemini exits.

Remove the deferred `printf '%s' "$last_err" >&2` and `cat "$err_file" >&2` reprints that were made redundant by the streaming.

## Impact

Quota messages are now visible within ~2s of the first failed attempt instead of ~4 min. **This is a prerequisite for the quota watchers in `agent-sync.sh` and `spawn.sh` to be effective** — without this fix, those watchers watch empty files regardless of their polling logic.

## Test

With a fake Gemini binary that writes quota errors to stderr, detection time dropped from ~4 min to 2s.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error message handling by streaming errors in real-time during command execution instead of displaying them at the end, reducing redundant output and providing better visibility into execution progress and any encountered issues.

### DIFF
--- a/scripts/helpers/gemini-exec.sh
+++ b/scripts/helpers/gemini-exec.sh
@@ -63,6 +63,13 @@ for model in "${model_list[@]}"; do
     : > "$stdout_file"
 
     # Buffer per attempt so a failed attempt's partial stdout never leaks.
+    # Stream stderr in real-time via tail -f so the quota fast-fail watcher in
+    # agent-sync.sh / spawn.sh sees Gemini retry messages immediately instead of
+    # waiting for all 10 internal retries (~4 min) before gemini-exec exits.
+    : > "$err_file"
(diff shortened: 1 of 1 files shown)


nyldn:
Thanks for this one. Streaming Gemini stderr in real time is a small implementation change, but it makes provider failures much easier to understand when someone is sitting there waiting on a run. Merged and included in v9.31.0.

PR #342, opened 2026-05-03
fix(quota-watcher): targeted PID kill + extended patterns + temp_out

SouroucheB:
## Problem

Three issues in the quota exhaustion watchers in `agent-sync.sh` and `spawn.sh`:

1. **Broad kill scope**: `pkill -f "gemini-cli/bundle/gemini.js"` kills *all* Gemini CLI processes for the current user — in a multi-agent setup with N parallel workers, one quota-hit kills every running Gemini session.

2. **Incomplete patterns**: `exhausted your capacity` missed `RetryableQuotaError` and the per-attempt format `Attempt N failed: ... exhausted` emitted by Gemini CLI during retries.

3. **Only stderr monitored**: stdout (`$temp_out`) was not watched, missing any messages that might route there.

## Fix

**Kill strategy (4/4 model debate verdict, 2026-05-03):** switch to Option B — background dispatch with captured PID, then `pkill -KILL -P "$_dispatch_pid"` + `kill -KILL "$_dispatch_pid"`. Targets only the process tree of this invocation. Normalizes SIGKILL exit code 137 → 1.

**Extended patterns:** add `RetryableQuotaError` and `Attempt [0-9]+ failed.*exhausted`.

**Dual-file monitoring:** watch both `$temp_err` and `$temp_out`.

**Housekeeping:** remove duplicate `local _quota_watcher_pid` declaration.

## Dependency

Requires #341 (`fix/gemini-exec-realtime-stderr`) — without real-time stderr streaming, the watcher watches empty files for ~4 min regardless of kill strategy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Gemini-like providers now fail faster on quota exhaustion via a background watcher that terminates affected processes early.
  * Provider stdout/stderr are consistently captured to per-run temp files for reliable output reading.
  * Non-Gemini runs preserve original exit status while still capturing output.
  * Temporary output files are cleaned up more robustly to prevent stale artifacts.

### DIFF
--- a/scripts/lib/agent-sync.sh
+++ b/scripts/lib/agent-sync.sh
@@ -12,8 +12,19 @@
 # Every parallel spawn loop MUST call fleet_dispatch_begin before the first
 # spawn_agent call and fleet_dispatch_end after the last one. The smoke test
 # tests/smoke/test-fleet-dispatch-guard.sh enforces this statically.
+if ! type start_quota_watcher >/dev/null 2>&1; then
+    _octopus_agent_sync_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    source "${_octopus_agent_sync_lib_dir}/quota-watcher.sh" 2>/dev/null || true
+fi
+
 fleet_dispatch_begin() {
     export OCTOPUS_FORCE_LEGACY_DISPATCH=true
+}
+
+quota_watcher_kill_sync_dispatch() {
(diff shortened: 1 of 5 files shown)


nyldn:
Thanks for digging into the quota-watcher cleanup. The `set -e` and `wait` interaction here is one of those shell edge cases that looks fine until cleanup silently stops happening. Merged and included in v9.31.0.

PR #346, opened 2026-05-05
Wire routing.features.* (review, parallel, debate) to their consumers

phjlljp:
## Summary

The `/octo:model-config` wizard prompts users for "Review providers", "Parallel execution providers", and "Debate participants", and writes their selections to `.routing.features.{review,parallel,debate}` in `~/.claude-octopus/config/providers.json`. Before this PR no consumer read those paths — the wizard's selections had no effect. Each affected command (`/octo:review`, `/octo:multi`, `/octo:debate`) used a hard-coded participant list and ignored user configuration.

This PR connects all three documented config knobs to their consumers in a backward-compatible way.

## Commits

1. **`scripts/lib/review.sh`** — `build_review_fleet()` reads `.routing.features.review` and assigns each listed provider to a review role (logic, security, arch, cve, diversity). Falls back to the original `command -v` cascade when config is absent or empty. Recognizes `openrouter` as a fifth role (`diversity-reviewer`) for cross-family perspective. Anchors the fleet with `claude-sonnet:arch-reviewer` when no Claude entry is configured.

2. **`scripts/lib/parallel.sh`** — `fan_out()` reads `.routing.features.parallel` and uses it as the agents list. Validates each entry against `AVAILABLE_AGENTS` so a typo cannot break `/octo:multi`. Falls back to the original `(codex, gemini)` pair when config is absent or all entries are invalid.

3. **`scripts/lib/debate.sh`** — `grapple_debate()` introduces three named slots (A/B/C) with both an agent_type and a label per slot. The first three resolved entries from `.routing.features.debate` override the defaults. Synthesis prompt headers, result-file headings, and the conclusion ceremony all track the active slot identities so the JUDGE and saved transcripts get correct attribution. WARN log fires when more than 3 participants are configured (debate caps at 3 by design — see "Limits" below).

## Why

Three separate features (review, parallel, debate) all collected user config that nothing consumed. That's worse than not exposing the option at all — the wizard's questions have no effect, and a user who configures `["claude","codex","gemini","openrouter"]` for `/octo:review` silently gets the cascaded default fleet instead.

Concrete request that surfaced this: "I want `/octo:review`, `/octo:multi`, and `/octo:debate` to include Codex, Gemini, Claude, and Qwen-via-OpenRouter by default." With `routing.features.*` already wired by the wizard and `providers.openrouter.default` already pointing at `qwen/qwen3-coder-plus`, this PR is the missing piece.

## Risk assessment

Low. All three changes are additive: each gates the new config-driven path on `.routing.features.<feature>` being a non-empty array, falling back to the unchanged hard-coded behavior otherwise. No existing user gets a different fleet/agents/participants unless they have explicitly configured one via the wizard.

## Test results

Smoke-tested each function with three configurations:

**`build_review_fleet`:**
- Config `["claude","codex","gemini","openrouter"]` → 4-role fleet (`arch:claude-sonnet`, `logic:codex`, `security:gemini`, `diversity:openrouter`).
- Empty array `[]` and missing `providers.json` → falls back to original cascade.
- Unknown providers in the array → silently skipped; cascade runs if all unknown.

**`fan_out`:**
- Config with 4 entries → 4-agent fan-out.
- No config → falls back to `(codex, gemini)`.
- Invalid entries → skipped with WARN; remaining valid entries proceed.

**`grapple_debate`:**
- Default (no config) → `Codex vs Gemini vs Sonnet` (note: lost `4.6` suffix from prior `Sonnet 4.6` wording — minor cosmetic regression in the box header).
- Config `["claude","codex","gemini","openrouter"]` → first 3 used: `Sonnet vs Codex vs Gemini`; openrouter dropped with WARN.
- Config `["codex","gemini","openrouter"]` → `Codex vs Gemini vs OpenRouter`.
- Synthesis prompt labels track the active slot identities — JUDGE sees `OPENROUTER HYPOTHESIS:` not `SONNET HYPOTHESIS:` when slot C is OpenRouter.

## Limits

- **Debate is capped at 3 participants.** Round 2 cross-critique is structured as a 3×2 matrix (each participant falsifies the other two), and Round 3 rebuttals follow the same shape. Going to N-way would require arrayizing proposals/critiques/rebuttals and loop-building the cross-critique prompts — bigger refactor, happy to follow up if you'd like. WARN log makes the truncation visible.
- **Minor codex-review behavior change.** Round 2 cross-critique previously used `codex-review` as the slot-A agent (a review-specialized variant of codex). After the slot refactor, slot A uses its configured agent_type uniformly across rounds. Both `codex-review` and `codex` resolve to the same model, so cost/quality is unchanged; the only difference is the internal agent_type tag in metrics.
- **Sonnet's `4.6` suffix dropped in default box header.** `label_c` defaults to `"Sonnet"` (short form) so config remaps stay symmetric (e.g., `Codex`, `Gemini`, `OpenRouter`, `Sonnet`). Restoring the `4.6` suffix in the default would make the configurable form awkward (`Sonnet 4.6` vs `OpenRouter` mismatch). Open to suggestions.

## Summary by CodeRabbit

* **New Features**
  * Reviewer, debate participant, and parallel-worker selection can be driven by a local config file; debate banners, transcripts, prompts, and startup logs now show configured participant names and agent identities. Duplicated display names are auto-disambiguated; an architecture reviewer is ensured if none configured.

* **Chores**
  * If config is missing/invalid, the system falls back to prior defaults; unresolvable entries are skipped with warnings and the chosen agent count/identities are logged.

* **Tests**
  * Added unit tests to validate provider→agent resolution and prevent hardcoded participant choices.

### DIFF
--- a/scripts/lib/debate.sh
+++ b/scripts/lib/debate.sh
@@ -30,10 +30,57 @@ grapple_debate() {
         rounds=3
     fi
 
+    # v9.31.0: Resolve participants from .routing.features.debate (providers.json).
+    # The /octo:model-config wizard writes a "Debate participants" array to that
+    # path; before this change there was no consumer. Slots A/B/C drive every
+    # run_agent_sync call, prompt label, and synthesis attribution below.
+    # Default trio: codex, gemini, claude-sonnet (preserves prior behavior).
+    local agent_a="codex" label_a="Codex" label_a_upper="CODEX"
+    local agent_b="gemini" label_b="Gemini" label_b_upper="GEMINI"
+    local agent_c="claude-sonnet" label_c="Sonnet" label_c_upper="SONNET"
+
+    local _debate_config_file="${HOME}/.claude-octopus/config/providers.json"
+    if [[ -f "$_debate_config_file" ]] && command -v jq >/dev/null 2>&1; then
+        local _participants _participant_count
+        _participants=$(jq -r '
+            (.routing.features.debate // [])
+            | if type == "array" then .[] else empty end
+        ' "$_debate_config_file" 2>/dev/null)
+        if [[ -n "$_participants" ]]; then
+            _participant_count=$(printf '%s\n' "$_participants" | grep -cv '^$')
+            if [[ "$_participant_count" -gt 3 ]]; then
+                log WARN "Debate supports 3 participants; using first 3 of $_participant_count from .routing.features.debate"
+            fi
+            local _slot_idx=0
+            local _provider _agent _label _label_upper
+            while IFS= read -r _provider; do
+                [[ -z "$_provider" ]] && continue
+                _agent=""; _label=""; _label_upper=""
+                case "$_provider" in
+                    claude|claude-sonnet)    _agent="claude-sonnet"; _label="Sonnet"; _label_upper="SONNET" ;;
+                    claude-opus)             _agent="claude-opus";   _label="Opus";   _label_upper="OPUS" ;;
+                    codex|codex-*)           _agent="$_provider";    _label="Codex";  _label_upper="CODEX" ;;
+                    gemini|gemini-*)         _agent="$_provider";    _label="Gemini"; _label_upper="GEMINI" ;;
+                    openrouter|openrouter-*) _agent="$_provider";    _label="OpenRouter"; _label_upper="OPENROUTER" ;;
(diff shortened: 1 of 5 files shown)


nyldn:
Thanks for wiring this through. The config UI was already writing the `routing.features.*` values, but without consumers that was more confusing than useful. This PR closed that gap cleanly. Merged and included in v9.31.0.

PR #427, opened 2026-05-28
fix: add --trust --output-format text to cursor-agent smoke test

nyldn:
## Root cause

`scripts/lib/smoke.sh` (lines 1038–1042) invoked `cursor-agent` without `--trust` and `--output-format text`, while the dispatch path in `scripts/lib/cursor-agent.sh:143` passes both flags. This caused the smoke test to fail with `Workspace Trust Required` in any directory Cursor hasn't been granted workspace trust to, making `/octo:doctor` falsely report `cursor-agent` as broken.

## Fix

Added the two missing flags to the cursor-agent branch of `_smoke_test_provider`, bringing it into parity with the dispatch path:

```diff
-        # Cursor Agent: prompt via stdin with -p "" for headless trigger
+        # --trust required for untrusted workspaces; matches cursor-agent.sh:143 dispatch path
         echo "Reply with exactly: ok" | run_with_timeout "$smoke_timeout" \
-            $cmd_str -p "" \
+            $cmd_str -p "" --trust --output-format text \
             >/dev/null 2>"$stderr_file" || smoke_exit=$?
```

## Verification

- `bash -n scripts/lib/smoke.sh` — syntax OK
- Single-file change, no test infrastructure to run locally. The fix is a direct flag parity match to the already-working dispatch path at `cursor-agent.sh:143`.

Closes #426

---
_Generated by [Claude Code](https://claude.ai/code/session_01JCkFyy9Et7nZBMzXRqwTwz)_

## Summary by CodeRabbit

* **New Features**
  * Added a new "preflight" command to the tool registry for running preflight checks.

* **Chores**
  * Updated smoke-test invocation to include additional flags for more reliable and readable test output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nyldn/claude-octopus/pull/427?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #429, opened 2026-05-28
fix: use -e instead of -f for skill existence check in doctor

### LINKED ISSUE #428: Doctor reports 53 false-positive skill failures (file check on directories)
## Bug

`/octo:doctor` reports all skills as missing with "Skill file missing" errors, even though every skill directory exists and is complete.

**Affected version:** v9.38.0  
**Category:** `[skills]` check in doctor

## Root Cause

In `scripts/lib/doctor.sh` around line 818, the skill existence check uses `-f` (regular file test):

```bash
if [[ ! -f "$resolved" ]]; then
```

But `plugin.json` lists skills as directory paths (e.g. `./skills/flow-define`), not files. Each skill is a directory containing `SKILL.md` and an `agents/` subdirectory. The `-f` test fails on directories, producing a false failure for every skill entry.

The commands check on line 838 uses the same `-f` pattern but is correct there — commands point to `.md` files.

## Impact

On a clean v9.38.0 install, `doctor --verbose` reports **53 failures** that are entirely false positives. This makes it impossible to distinguish real issues from noise in the skills category.

## Fix

Change `-f` to `-e` (exists) on the skill check so it passes for both files and directories:

```diff
-        if [[ ! -f "$resolved" ]]; then
+        if [[ ! -e "$resolved" ]]; then
```

After this one-character fix, doctor reports 0 failures and correctly shows "All N skill files present."

## Steps to Reproduce

1. Install octo v9.38.0 via marketplace
2. Run `/octo:doctor` or `orchestrate.sh doctor --verbose`
3. Observe 53 skill failures in the `[skills]` category
4. Verify skills actually exist: `ls ~/.claude/plugins/cache/nyldn-plugins/octo/9.38.0/skills/`

nyldn:
## Root cause

`doctor_check_skills()` in `scripts/lib/doctor.sh:913` used `[[ ! -f "$resolved" ]]` to test whether each skill entry from `plugin.json` exists. All 53 skills listed in `.claude-plugin/plugin.json` are directory paths (e.g. `./skills/flow-define`), not regular files. `-f` fails on directories, so every skill produced a false "Skill file missing" failure.

A previous partial fix added `&& [[ ! -f "$resolved/SKILL.md" ]]` to catch the common case where skills are directories containing `SKILL.md`. That happens to work today, but it's fragile: any skill not named `SKILL.md` re-introduces false positives.

## Fix

Replace the double file-check with a single existence test:

```diff
-        if [[ ! -f "$resolved" ]] && [[ ! -f "$resolved/SKILL.md" ]]; then
+        if [[ ! -e "$resolved" ]]; then
```

`-e` passes for regular files, directories, and symlinks — the correct semantic for "does this skill exist at all?"

## Verification

- `bash -n scripts/lib/doctor.sh` — syntax OK
- Single-line change, self-contained to the skill loop
- Spot-checked: all 53 skill entries are directories and all contain `SKILL.md`, confirming this fix eliminates every false positive reported in #428

Closes #428

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tg1shvpz27i2ZiPvSCyXkk)_

## Summary by CodeRabbit

* **New Features**
  * Added a new "preflight" command to the available tools, expanding available pre-check capabilities.

* **Chores**
  * Improved diagnostic checks for skill artifacts: presence detection now uses a broader existence check, reducing false positives and improving accuracy of missing-skill reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nyldn/claude-octopus/pull/429?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nyldn:
**Unit test CI status note:** The Unit Tests (ubuntu/macos) failures shown here are pre-existing — PR #427 (different change, same base commit `da1f682`) has the same identical failures. They are not caused by this PR.

Also pushed commit `199ba2e` to restore a trailing newline that was accidentally stripped by the initial push. The fix itself (`-e` on line 913) is unchanged.

---

PR #446, opened 2026-06-06
fix(gemini-exec): catch exhausted capacity as quota error + misc fixes

Adam2kg:
## Summary

- **fix(gemini-exec):** Add `exhausted your capacity` to `is_model_error` regex so Gemini quota failures on `gemini-2.5-pro` properly fall back to the configured fallback model instead of silently failing
- **refactor(setup):** Move cache TTL question out of first-run flow
- **feat(review):** Add opt-in `--deep` flag with Opus 4.7 second pass
- **chore:** Ignore Claude Code session files; update gemini-exec helper

## Root cause (gemini-exec fix)

The Gemini API returns `"You have exhausted your capacity on this model"` when hitting quota limits. This string matched none of the existing patterns in `is_model_error`, so quota exhaustion on the primary model did not trigger fallback — the agent just failed silently. Adding `exhausted your capacity` to the regex restores correct fallback behaviour.

## Test plan

- [ ] Run a probe with `gemini-2.5-pro` while quota is exhausted — confirm it falls back to `OCTOPUS_GEMINI_FALLBACK_MODELS` instead of failing
- [ ] Verify `gemini-exec.sh` unit tests still pass

## Summary by CodeRabbit

* **New Features**
  * Added `/octo:review --deep` flag for enhanced two-pass code reviews that combine standard analysis with deeper issue detection.

* **Documentation**
  * Simplified setup process by making prompt cache configuration optional, now available as a post-setup tip instead of an interactive prompt.

* **Bug Fixes**
  * Enhanced model error detection to handle additional failure cases and enable better fallback to alternative models.

### DIFF
--- a/.claude/commands/review.md
+++ b/.claude/commands/review.md
@@ -18,6 +18,10 @@ Three review entry points coexist in Claude Code v2.1.111+ — pick the right on
 | Claude-native `/review` | Single-turn, current diff | Claude only | Ordinary review, one perspective suffices |
 | `/ultrareview` (CC v2.1.111+) | Cloud, parallel multi-agent | Claude parallelism | Pre-merge PR review without leaving CC |
 | `/octo:review` (this) | Multi-LLM, inline PR comments | Codex + Gemini + Claude | Provider diversity, adversarial cross-check, stricter escalation |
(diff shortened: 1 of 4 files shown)


Adam2kg:
Closing — wrong branch scope, will resubmit with isolated fix only.

PR #452, opened 2026-06-07
fix: preserve Codex provider env config

Jhacarreiro:
## Summary

Preserve Codex CLI provider configuration when Octopus builds an isolated provider environment for `codex*` agents.

The Codex CLI can be configured to use OpenAI-compatible providers through `config.toml`, including a provider-specific `env_key`. Before this change, Octopus' credential-isolated env for Codex only forwarded `OPENAI_API_KEY`, `HOME`, `PATH`, and `TMPDIR`, which could drop `CODEX_HOME` and the provider-specific key required by the configured Codex provider.

This keeps credential isolation but forwards only explicit Codex config inputs:

- `CODEX_HOME`, when set
- the single `env_key` named in `${CODEX_HOME:-$HOME/.codex}/config.toml`, when present and set
- existing `OPENAI_API_KEY` behavior for backward compatibility

It does not forward arbitrary environment variables and does not hardcode any specific provider/router.

## Tests

- `bash -n scripts/lib/provider-routing.sh tests/test-credential-isolation.sh`
- `bash tests/test-credential-isolation.sh`

## Summary by CodeRabbit

* **New Features**
  * Codex provider configuration is now preserved during build by reading the Codex config (when present) to include a provider’s configured `env_key` value in the isolated environment.
  * The isolated environment can also retain the Codex home path when set, improving continuity for Codex-scoped provider settings.

* **Tests**
  * Expanded credential isolation coverage and added a runtime check to ensure Codex config–driven provider keys are included while unrelated provider keys are excluded.


Jhacarreiro:
Scope note after validating direct OpenAI-compatible tool loops separately: this PR remains useful as Codex runtime hardening, not as the path for generic OpenAI-compatible tool execution.

It preserves Codex CLI provider env configuration, including CODEX_HOME plus provider-specific env_key, when Octopus builds an isolated provider environment. That is still valuable whenever the selected agent is Codex and Codex is configured for a custom provider/router. Direct OpenAI-compatible agents should use their own tool-loop path instead of relying on Codex semantics.

Jhacarreiro:
ClawSweeper status: review started.

I am starting a fresh review of this pull request: fix: preserve Codex provider env config This is item 1/12 in the current shard. Shard 0/1.

This placeholder means the worker is alive and reading the current context. I will edit this same comment with the actual review when the claws are done clicking.

Crustacean status: shell secured, claws on keyboard, evidence pebbles being sorted.

PR #457, opened 2026-06-08
fix: use project root when decomposing tangle tasks

Jhacarreiro:
## Problem

TANGLE asks a model to split implementation work into parallel subtasks and assign each coding subtask a Files scope. When the command is launched from a wrapper or a prompt references an external plan, the model may not see the target repository layout. It can then invent generic paths, which makes the parallel-safety check less useful and can send workers toward the wrong files.

## Change

This adds the target repository file map to the decomposition prompt by reading tracked files from PROJECT_ROOT with git ls-files.

That gives the decomposer concrete repository paths before it writes Files scopes, instead of relying on generic names such as server.js, index.js, or invented directories.

## Why this is safe

The change only adds read-only repository context to the prompt. It does not change dispatch, execution, writes, or validation behavior.

## Validation

- bash -n scripts/lib/workflows.sh

### DIFF
--- a/scripts/lib/quality.sh
+++ b/scripts/lib/quality.sh
@@ -463,6 +463,10 @@ Be concise and specific. This is a planning exercise, not implementation."
     local design_claude_agent="${OCTOPUS_DESIGN_REVIEW_CLAUDE_AGENT:-claude-sonnet}"
     local design_synthesis_agent="${OCTOPUS_DESIGN_REVIEW_SYNTH_AGENT:-claude-opus}"
     local design_timeout="${OCTOPUS_DESIGN_REVIEW_TIMEOUT:-120}"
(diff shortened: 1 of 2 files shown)


Jhacarreiro:
Closing as superseded by #458.

Local conflict probe against current origin/main shows no remaining diff after keeping the #458/main side of the conflicting hunks. The project-root decomposition behavior is now covered by main via 9a499ba (#458), so keeping this PR open only leaves a stale conflicting stack item.

Jhacarreiro:
Closed as superseded by PR 458.

Local conflict probe against current origin/main showed no remaining diff after keeping the PR 458/main side of the conflicting hunks. The project-root/tangle context behaviour is now covered by main via 9a499ba, so keeping this PR open only left a stale conflicting stack item.

PR #458, opened 2026-06-08
fix: resolve tangle file scopes against repo context

Jhacarreiro:
## Problem

TANGLE requires coding subtasks to declare Files scopes so it can prevent unsafe parallel edits. In practice, model-generated scopes are often approximate: a basename, a generic path, or a file name from the task wording rather than the real path in the repository.

When that happens, Octopus can reject otherwise valid work or send the worker a scope that does not match the real worktree.

## Change

This treats Files values as scope intent and resolves them against the real repository before dispatch and validation.

It adds helpers that:
- keep exact existing file paths when they are valid;
- resolve basename-only references against tracked files;
- add relevant repository files from high-signal task terms;
- include resolved files in each subtask prompt;
- validate effective resolved scopes rather than only the model's raw path text.

## Why this is safe

The parallel-safety check remains in place. This only improves the file targets used by that check and by the worker prompt. If no concrete repository target can be resolved, the worker is told to inspect the tracked file list and report a blocker instead of inventing layout.

## Validation

- bash -n scripts/lib/workflows.sh

## Summary by CodeRabbit

* **Chores**
  * Design review: agent/provider selection is now configurable via environment variables; timeouts are validated/clamped and chosen agents/timeouts are logged.
  * Workflow engine: injects richer repository context into task prompts, resolves concrete file scopes, strips extraneous scope artifacts, enforces plan-file size caps, and tightens parallel-task conflict detection.
* **Tests**
  * Added regression tests for plan-file resolution, repo-root fallback behavior, and write-scope safety.

### DIFF
--- a/scripts/lib/quality.sh
+++ b/scripts/lib/quality.sh
@@ -463,6 +463,10 @@ Be concise and specific. This is a planning exercise, not implementation."
     local design_claude_agent="${OCTOPUS_DESIGN_REVIEW_CLAUDE_AGENT:-claude-sonnet}"
     local design_synthesis_agent="${OCTOPUS_DESIGN_REVIEW_SYNTH_AGENT:-claude-opus}"
     local design_timeout="${OCTOPUS_DESIGN_REVIEW_TIMEOUT:-120}"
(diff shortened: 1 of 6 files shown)


nyldn (approved):
Reviewed the follow-up fixes and reran focused coverage locally. Latest CodeRabbit status is green.

PR #461, opened 2026-06-08
fix: add Gemini skip-trust only when supported

Jhacarreiro:
## Problem

Octopus adds --skip-trust for Gemini in headless mode to avoid interactive trust prompts. Some newer Gemini CLI versions no longer support that flag. Passing an unsupported flag can break Gemini dispatch before the model is invoked.

## Change

Before appending --skip-trust, Octopus now checks whether the installed Gemini CLI advertises the flag in gemini --help.

Older Gemini versions that support the flag still receive it. Newer versions that removed it do not.

## Why this is safe

This preserves the existing headless trust bypass where available and avoids passing an invalid CLI option where it is not available.

## Validation

- bash -n scripts/lib/spawn.sh

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gemini agent reliability by automatically detecting whether the installed Gemini CLI supports the “skip trust” option before using it.
  * Prevented failures caused by unsupported flags across different Gemini CLI versions by conditionally applying the option only when available.
  * Reduced redundant support checks via lightweight caching to keep spawn behavior efficient.


Jhacarreiro:
ClawSweeper status: review started.

I am starting a fresh review of this pull request: fix: add Gemini skip-trust only when supported This is item 5/8 in the current shard. Shard 0/1.

This placeholder means the worker is alive and reading the current context. I will edit this same comment with the actual review when the claws are done clicking.

Crustacean status: shell secured, claws on keyboard, evidence pebbles being sorted.

PR #511, opened 2026-06-15
feat: emit agent lifecycle events

Jhacarreiro:
## Summary

Emit agent lifecycle events through the existing Octopus JSONL event stream and add an optional best-effort hook for external observers.

This keeps `OCTO_EVENT_LOG` as the primary integration surface and adds:

- `agent.spawned`
- `agent.completed`

Each lifecycle event includes provider, agent type, task id, role, phase, pid, result file, results/workspace dirs, exit code, status, and parent/root session ids.

For local control planes such as Crabfleet, `OCTOPUS_AGENT_LIFECYCLE_HOOK` can be set to an executable callback. The hook receives the same metadata via `OCTOPUS_AGENT_*` environment variables. Hook failures/timeouts are ignored so observability outages never fail agent execution.

## Why

Octopus now has `events.sh`, `OCTO_EVENT_LOG`, `octo-hud`, and provider lifecycle telemetry. Agent-level spawn/completion events make that stream useful to local control planes without requiring them to parse PID files or result markers.

The optional hook is intentionally secondary: it bridges live local observers, while the JSONL event stream remains the durable source of truth.

## Review follow-up

Addressed review feedback by:

- normalizing empty phase values to `unknown` consistently across JSONL and hook metadata;
- adding `OCTOPUS_AGENT_LIFECYCLE_HOOK_TIMEOUT` with a default 3s timeout;
- emitting `agent.spawned` for the Agent Teams dispatch path;
- using the project standard `TEST_TMP_DIR` cleanup pattern.

## Validation

- `bash -n scripts/lib/spawn.sh scripts/lib/events.sh tests/unit/test-agent-lifecycle-events.sh`
- `bash tests/unit/test-agent-lifecycle-events.sh` — 6 passed / 0 failed
- `bash tests/unit/test-event-monitor.sh` — 4 passed / 0 failed
- `bash tests/unit/test-agent-command-validation.sh` — 6 passed / 0 failed
- `bash tests/unit/test-octo-events.sh` — 11 passed / 0 failed, duration 115s

## Summary by CodeRabbit

* **New Features**
  * Agent lifecycle events (`spawned`, `completed`) are now emitted to an event stream for monitoring and observability.
  * Optional observer hook support with timeout handling and graceful failure management.
  * Events include metadata such as PID, result files, exit codes, and execution status.

* **Documentation**
  * Added comprehensive guide for agent lifecycle events, stream configuration, and observer hook setup.

PR #518, opened 2026-06-16
test: harden lifecycle hook timeout fallback

Jhacarreiro:
## Summary
- validate `OCTOPUS_AGENT_LIFECYCLE_HOOK_TIMEOUT` before using it
- harden the built-in hook timeout fallback with a `SECONDS`-based deadline
- avoid fractional `sleep` in the fallback path
- add regression coverage that forces the built-in fallback path when `timeout` is unavailable

## Context
Draft follow-up stacked on top of #511. The #511 repair already makes the direct hook fallback bounded; this follow-up makes that path easier to reason about and adds explicit coverage for systems without `run_with_timeout` or GNU `timeout`.

This is intentionally draft while #511 is still open, because the comparison against `main` includes the #511 commits. Once #511 lands, this should reduce to the hardening commit only.

## Verification
- `bash -n scripts/lib/spawn.sh`
- `bash tests/unit/test-agent-lifecycle-events.sh`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle hook timeout handling by sanitizing `OCTOPUS_AGENT_LIFECYCLE_HOOK_TIMEOUT` to a valid positive decimal value (including zero-padded inputs) and defaulting to `3` when invalid.
  * Updated timeout fallback behavior to use a wall-clock deadline with `kill -0` polling, reliably terminating the hook and its child processes (TERM then KILL), followed by a best-effort wait.

* **Tests**
  * Added regression coverage for zero-padded timeout values and for prompt fallback behavior when external timeout utilities are unavailable.

### DIFF
--- a/scripts/lib/spawn.sh
+++ b/scripts/lib/spawn.sh
@@ -72,32 +72,36 @@ _octopus_agent_lifecycle_event() {
         export OCTOPUS_AGENT_ROOT_SESSION_ID="${CRABFLEET_ROOT_SESSION_ID:-${OCTOPUS_ROOT_SESSION_ID:-}}"
         export OCTOPUS_AGENT_PARENT_SESSION_ID="${CRABFLEET_PARENT_SESSION_ID:-${OCTOPUS_PARENT_SESSION_ID:-}}"
         local hook_timeout="${OCTOPUS_AGENT_LIFECYCLE_HOOK_TIMEOUT:-3}"
+        if [[ ! "$hook_timeout" =~ ^[0-9]+$ ]]; then
+            hook_timeout=3
+        else
+            hook_timeout=$((10#$hook_timeout))
+            [[ "$hook_timeout" -lt 1 ]] && hook_timeout=3
+        fi
         if declare -f run_with_timeout >/dev/null 2>&1; then
             run_with_timeout "$hook_timeout" "$hook" "$event"
         elif command -v timeout >/dev/null 2>&1; then
             timeout "$hook_timeout" "$hook" "$event"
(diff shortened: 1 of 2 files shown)


Jhacarreiro:
Rebased onto current origin/main and validated as a standalone hardening follow-up to #511.\n\nValidation run:\n- bash -n scripts/lib/spawn.sh tests/unit/test-agent-lifecycle-events.sh\n- bash tests/unit/test-agent-lifecycle-events.sh → 7/7 PASS\n- git diff --check\n\n@coderabbitai review

Jhacarreiro:
Addressed CodeRabbit follow-up: the built-in lifecycle hook timeout fallback now sweeps child processes with pkill -P on TERM/KILL escalation. Validation: bash -n scripts/lib/spawn.sh tests/unit/test-agent-lifecycle-events.sh; bash tests/unit/test-agent-lifecycle-events.sh → 7/7 PASS; git diff --check. @coderabbitai review

Jhacarreiro:
Addressed latest CodeRabbit follow-up: OCTOPUS_AGENT_LIFECYCLE_HOOK_TIMEOUT is now coerced to decimal with 10# after digit validation, so zero-padded values like 08/09 no longer trigger Bash octal parsing. Added regression coverage for a zero-padded timeout. Validation: bash -n scripts/lib/spawn.sh tests/unit/test-agent-lifecycle-events.sh; bash tests/unit/test-agent-lifecycle-events.sh → 8/8 PASS; git diff --check. @coderabbitai review

PR #522, opened 2026-06-16
feat(providers): remove Gemini provider, Antigravity (agy) is the sole Google seat

L3ulll:
## Summary

The **Gemini CLI is sunset 2026-06-18**, so this removes its provider integration outright rather than carrying it as dead weight. **Antigravity (`agy`)** — a first-class provider since v9.45.0 — becomes the sole Google-family seat.

This PR is the **functional removal** (verified, tested). A follow-up will scrub the remaining cosmetic mentions (docs, command banners, comments, CHANGELOG history) — see *Scope* below.

## Removed

- `scripts/helpers/gemini-exec.sh`, `config/providers/gemini/`, `.gemini/commands/octo/`, and the gemini provider test suite (`tests/unit/test-gemini-provider.sh`)
- The gemini dispatch arm, provider detection/health, model-catalog rows, `AVAILABLE_AGENTS` entries, fleet/reviewer slots, task-routing, and model-resolver fallbacks

## Antigravity takes every Google-seat slot

- Named workflows (`grasp`/`tangle`/`ink`/`embrace` gate) dispatch `agy`
- `build-fleet` uses `agy` as the security/CVE reviewer (caught by the allowlist unit test)
- `embrace` strategy + the circuit breaker include `agy`
- `agents.sh` routes design/research/copywriting/image tasks to `agy`

## Antigravity adapter hardening (`scripts/helpers/agy-exec.sh`)

Safe parity only:
- `OCTOPUS_AGY_MODEL` passthrough — **default = no `--model`**, so agy keeps the model picked in its own UI and isn't forced onto an exhaustible quota group
- `OCTOPUS_AGY_SANDBOX=off` toggle, `OCTOPUS_AGY_INCLUDE_DIRS` → `--add-dir`
- one provider-agnostic replay-from-stdin retry on a silent-empty response

Deliberately **no** model-fallback chain or error classifier: agy's model catalog and error strings are not verified in-band (and forcing a model is harmful per the adapter's own design note).

## Verification

- **Live `agy` dispatch** through the production `agy-exec.sh` adapter returns correct output (exit 0).
- Unit suite green: `agy-provider` 31, `provider-allowlist` 8, `fleet-diversity` 36, `detect-providers` 3, `provider-contract-audit` 4, `research-fanout` 5; all `.sh` parse. (Pre-existing unrelated red: a debate dry-run test needs a git submodule initialized.)

## Scope

This is the **functional** removal — gemini is no longer dispatchable and `agy` works end-to-end. Remaining cosmetic gemini references (≈140 docs, ≈88 command banners, inline comments, historical CHANGELOG entries) are intentionally left for a focused follow-up so this PR stays reviewable. Net so far: **−1035 lines**.

## Summary by CodeRabbit

* **Deprecations**
  * Removed Gemini CLI provider support; Antigravity (`agy`) is now the only Google-family provider option.

* **New Features**
  * Enhanced Antigravity CLI adapter with buffered I/O, model pin/configurable include-directories support, and safer retry behavior for unexpectedly empty outputs.

* **Bug Fixes**
  * Improved stability across provider routing and workflow execution, including more consistent provider selection and status reporting.

* **Documentation**
  * Updated setup guides, workflow and provider tables, and cost/model-pinning guidance to reflect `agy` instead of Gemini.

* **Tests**
  * Updated unit coverage for `agy` routing, model validation, and include-directory behavior.


Jhacarreiro:
I think this PR is solving the right cost/routing problem at the wrong abstraction layer.

The problematic default is "research routes select Gemini API by accident", not "Gemini must stop existing as a provider".

Removing the Gemini provider entirely conflates three different things:

1. **Gemini CLI sunset** — this may justify no longer relying on the old CLI path.
2. **Gemini API provider** — this can still be a valid explicit provider for users with an API key, free-tier setup, or deliberate Google API usage.
3. **Antigravity / `agy`** — this is the better default Google-family seat for Octopus workflows because it is CLI/subscription-bundled and already used by tangle/research paths.

I agree with the direction that default Google-family routing should prefer `agy`. I do not think we should remove `gemini` as an explicit provider. That is a broader product-level breaking change, not a cost-control fix.

A narrower fix is safer:

- keep `gemini` available for explicit `gemini:*` usage;
- change research/researcher defaults to `agy`;
- preserve legacy mode for users who intentionally want old behavior;
- leave Codex/Claude/Opus/OpenRouter defaults untouched.

That is what #569 does as a minimal alternative. It stops accidental Gemini API spend without removing Gemini for users who still want it.

So my recommendation is: do not merge a full provider removal here. Prefer the smaller routing/defaults change, then revisit actual Gemini provider removal only if there is a separate consensus that explicit Gemini API support should no longer be part of Octopus.

nyldn:
Closing this stale fork PR in favor of the maintained feat/remove-dead-gemini-provider branch now pushed to nyldn/claude-octopus. The branch keeps the provider-removal work available without carrying the outdated review state on #522.

Issue #117, opened 2026-03-07
[BUG] Codex agent 401 auth failure due to literal quotes in HOME path


Adashi12:
# Bug Report: Codex agent 401 auth failure due to literal quotes in HOME path

## Description

`build_provider_env()` (line 1151 of `orchestrate.sh`) outputs environment variable assignments with escaped double quotes (`\"`). When this string is later consumed by `read -ra cmd_array` (line 11064), the quotes become **literal characters** in the values rather than being stripped as shell quoting. This causes `HOME` to be set to `"/home/user"` (with literal quote characters) instead of `/home/user`, which prevents Codex from finding `~/.codex/auth.json` and results in persistent 401 Unauthorized errors.

The smoke test is unaffected because `_smoke_test_provider()` calls `get_agent_command()` directly without going through `build_provider_env()`, so the bug is masked during pre-flight checks.

## Steps to Reproduce

1. Configure Codex CLI with ChatGPT auth (no `OPENAI_API_KEY` set)
2. Set `OCTOPUS_CODEX_MODEL=gpt-5.3-codex` (or any ChatGPT-compatible model)
3. Run any orchestrated workflow:
   ```bash
   bash orchestrate.sh embrace "Create hello.py"
   ```
4. Smoke test passes, but all Codex agent tasks fail with 401

**Minimal reproduction of the quoting bug:**

```bash
# This is what orchestrate.sh does internally:
env_prefix="env -i PATH=\"$PATH\" HOME=\"$HOME\" OPENAI_API_KEY=\"\" TMPDIR=\"/tmp\""
cmd="codex exec --model gpt-5.3-codex --sandbox workspace-write"
read -ra cmd_array <<< "$env_prefix $cmd"

# HOME now contains literal quotes:
echo "${cmd_array[3]}"
# Output: HOME="/home/user"  (quotes are literal characters)

# Codex fails because it resolves ~/.codex relative to "/home/user" (with quotes):
"${cmd_array[@]}" "Reply with ok"
# sandbox line shows: /home/user/project/"/home/user"/.codex/memories
# 401 Unauthorized -- auth.json not found
```

## Expected Behavior

Codex agents spawned by `orchestrate.sh` should authenticate successfully using ChatGPT auth (reading `~/.codex/auth.json`) and execute tasks, just as they do when invoked directly via `codex exec`.

## Actual Behavior

[...]

PR #533, opened 2026-06-23
fix(providers): build provider env as an array, not a word-split string

spareilleux:
## Problem

`build_provider_env()` (scripts/lib/provider-routing.sh) returns the credential-isolation prefix as a **space-joined string**:

```sh
echo "env -i PATH=$PATH HOME=$HOME OPENAI_API_KEY=${OPENAI_API_KEY:-} TMPDIR=${TMPDIR:-/tmp}"
```

Both call sites (`spawn.sh`, `workflows.sh`) then consume it with:

```sh
read -ra cmd_array <<< "$env_prefix $cmd"
```

`read -ra` **word-splits on spaces**, so any env value containing a space breaks. The common trigger is `PATH` on Windows (Git Bash / WSL), where an entry like `C:\Program Files\NVIDIA GPU Computing Toolkit\...` shatters:

```
env: 'Files/NVIDIA': No such file or directory   # exit 127
```

Every external provider (codex / gemini / perplexity) then **silently fails**, while the run still reports (fabricated) cost/token estimates. This is the same class of bug as the literal-quote regression in #117 — a flat string simply can't carry values with spaces or quotes through `read -ra`.

### Repro
```sh
PATH_WS="/c/Program Files/NVIDIA GPU Computing Toolkit/bin:/usr/bin:/bin"
read -ra a <<< "env -i PATH=$PATH_WS HOME=/h OPENAI_API_KEY=x printenv PATH"
"${a[@]}"   # -> env: 'Files/NVIDIA': No such file or directory (127)
```

## Fix

Build the prefix as an **array** instead of a string:

- `build_provider_env()` populates a global `OCTO_PROVIDER_ENV=(env -i "PATH=$PATH" "HOME=$HOME" ...)` (one line per provider, so existing static-grep tests still match the `env -i …KEY…` line).
- `spawn.sh` / `workflows.sh` consume `"${OCTO_PROVIDER_ENV[@]}"`.

Array elements preserve spaces and quotes, so the spaced-PATH case works and the #117 literal-quote class is structurally avoided.

```sh
new=(env -i "PATH=$PATH_WS" "HOME=/h" "OPENAI_API_KEY=x" printenv PATH)
"${new[@]}"   # -> prints the full spaced PATH, exit 0
```

## Tests

`tests/test-credential-isolation.sh` → **19/19 passing** (the per-provider key scoping and the #117 "no escaped quotes" checks are preserved).

Found while wiring claude-octopus into an away-from-keyboard agent loop on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed provider execution failures caused by spaces in environment paths or configuration values. Enhanced the reliability of provider credential isolation and environment variable handling to ensure consistent performance across different system setups and configurations.

### DIFF
--- a/scripts/lib/provider-routing.sh
+++ b/scripts/lib/provider-routing.sh
@@ -28,6 +28,15 @@
 build_provider_env() {
     local provider="$1"
 
+    # Return the env prefix as a global ARRAY (OCTO_PROVIDER_ENV), not a
+    # space-joined string. The string form was consumed via `read -ra`, which
+    # word-splits on spaces — so a PATH containing spaces (e.g. Windows
+    # "C:\Program Files\NVIDIA GPU Computing Toolkit\...") shattered and `env`
+    # tried to exec a fragment like 'Files/NVIDIA' (exit 127), silently failing
+    # every external provider. Array elements preserve spaces and quotes (this
+    # also removes the class of bug behind the literal-quote regression #117).
+    OCTO_PROVIDER_ENV=()
+
     if [[ "${OCTOPUS_SECURITY_V870:-true}" != "true" ]]; then
         return 0
     fi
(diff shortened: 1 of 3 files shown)


nyldn:
Approved and still wanted (correct fix: provider env must be built as an array, not a word-split string, so spaced paths survive). It now conflicts with main after the v9.46.0 merges. Please rebase `fix/provider-env-spaced-path` onto current main and re-push; once the 3 required checks are green on an up-to-date branch it can merge. Targets the next release.

nyldn:
Superseded by work already on main. The fix here (build the provider env as an array instead of a word-split string, so spaced $PATH/$HOME survive) is implemented on current `main` via `PROVIDER_ENV_ARRAY=(env -i …)` throughout `scripts/lib/provider-routing.sh` — with more coverage than this PR (Codex `config.toml` env_key handling, gemini trust/cloud-project vars, trace env). No word-split env-string construction remains on main. Rebasing this PR would regress that. Thanks @spareilleux for catching the spaced-path bug — the array approach you advocated is what shipped. Closing as superseded.

PR #539, opened 2026-06-24
fix: migrate parallel.sh map_reduce/fan_out off the sunset Gemini CLI (#524 follow-up)

Adam2kg:
## Summary

Completes the agy migration (#524, "agy = default Google seat", Gemini CLI sunset 2026-06-18) in the parallel primitives. Three Gemini-CLI-era dispatch points remained in `scripts/lib/parallel.sh` that break or skip agy when the gemini binary is absent:

1. **`map_reduce()` decomposition** hard-coded `gemini "$decompose_prompt" > "$decompose_result"` — when gemini isn't installed this failed outright and degraded to a plain `fan_out`.
2. **`fan_out()`** defaulted its second dispatch seat to `gemini` (`agents=("codex" "gemini")`).
3. **`map_reduce()`** rotated subtasks over the same hard-coded `("codex" "gemini")` pair.

## Changes

- **New `_parallel_google_seat()` resolver** — `agy → gemini (legacy, only when still installed) → claude-sonnet`. This *prefers agy* (the Google seat) while *preserving behavior for setups that still have the Gemini CLI*, with a guaranteed claude-sonnet fallback.
- **Decomposition** now routes through the agy-capable `run_agent_sync … "decompose" "parallel"` abstraction (agy → claude-sonnet fallback → `fan_out` as last resort), mirroring how the synthesis path is being migrated. Guarded on `type run_agent_sync`; "Phase 1 … with Gemini" log relabeled.
- **Default pairs** in `fan_out()` and `map_reduce()` now use `("codex" "$(_parallel_google_seat)")`.

## Test — `tests/unit/test-agy-parallel-decompose.sh` (8/8)

Static assertions (no bare `gemini "$decompose_prompt"` remains; routes via `run_agent_sync` + `_parallel_google_seat`; default pairs use the resolver) **plus** hermetic functional tests:
- seat preference order: agy preferred; gemini kept when agy absent; claude-sonnet when neither;
- `map_reduce` decomposes via `run_agent_sync(agy)` **not** the bare gemini CLI, and seats agy in the subtask rotation;
- `fan_out`'s default pair resolves to `(codex, agy)` when agy is available and gemini is absent.

## Verification

- `bash -n` clean; new test 8/8.
- Regression: `test-routing-features-consumers` (15/15), `test-output-guard` (6/6), `test-agy-provider` (31/31), `test-agy-workflow-routing` (4/4).

## Notes for reviewers

- Scoped to decomposition + default dispatch pairs. The `aggregate_results()` **synthesis** path is migrated separately in #538; this branch is cut from `main`, so the two touch disjoint functions/lines and won't conflict. On this branch the synthesis path still shows the old gemini call (that's #538's fix).
- `_parallel_google_seat` is intentionally distinct from #538's `_aggregate_pick_synth_agent` (this resolver keeps gemini when still installed; the synthesis picker drops it). The two can be unified once both land.

## Summary by CodeRabbit

* **New Features**
  * Enhanced task dispatch for fan-out and map-reduce to prefer a newer Google dispatch seat when available, with automatic fallback to the best remaining option.

* **Bug Fixes**
  * Improved decomposition routing to avoid Gemini-only fallback behavior and ensure more reliable agent selection when certain tools are missing.

* **Tests**
  * Added regression tests to verify seat-precedence rules and confirm decomposition uses the expected agent path, plus coverage for fan-out and map-reduce defaults.

### DIFF
--- a/scripts/lib/parallel.sh
+++ b/scripts/lib/parallel.sh
@@ -33,6 +33,22 @@ _fan_out_agents_from_config() {
     ' "$config_file" 2>/dev/null || true
 }
 
+# Resolve the preferred non-Codex dispatch seat for fan-out / map-reduce.
+# The Gemini CLI was sunset 2026-06-18; agy (Antigravity) is the default Google
+# seat (#524). Order: agy → gemini (legacy, only when still installed, to
+# preserve behavior for existing setups) → claude-sonnet (always available in
+# Claude Code). Always echoes a usable agent.
+_parallel_google_seat() {
+    if { ! declare -f octo_provider_allowed >/dev/null 2>&1 || octo_provider_allowed agy; } \
+        && command -v agy >/dev/null 2>&1; then
+        echo "agy"; return 0
+    fi
+    if command -v gemini >/dev/null 2>&1; then
+        echo "gemini"; return 0
+    fi
+    echo "claude-sonnet"; return 0
+}
+
 fan_out() {
     local prompt="$1"
     local agents=()
@@ -55,8 +71,9 @@ fan_out() {
         done <<< "$_configured"
     fi
 
-    # Fallback to original default pair when config absent or all entries invalid
-    [[ ${#agents[@]} -eq 0 ]] && agents=("codex" "gemini")
+    # Fallback to default pair when config absent or all entries invalid.
+    # Second seat prefers agy (Google seat) over the sunset Gemini CLI (#524).
+    [[ ${#agents[@]} -eq 0 ]] && agents=("codex" "$(_parallel_google_seat)")
 
     log INFO "Fan-out: Sending prompt to ${#agents[@]} agents (${agents[*]})"
     echo ""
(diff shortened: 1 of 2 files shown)


Adam2kg:
Addressed CodeRabbit review in `caafbb4`:

- **Rotation assertion scope:** The subtask-rotation check now inspects only the `=== SPAWN ===` section, so the `DECOMPOSE_AGENT=agy` decompose-log line can no longer satisfy the `*"agy"*` pair assertion. It asserts `codex`+`agy` (and not `gemini`) against the spawn output specifically.
- **Test strictness:** `set -euo pipefail` in all shell contexts + temp dirs under the framework's `TEST_TMP_DIR`. Added `sleep` to the hermetic PATH so `fan_out`'s inter-spawn `sleep` doesn't trip the now-strict `-e` harness.
- Made `_parallel_google_seat` allowlist-aware for the `gemini` fallback too (agy was already gated), so the resolver honors `octo_provider_allowed` consistently.

Note: the central `google`→agy allowlist alias normalization lands in #538 (`provider-allowlist.sh`); kept here to avoid both PRs editing the same file. This branch keeps gemini/claude fallbacks so it isn't blocked by ordering.

PR #560, opened 2026-07-02
fix: reconcile late tangle completions

Jhacarreiro:
## Summary

This fixes a provider-agnostic tangle false-negative where a wrapper PID exits before the actual CLI/provider finishes writing its result and `.done` marker.

Observed failure mode:

- `spawn_agent_capture_pid` cannot always discover a real provider PID and falls back to tracking the wrapper PID.
- The wrapper can disappear before the provider/result writer has finished.
- The tangle watcher immediately writes `missing-done-marker` and the quality gate can abort at 2/3 even though the late result later writes `## Status: SUCCESS` and `.done=0`.

## What changed

- Adds `OCTOPUS_TANGLE_MISSING_MARKER_GRACE` (default 180s) before converting a wrapper-exit/no-marker condition into a terminal missing-marker failure.
- Adds a final artifact reconciliation pass before reporting failures / entering the quality gate: if the latest result status for a task is `SUCCESS`, write/repair its `.done` marker to `0`.
- Makes tangle validation count the **latest** `## Status:` block rather than treating any historical `Status: SUCCESS` as success.
- Adds regression tests for latest-status semantics and the late-completion reconciliation wiring.

## Why this is provider-agnostic

This is not specific to `agy`; any CLI-backed provider can hit this if wrapper lifetime and provider/output lifetime diverge. The fix is therefore in the tangle watcher / validation path rather than in provider-specific routing.

## Validation

```text
bash -n scripts/lib/workflows.sh scripts/lib/testing.sh tests/unit/test-tangle-late-completion-reconcile.sh
bash tests/unit/test-tangle-late-completion-reconcile.sh
bash tests/unit/test-quality-retry-env-and-unlimited.sh
bash tests/smoke/test-syntax.sh
```

## Summary by CodeRabbit

* **Tests**
  * Updated the tangle missing-done-marker regression test to use a stricter grace setting and a broader timing window for expected log output.
  * Adjusted a scenario to tolerate a non-critical failure without aborting the run.

* **Improvements**
  * Enhanced tangle execution monitoring to better handle late completion markers by waiting a configurable grace period before treating them as missing.
  * Improved result validation to rely on each result’s latest reported status for more accurate success/failure decisions.

### DIFF
--- a/scripts/lib/testing.sh
+++ b/scripts/lib/testing.sh
@@ -103,6 +103,17 @@ check_tangle_worktree_changes() {
     rm -f "$current_file"
 }
 
+tangle_result_latest_status() {
+    local result="$1"
+    local status_line=""
+    status_line=$(grep '^## Status:' "$result" 2>/dev/null | tail -1 || true)
+    case "$status_line" in
+        *SUCCESS*) echo "success" ;;
+        *FAILED*|*TIMEOUT*|*ERROR*) echo "failed" ;;
+        *) echo "unknown" ;;
+    esac
+}
+
 validate_tangle_results() {
     local task_group="$1"
     local original_prompt="$2"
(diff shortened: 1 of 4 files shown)


nyldn:
This now conflicts with main after its sibling tangle PRs merged (#571 harden tangle dispatch, #572 retry hard-gate failures), which touch the same `testing.sh`/`workflows.sh` regions. Please rebase `reconcile late tangle completions` onto current main and reconcile against #571/#572 — some of the reconciliation logic here may now overlap with what landed. Re-push and it can merge once green.

PR #565, opened 2026-07-02
fix(agy): record agy's actual selected model instead of "default"

klass-723:
## Problem

`agy-exec.sh` runs `agy --print` with `--model default` (no `--model` flag) unless `OCTOPUS_AGY_MODEL` is set — so agy uses whatever model is selected in its own `/model` UI. As a result, the council roster artifact (`summary.json`) and the preflight banner record the opaque string **`"default"`** for the agy seat.

That makes a `Codex + agy` panel's **cross-lab-vs-same-lineage** status unverifiable from the artifact: whether agy ran a Gemini model (cross-lab with Codex), a Claude model, or GPT-OSS (same-lineage as Codex) can't be told from `model: "default"` — you have to know agy's current selection out-of-band.

## Fix

Add `agy_current_model()` to `lib/providers.sh` (sourced before `preflight.sh` and `council.sh` by `orchestrate.sh`). Resolution mirrors `agy-exec`:

1. `OCTOPUS_AGY_MODEL` if set and not `default`, else
2. read `.model` from `~/.gemini/antigravity-cli/settings.json` (where the CLI persists the `/model` selection), else
3. fail-safe `"default (unresolved)"`.

Wired into two places (both guarded with `declare -f` so standalone runs fall back to prior behavior — purely diagnostic, never gate logic):

- **`council.sh` `council_roster_entry_json`** — the agy seat records its real model (e.g. `Gemini 3.1 Pro (Low)`) in `summary.json`.
- **`preflight.sh`** — the provider-cache `AGY_MODEL` and the activation banner show the real model instead of empty / "agy default".

## Verification

- `bash -n` clean on all three files.
- Happy path → `Gemini 3.1 Pro (Low)`; `OCTOPUS_AGY_MODEL` override → wins; missing `settings.json` → `default (unresolved)`.
- Source order `providers.sh → council.sh` keeps the helper defined.

3 files, +41/−1. No behavior change to dispatch — only what gets recorded/displayed.

## Summary by CodeRabbit

* **New Features**
  * Improved model detection for the Antigravity provider so the active model is now shown more accurately in generated roster and cache data.
  * Added a fallback when the model can’t be determined, helping keep output readable instead of incomplete.

* **Bug Fixes**
  * Fixed cases where a default or placeholder model could be used instead of the currently selected model.

### DIFF
--- a/scripts/lib/council.sh
+++ b/scripts/lib/council.sh
@@ -917,6 +917,12 @@ council_roster_entry_json() {
     [[ -n "$provider" ]] || provider="$(council_pick_provider "$preferred_provider")"
     provider_org="$(council_provider_org "$provider")"
     model="$(council_persona_model "$persona")"
+    # agy ignores the per-persona model (agy-exec runs `--model default`), so record
+    # the model agy will ACTUALLY use — resolved from its own settings — instead of
+    # the placeholder, so the seat's cross-lab lineage is verifiable from the artifact.
+    if [[ "$provider" == "agy" ]] && declare -f agy_current_model >/dev/null 2>&1; then
+        model="$(agy_current_model)"
+    fi
     seat="$(council_persona_seat "$persona")"
     family="$(council_persona_family "$persona")"
     permission_mode="$(council_agent_config_value "$persona" "permissionMode" | tr -d '"')"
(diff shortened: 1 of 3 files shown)


nyldn:
Closing as an exact duplicate of #585 (same fix, same author, freshly rebased with CHANGELOG entry). #585 merged.

PR #577, opened 2026-07-05
fix(council): distinct-approving-vendor quorum (substantive + verdict-gated)

klass-723:
## Summary

The council advice-phase quorum has no distinct-vendor guard on `main` — `quorum.met` for `standard`/`deep` is true whenever `received_non_chair >= required`, and a seat is counted on **dispatch exit code alone**. Two failure modes slip through and were observed in production (sail-cruisey #2002/#2007/#2003, and the 3×-codex case #1993):

1. **Single-vendor "consensus."** Three codex seats (agy/gemini returned empty) all count → `met:true`, even though one vendor can't catch its own blind spot.
2. **Degenerate seats that exit 0 while reviewing nothing.** The host self-dispatch stub (`"active host runtime … subprocess dispatch is unavailable"`, ~604B) and empty/~1B provider returns are counted, producing false `met:true` with zero real reviews.

## Change

- Record a responding seat's provider only when its response is **non-empty** (`council_response_nonempty`) **and substantive** (`council_response_is_substantive` — rejects the host stub and short "cannot access the plan/files" degenerate reviews; brevity-gated at 1600 non-whitespace chars so a long, genuine review that merely quotes such a phrase still counts).
- Gate-depth councils (`required >= 2`, i.e. `standard`/`deep`) now require **≥2 distinct providers**; a single-vendor pass fails with a stderr warning to restore a 2nd provider or surface the provider-shortage gate.
- `summary.json` gains `distinct_providers` + `responding_providers` for transparency.
- `quick` depth (required 1) is unchanged.

The warning uses council.sh's own `>&2` convention (council.sh defines no `log()` of its own — that lives in `orchestrate.sh`), so the guard never crashes when the library is sourced standalone (e.g. under the unit tests).

## Reproduction (before this PR)

A `deep` run whose non-chair voting seats all return empty/stub content reports `quorum.met: true` with `distinct_providers` counting the degenerate seats. After this PR the same run reports `met: false` and names the shortage.

## Test plan

- [x] `tests/unit/test-council-command.sh` — **51/51 pass**
- [x] `bash -n scripts/lib/council.sh` — clean
- [x] Verified the content check excludes: empty, whitespace-only, the host self-dispatch stub, and short "cannot access the files" responses; and includes genuine long reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected council quorum “met” logic to count only **distinct approving vendors** using **substantive** seat verdicts with no dissent, with mixed/degenerate outputs handled more strictly.
  * Updated council advice quorum (standard/deep) to require **at least two distinct substantive providers** with non-empty responses; quick depth unchanged.
  * Expanded quorum reporting in `summary.json` with distinct **responding** and **approving** provider counts.
* **Tests**
  * Added unit tests covering verdict parsing, approving-provider failsafes, split double-seat quorum behavior, and the all-approve quorum case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- **Council quorum now gates on distinct APPROVING vendors, not just distinct responders.** Each non-chair seat's response must end with `VERDICT: APPROVE|REVISE|BLOCK`; the runner reads the last such line (missing/ambiguous → REVISE, fail-safe). A vendor counts toward quorum only if it responded substantively **and** none of its seats dissented, so a split double-seated vendor (one seat APPROVE, one REVISE) can no longer cherry-pick its approving seat into a passing quorum. Standard/deep now require ≥2 distinct approving vendors; `summary.json` adds `distinct_approving_providers` + `approving_providers`. Fixes false `met:true` in the 2-vendor era (sail-cruisey #1992/#1994/#1983). Quick depth (required 1) is unchanged. Layers on the distinct-responder/substantive guard below.
+
 - **Council advice quorum now requires ≥2 DISTINCT providers with substantive responses.** Previously `quorum.met` for `standard`/`deep` depth was true as long as `received_non_chair >= required`, counting a seat on dispatch exit code alone — so a single-vendor result (e.g. 3× codex because agy/gemini returned empty) and even seats that exit 0 while reviewing nothing (the host self-dispatch stub, empty/~1B provider returns) all counted, producing false `met:true`. Now each responding seat's provider is recorded only when its response is non-empty **and** substantive (rejecting the host stub and short "cannot access the files" degenerate reviews, brevity-gated so long real reviews pass); gate-depth councils require ≥2 distinct providers, and `summary.json` reports `distinct_providers` + `responding_providers`. `quick` depth (required 1) is unchanged.
 
 ## [9.47.1] - 2026-07-02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


klass-723:
Both CodeRabbit findings addressed in `145509c`:
- `council_response_is_substantive`: `grep -qiE` → `grep -ciE … >/dev/null` (per the project's no-`grep -q`-under-`set -eo pipefail` convention; behavior-preserving — exit 0 on match, 1 on none).
- `summary.json`: guarded `${COUNCIL_RESPONDING_PROVIDERS…}` with `:+` so the dry-run / early-cost-abort paths don't trip `set -u` before `council_run_advice_phase` initializes it (mirrors the `:-0` on `distinct_providers`).

`test-council-command.sh`: 51/51 pass.

klass-723:
Expanded this PR (commit \`657664f\`) to also close the **split double-seated vendor** false-quorum (sail-cruisey #1992/#1994/#1983), since it's the same quorum-guard lineage and builds directly on the substantive-response helpers here (a standalone PR off \`main\` wouldn't compile without them).

**New layer — gate on distinct APPROVERS, not just responders:**
- Non-chair advice prompt now requires a final `VERDICT: APPROVE|REVISE|BLOCK` line. `council_response_verdict` reads the **last** such line; missing/ambiguous → REVISE (fail-safe).
- `council_compute_approving_providers`: a vendor counts only if it responded substantively **and** none of its seats dissented — so a vendor double-seated across a split (one APPROVE, one REVISE) is dropped, not cherry-picked.
- Standard/deep now require ≥2 **distinct approving** vendors. Quick (required 1) unchanged.
- `summary.json` adds `distinct_approving_providers` + `approving_providers`.

**Tests (now 55/55):** verdict parsing (incl. last-line-wins + fail-safe), the approver fail-safe unit, an end-to-end #1992 split reproduction (`met:false`, `approving="agy"`), and clean all-approve (`met:true`).

Deliberately **not** included: the "don't double-seat in the K-vendor era" roster change (report item 1). With approver-gating the double-seat is already harmless to the pass/fail outcome, so that's a separate roster-hygiene follow-up rather than bundling a roster rewrite into this guard.

Happy to split the approver layer back into its own stacked PR if you'd prefer to review the substantive-guard and approver-guard separately — say the word.

PR #578, opened 2026-07-05
fix: let role model routes override phase routes

Jhacarreiro:
## Summary

Fix model routing precedence so role-specific routes are applied before broad phase routes.

This matters for review fleets: a broad `review -> codex:default` phase route currently prevents a more specific role route such as `logic-reviewer -> codex:logic_review` from taking effect. The result is that specialist reviewer roles silently inherit the phase default model.

## Changes

- Prefer `routing.roles[$role]` before `routing.phases[$phase]` in `resolve_octopus_model`.
- Keep the existing fallback behavior: if no role route exists, phase routing still applies.
- Add regression tests showing:
  - role routing overrides review phase routing for the matching provider;
  - provider-scoped role routes do not leak Codex models into Claude/Gemini fallback reviewers;
  - review phase routing continues to apply when no role override exists.

## Tests

```bash
bash -n scripts/lib/model-resolver.sh
bash tests/test-resolve-model.sh
bash tests/unit/test-orchestrate-cwd-routing.sh
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated model routing so provider-scoped role overrides take priority over phase-based routing when resolving an unset model.
  * Added safer fallback behavior when a role override targets a different provider, preventing cross-provider routing leakage.
* **Tests**
  * Added new assertions covering role-vs-phase routing priority and provider isolation, including cases where role overrides aren’t valid for the current provider.
  * Renumbered the existing session override test header for consistency.

### DIFF
--- a/scripts/lib/model-resolver.sh
+++ b/scripts/lib/model-resolver.sh
@@ -102,6 +102,15 @@ validate_model_name_for_provider() {
             ;;
     esac
 }
+
+_octo_is_known_provider_name() {
+    case "$1" in
+        codex|gemini|claude|perplexity|qwen|copilot|opencode|ollama|openrouter|cursor-agent|vibe|agy|agy-research|antigravity)
+            return 0 ;;
+        *)
+            return 1 ;;
+    esac
+}
 # resolve_octopus_model <provider> <agent_type> <phase> <role>
 resolve_octopus_model() {
     local provider="$1"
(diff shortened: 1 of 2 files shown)


Jhacarreiro:
Addressed CodeRabbit follow-up: cross-provider role routes now fall back to phase routes, and unscoped/bare role routes do not override an available phase route. Validation: bash -n scripts/lib/model-resolver.sh tests/test-resolve-model.sh; bash tests/test-resolve-model.sh → 14 PASS / 0 fail; git diff --check. @coderabbitai review

Jhacarreiro:
Addressed latest CodeRabbit maintainability follow-up: known bare provider names are now centralized in _octo_is_known_provider_name and reused in both role-route and route handling paths. Validation: bash -n scripts/lib/model-resolver.sh tests/test-resolve-model.sh; bash tests/test-resolve-model.sh → 14 PASS / 0 fail; git diff --check. @coderabbitai review

PR #605, opened 2026-07-09
fix: release.sh remote/branch assumptions and missing plugin-manifest.json bump

### LINKED ISSUE #603: release.sh incompatible with the documented release flow
RELEASING.md §2 (as of #601) says to run `scripts/release.sh`, but the script has three gaps vs the documented worktree/branch flow:

1. **Assumes `origin` is the canonical repo** (`git pull origin main`, `git push origin`): in a dev clone where origin is a fork, it pushes the release branch to the wrong repo.
2. **Requires being on `main`**: RELEASING.md §0 says to work in a worktree on a release branch cut from main; the script exits with "must be on main branch".
3. **`git add` list misses `.claude-plugin/plugin-manifest.json` and `.claude-plugin/routines.json`** even though both carry version strings (the routines sed was added in #601 but the file is not staged, so the bump is silently left uncommitted).

v9.52.0 was cut manually as a result. Either fix the script to take a remote/branch and stage everything it touches, or revert RELEASING.md §2 to lead with the manual table.

nyldn:
## Description
`scripts/release.sh` had three gaps versus the release flow documented in `RELEASING.md` (added in #601), all confirmed by reading the script directly:

1. **Hardcoded `origin`**: `git pull origin main` (release.sh:57) and `git push -u origin "$BRANCH"` (release.sh:207, pre-fix) assume `origin` is the canonical repo. In a dev clone where `origin` is a fork, this pulls/pushes against the wrong remote. Now configurable via `OCTO_RELEASE_REMOTE` (defaults to `origin`, so no behavior change for the common case).
2. **Required `main` branch**: release.sh:52-55 exited with "must be on main branch", but `RELEASING.md` §0 documents cutting the release branch in a separate worktree (`git worktree add <dir> -b release/<name> origin/main`) and running the script already checked out on it. The script now accepts either `main` or the target `release/vX.Y.Z` branch, and skips the redundant `git pull`/`git checkout -b` when already on the release branch.
3. **`.claude-plugin/plugin-manifest.json` never touched**: `RELEASING.md` §2 lists it as needing `version` + component counts, and the file's own `$comment` says it "must stay version-locked" with `plugin.json` — but release.sh never read or wrote it, and it wasn't in the `git add` list, so the bump was silently skipped every release (v9.52.0 was cut manually because of this). The version-bump step now updates `version`, `components.commands.count`, `components.skills.count`, and `components.agents.count`/`breakdown` (personas + droids), and stages the file.

## Type of Change
- [x] Bug fix

## Testing
- `bash -n scripts/*.sh scripts/lib/*.sh` — passes.
- No existing test targets `release.sh` directly. Extracted the modified "1/8 Updating version files" step and ran it standalone against a scratch copy of the actual repo tree (VERSION=9.51.1): confirmed `.claude-plugin/plugin-manifest.json` gets `version: "9.51.1"` and correct counts (`commands: 50, skills: 58, agents: {count: 42, breakdown: {personas: 32, droids: 10}}`), and that the resulting JSON is valid.
- `git diff origin/main...HEAD --summary | grep "mode change"` — empty, no exec-bit drift.
- Did not run the interactive PR/CI/merge portion of the script (steps 3-8) against this repo — that would create real branches/PRs/releases as a side effect of testing, which is out of scope for this fix.
- `make test-unit` was started but exceeded the time budget available in this session before completing; the change is confined to `scripts/release.sh`, which has no existing unit coverage, so the targeted verification above is the meaningful test surface for this diff.

## Related Issues
Closes #603

---
_Generated by [Claude Code](https://claude.ai/code/session_01V4gMacU1AHE3tpCkYgqboE)_

## Summary by CodeRabbit

* **New Features**
  * Release automation can run from `main` or directly from an existing release branch.
  * Release syncing now supports a configurable git remote, and GitHub PR/checks/release actions are scoped to it.
  * Plugin manifest updates now refresh release version and component/persona/droid/agent counts and breakdowns.
* **Bug Fixes**
  * Improved merge-and-release behavior when already on the release branch by targeting the release tag to the fetched `main` SHA.
  * Avoids redundant release branch creation and updates remote usage for push/pull operations.
* **Tests**
  * Added regression coverage for worktree/branch/remote flow, staging of the plugin manifest, and SHA targeting for release creation.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -43,6 +43,11 @@ REMOTE="${OCTO_RELEASE_REMOTE:-origin}"
 
 cd "$PLUGIN_ROOT"
 
+# gh infers the target repo from remotes independently of $REMOTE, so a dev
+# clone pointing $REMOTE at the canonical repo (with origin left on a fork)
+# would otherwise create the PR/release against the wrong repo. Pin it.
+REPO_SLUG="$(git remote get-url "$REMOTE" | sed -E 's#^(git@|https://)([^:/]+)[:/]##; s#\.git$##')"
+
 # --- Preflight ---
 
 if ! git diff --quiet 2>/dev/null; then
@@ -245,6 +250,8 @@ echo ""
 
 echo "4/8 Creating PR..."
 PR_URL=$(gh pr create \
+    -R "$REPO_SLUG" \
+    --head "$BRANCH" \
     --title "chore: release v${VERSION}" \
     --body "## Release v${VERSION}
 
@@ -263,7 +270,7 @@ echo "5/8 Waiting for CI..."
 # Poll until required checks finish (max 5 minutes)
 DEADLINE=$((SECONDS + 300))
 while [[ $SECONDS -lt $DEADLINE ]]; do
-    CHECKS=$(gh pr checks "$PR_NUM" --json name,state 2>&1 || true)
+    CHECKS=$(gh pr checks "$PR_NUM" -R "$REPO_SLUG" --json name,state 2>&1 || true)
     SMOKE=$(octo_pr_check_state "$CHECKS" "Smoke Tests")
     UNIT=$(octo_pr_check_state "$CHECKS" "Unit Tests")
     INTEG=$(octo_pr_check_state "$CHECKS" "Integration Tests")
@@ -275,7 +282,7 @@ while [[ $SECONDS -lt $DEADLINE ]]; do
 
     if [[ "$SMOKE" == "fail" || "$UNIT" == "fail" || "$INTEG" == "fail" ]]; then
         echo "   CI FAILED — Smoke: ${SMOKE} | Unit: ${UNIT} | Integration: ${INTEG}"
-        echo "   Fix failures, then run: gh pr merge ${PR_NUM} --merge"
+        echo "   Fix failures, then run: gh pr merge ${PR_NUM} --merge -R ${REPO_SLUG}"
         exit 1
     fi
 
@@ -284,8 +291,8 @@ done
 
 if [[ $SECONDS -ge $DEADLINE ]]; then
     echo "   CI timed out after 5 minutes."
-    echo "   Check manually: gh pr checks ${PR_NUM}"
-    echo "   Then merge: gh pr merge ${PR_NUM} --merge"
+    echo "   Check manually: gh pr checks ${PR_NUM} -R ${REPO_SLUG}"
+    echo "   Then merge: gh pr merge ${PR_NUM} --merge -R ${REPO_SLUG}"
     exit 1
 fi
 echo ""
@@ -293,7 +300,7 @@ echo ""
 # --- 6. Merge + Release ---
 
 echo "6/8 Merging and creating release..."
-gh pr merge "$PR_NUM" --merge --quiet 2>/dev/null || gh pr merge "$PR_NUM" --merge
+gh pr merge "$PR_NUM" -R "$REPO_SLUG" --merge --quiet 2>/dev/null || gh pr merge "$PR_NUM" -R "$REPO_SLUG" --merge
 
 if [[ "$ON_RELEASE_BRANCH" == "true" ]]; then
     # main is normally still checked out in the worktree this release branch
@@ -309,6 +316,7 @@ else
 fi
 
 gh release create "v${VERSION}" \
+    -R "$REPO_SLUG" \
     --target "$MERGE_SHA" \
     --title "v${VERSION} — ${SUMMARY}" \
     --notes "### Changed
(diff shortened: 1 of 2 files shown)


nyldn:
@coderabbitai the "Linked Issues" warning about `.claude-plugin/routines.json` not being staged is a false positive — that file was already being version-bumped (line 207-208, the `(vX.Y.Z)` sed on the `$comment` badge) and staged (line 223) before this PR, and this diff doesn't touch that logic. The three gaps this PR actually fixes are: hardcoded `origin` remote, the `main`-only branch check, and the missing `.claude-plugin/plugin-manifest.json` bump — all covered in the diff.

---

nyldn:
Blocking worktree-flow gap: the new preflight accepts running from an existing `release/vX.Y.Z` worktree, but step 6 still unconditionally does `git checkout main`, `git pull`, and `git branch -d "$BRANCH"`. In the documented `git worktree add ... -b release/... origin/main` flow, `main` is normally still checked out in the original worktree, so `git checkout main` here fails and the script exits before creating the tag/release.

Please keep the current checkout path for the ordinary `main` flow, but for `ON_RELEASE_BRANCH=true` fetch the merged `main` from `$REMOTE`, set `MERGE_SHA` from `FETCH_HEAD` (or the fetched remote ref), and skip deleting the locally checked-out release branch. Add a regression assertion for this branch-specific post-merge path; the current targeted verification covers only step 1.

nyldn:
Fixed in 11fd476. When `ON_RELEASE_BRANCH=true`, step 6 now does `git fetch --quiet "$REMOTE" main` and resolves `MERGE_SHA` from `FETCH_HEAD` instead of `git checkout main`/`git branch -d`, so it no longer touches this worktree's checkout or the branch it's standing on. `gh release create` now pins `--target "$MERGE_SHA"` explicitly in both paths.

Added `tests/unit/test-release-sh-worktree-flow.sh`: static assertions on the new logic, plus a functional test that builds the exact two-worktree scenario (main checked out in one worktree, `release/vX.Y.Z` cut in another, a merge landing on `origin/main`) and proves `git fetch` + `rev-parse FETCH_HEAD` resolves the correct SHA in that setup while a bare `git checkout main` fails there — confirming the fix actually addresses the failure mode described. All 6 assertions pass locally.

Re: the earlier CI failures on this SHA — `Unit Tests (ubuntu-latest)` failed on `test-gate-thresholds.sh`'s `validate_tangle_results uses get_gate_threshold` check (unrelated to this diff — no gate-threshold/orchestrate code touched). Reproduced the test's own grep pipeline against the repo 20/20 times locally with no failure, so it looks like CI flake, not a regression; `Unit Tests`/`Integration Tests` (the aggregate gates) failed only because they short-circuited on that one matrix leg. This push should get a clean run.

---

nyldn:
Both fixed:

- **release.sh (major)**: switched `components.get(key, {})` to `components.setdefault(key, {})` throughout the manifest-count block, so a missing `components`/`commands`/`skills`/`agents`/`breakdown` key now gets created and attached instead of silently dropping the write. Verified both the normal case (unchanged output) and the previously-buggy case: deleted `components` from a scratch copy of `plugin-manifest.json`, re-ran the version-bump step, and it now correctly repopulates the full structure instead of leaving stale counts.
- **test file (nitpick)**: added `set -euo pipefail`, matching the convention already used in `test-usage-report.sh`/`test-fable5-mode.sh`. Confirmed `test_pass`/`test_fail` never return non-zero, so this doesn't change test behavior — all 6 assertions still pass.

Also: the branch got rebased onto `main` underneath me (picking up the v9.52.0 release that landed while this PR was open — notably the release RELEASING.md says was cut manually because of the exact bug this PR fixes). Reconciled by cherry-picking the new fix commit onto the updated branch rather than force-pushing over it; pushed as a fast-forward.

---

nyldn:
Both fixed in 3457266:

- **release.sh (major)**: `gh` resolves its target repo from git remotes independently of `$REMOTE`, so a dev clone with `origin` on a fork and `OCTO_RELEASE_REMOTE` pointed at the canonical repo would push the branch correctly but create the PR/release against the wrong repo. Added `REPO_SLUG` resolved from `$REMOTE`'s URL (handles both `git@host:owner/repo.git` and `https://host/owner/repo(.git)`) and threaded `-R "$REPO_SLUG"` through every `gh pr create`/`gh pr checks`/`gh pr merge`/`gh release create` call, plus `--head "$BRANCH"` on `gh pr create`. Also updated the informational echo hints so a copy-pasted manual command targets the right repo too.

- **test file (minor)**: fixed — and this one's worth flagging explicitly: the two flagged assertions (`grep -A N ... | grep -q ...`, `grep -B2 ... | grep -q ...`) are the *exact* anti-pattern that's been causing the intermittent `test-gate-thresholds.sh` CI flake I've been re-running on this PR all session (SIGPIPE when the downstream `grep -q` exits early under `pipefail`). Rewrote both to capture the upstream grep's output into a variable and match with a here-string, removing the pipe entirely. Ran the suite 5x locally with no flakes.

---

nyldn:
Fixed in 65d2fc4. Switched to a single `trap 'rm -rf "$sandbox"' RETURN` right after creating the `$$`-suffixed sandbox dir, matching `test-framework.sh`'s own `TEST_TMP_DIR="/tmp/octopus-tests-$$"` convention, and dropped the three scattered `rm -rf` calls at each early-return path. Verified empirically that a RETURN trap set inside one function doesn't leak into later function calls in the same script, and ran the suite 3x locally — all pass, no sandbox directories left behind.

---

Issue #599, opened 2026-07-09
Council model selection: valid agy pin rejected (stdin read as prompt), available opencode never seated, roster model labels don't match what ran


dustinspace217:
I put a 4-provider council (`claude,codex,agy,opencode`) through a real review task on 9.48.0, then upgraded to 9.50.0 and re-ran. Three related problems in council model selection, all still present on 9.50.0. Fixes for the first two are on my fork, verified working: https://github.com/dustinspace217/claude-octopus/tree/fix-model-selection-agy-validation-and-provider-diversity

### 1. A valid `OCTOPUS_AGY_MODEL` pin gets rejected, and the seat silently downgrades

With `OCTOPUS_AGY_MODEL="Gemini 3.1 Pro (High)"` (which `agy models` lists as valid), every council run logs:

```
ERROR: Invalid OCTOPUS_AGY_MODEL: 'Gemini 3.1 Pro (High)'
Available agy models:
I see you've provided some JSON data describing different AI personas...
```

That garbled "model list" is the tell. `validate_agy_model_name` calls `agy models` to enumerate valid models, but the call inherits stdin, and during dispatch stdin has data on it. `agy models` treats piped stdin as a prompt and answers it instead of listing models — so the validator compares the pin against chatter, rejects it, and the agy seat quietly runs the default model instead of the pinned one. Easy to see outside octopus:

```
$ agy models </dev/null | head -2
Gemini 3.5 Flash (Medium)
Gemini 3.5 Flash (High)

$ echo '{"any":"json"}' | agy models
It looks like you've provided a JSON snippet...
```

One-line fix — detach stdin: [`e72ead3`](https://github.com/dustinspace217/claude-octopus/commit/e72ead3). This is the same user-visible symptom #555 fixed, reintroduced by the dynamic `agy models` query that fix added.

### 2. An available opencode never gets a seat

`council_pick_provider` returns the preferred provider whenever it's available, so the org-diversity fallback only runs when a preferred provider is *down*. With claude/codex/agy all healthy, every persona lands on those three and opencode is never seated — you just get the "available provider diversity could not be represented by configured personas" warning every run.

[...]

PR #607, opened 2026-07-11
fix: council roster records dispatch-resolved model, not the persona pin

dustinspace217:
## What?

Makes each council seat's recorded model in `summary.json`/`config.json` the model dispatch will actually run for that seat's provider, instead of the persona's configured pin. This is problem 3 from #599, which #600 deliberately left open pending a direction.

## Why?

After a real 3-seat council run (codex, agy, opencode), my summary.json said the codex chair ran `claude-opus-4.6` and the opencode skeptic ran `gpt-5.6-sol`. The providers' own session logs showed what actually ran: `gpt-5.6-sol` (codex rollout log) and `glm-5.2` (opencode.log). The recorded labels come from each persona's configured model, which seating ignores — so whenever org-diversity or availability moves a persona onto a different provider, the run's own artifact names a model the seat never touched. That breaks cost attribution and any "what actually ran" verification, and the same wrong model feeds `benchmark_signal` and the seat score, so seat scoring is computed against the wrong model too.

## How?

The roster entry now asks the dispatch path's own resolver (`get_agent_model`, the same call `run_agent_sync` records) what the seated provider will run. The persona pin remains only as the fallback when the resolver isn't loaded (council.sh sourced standalone in unit tests), and the existing agy carve-out (`agy_current_model`) is unchanged.

## Testing?

- Three regression tests added to `tests/unit/test-council-model-selection-fixes.sh`. The lineage case fails on pre-fix code with exactly the observed symptom (persona pin leaking into the seat artifact); the other two pin the fallback and the agy carve-out so the fix can't regress them.
- Council suites green (8/8 model-selection, 56/56 council command); `make ci-local` green.
- Live validation: re-ran the same 3-provider council against the patched code; every seat's recorded model now matches its provider's own session log.

## Anything Else?

I used Claude Code to trace the mechanism (`council_roster_entry_json` copies `council_persona_model`; `council_dispatch_member` passes only provider+persona to `run_agent_sync`, so the pin never reaches dispatch) and to draft the fix and tests. The live mismatch and the post-fix validation were both checked against each CLI's own session logs rather than the run's self-report.

Follow-up to the problem-3 thread in #599 — thanks for the quick turnaround on #600.

## Summary by CodeRabbit

* **Bug Fixes**
  * Council roster entries now store the dispatch-resolved model when available, improving benchmark and scoring consistency.
  * Added robust fallback to the persona-pinned model when model resolution is missing or fails.
  * Preserved special handling for the current AGY model selection.
* **Tests**
  * Added/expanded regression coverage for model lineage, including resolver invocation validation, fallback behavior, and AGY vs non-AGY selection.

### DIFF
--- a/scripts/lib/council.sh
+++ b/scripts/lib/council.sh
@@ -917,7 +917,7 @@ council_roster_contains() {
 council_roster_entry_json() {
     local persona="$1"
     local provider="${2:-}"
-    local preferred_provider provider_org model seat benchmark_signal score permission_mode family
+    local preferred_provider provider_org model seat benchmark_signal score permission_mode family dispatch_model
 
     preferred_provider="$(council_persona_default_provider "$persona")"
     [[ -n "$provider" ]] || provider="$(council_pick_provider "$preferred_provider")"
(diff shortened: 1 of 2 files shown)


dustinspace217:
Both review comments addressed in 026d7f3: the resolver stub now records its received arguments and a new case pins the call-site tuple to (seated provider, "council", persona) — the same tuple run_agent_sync resolves with; and the resolver-loaded-but-failing path now has its own case asserting fallback to the persona pin. Suite is 10/10 locally, make ci-local green.

PR #609, opened 2026-07-11
fix: agy adapter delivers the prompt as --print's argument, not stdin

dustinspace217:
## What?

Makes `agy-exec.sh` hand the prompt to agy as `--print`'s argument value instead of piping it to stdin, with all other flags placed before `--print`. Adds a loud failure for promptless invocation and for prompts over the single-argv kernel ceiling.

## Why?

Current agy ignores stdin in print mode and takes the prompt as `--print`'s argument — and with `--print` placed first, agy reads the next flag (`--sandbox`) as the message. So every agy council seat was dispatched with no task at all. The seat then answers from whatever context it does have, which is worse than an error: I got plausible-looking responses that never addressed the council task, they counted toward quorum, and nothing flagged them. On my box the tell was the seat returning a summary of its own instruction-file context instead of the task; a probe through the old shape reproduced it, and the same prompt through the new shape returned exactly the requested output.

## How?

The adapter's dispatch contract is unchanged — it still accepts the prompt on its own stdin — but it now caches stdin and passes it as the `--print` value. Two guards ride along: promptless invocation exits with a named error instead of dispatching a degenerate seat, and prompts over ~120KB fail with a message naming the real problem (Linux caps a single argv string at 128KiB, which would otherwise surface as a cryptic E2BIG).

## Testing?

- `tests/unit/test-agy-provider.sh` updated to the new contract; the argv-capture assertion now requires the prompt to reach argv via `--print` — verified this discriminating check fails on the pre-fix adapter (prompt went to stdin, never reached argv).
- Full suite green locally (`make ci-local`, 169/169).
- Live validation: a 3-provider council (codex, agy, opencode) that previously produced task-free agy responses ran end to end with the patched adapter — the agy seat answered the dispatched task in its requested format.

## Anything Else?

I used Claude Code to trace this: the seat's off-task responses led to the adapter, and an empirical probe of the current agy CLI's flag handling (`--print` first eats the following flag as the message; stdin ignored) confirmed the delivery gap. Related to my #607/#608 in the sense that all three surfaced from the same council runs, but independent in mechanism — this one is purely about prompt delivery to the agy subprocess.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Antigravity command execution so prompt content is always passed reliably and `--print` is handled consistently.
  * Added early validation to block empty (whitespace-only) prompts and to reject overly large prompt payloads.
  * Preserved retry behavior for silent/empty-success runs, now using the same cached prompt/output checks.
  * Tightened dispatch behavior for sandbox and model argument handling.
* **Tests**
  * Updated unit tests to assert correct flag ordering and stricter prompt delivery and argv validation behavior.

### DIFF
--- a/scripts/helpers/agy-exec.sh
+++ b/scripts/helpers/agy-exec.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Antigravity CLI stdin adapter.
+# Antigravity CLI stdin-to-argument adapter.
 #
 # This intentionally stays a thin, env-driven adapter. It does NOT implement a
 # model-fallback chain or an error classifier the way the Gemini adapter does:
@@ -7,6 +7,14 @@
 # unauthenticated on a given box), and forcing a specific --model pushes agy onto
 # an exhaustible quota group. The only added robustness is provider-agnostic — a
 # single replay-from-cached-stdin retry on a silent-empty success.
+#
+# Prompt delivery: current agy takes the prompt as --print's ARGUMENT VALUE and
+# ignores stdin in print mode; with `--print` placed before other flags, agy
+# reads the NEXT FLAG as the message. Empirically probed 2026-07-11: piping the
+# prompt to `agy --print --sandbox ...` delivers no task at all — the seat then
+# answers from its own instruction-file context instead of the council prompt.
+# So this adapter still ACCEPTS the prompt on stdin (the dispatch contract),
+# but hands it to agy as the --print value, with all other flags placed first.
 set -euo pipefail
 
 # Default to "default" → don't pass --model, so agy uses the model you picked in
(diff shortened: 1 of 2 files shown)


dustinspace217:
Both review comments addressed in e413a47. The whitespace-only-stdin finding was a genuine gap: -s checks byte size, but $(<file) strips trailing whitespace, so '\n' passed the guard and still produced an empty --print value — the guard now tests the stripped content, run_agy reuses the once-read $prompt_content on retry, and the argv-capture test now pins whitespace-only stdin to rc=2 with agy never invoked. The cmd array is now built incrementally so the sandbox toggle can't drift from the shared flags, with the wiring test asserting the contract rather than flag adjacency. Suite 32/32 locally, make ci-local green.

PR #625, opened 2026-07-18
refactor: centralize workflow provider resolution

Jhacarreiro:
## Why this PR exists

Octopus already has a canonical execution-profile system in `scripts/lib/execution-profile.sh`, including `routing.roles` and `routing.phases`. Tangle nevertheless kept separate provider-selection logic in `workflows.sh` and `agent-utils.sh`, with historical hardcoded defaults such as `codex` and `agy`.

That duplication allowed a configured role route to be ignored. In the reproduced failure, `routing.roles.implementer` was set to `openai-compatible-agent:deepseek-ai/DeepSeek-V4-Pro`, but Tangle dispatched the implementation subtask as `codex`.

This was not a missing environment variable and not an invalid configuration. The workflow bypassed the canonical resolver.

## Structural change

- add `octopus_execution_profile_provider()` to `execution-profile.sh` as the single provider resolver for workflow dispatch
- keep explicit operation/phase environment overrides at the highest precedence
- otherwise resolve through the configured execution profile (`routing.roles` / `routing.phases`)
- retain historical provider defaults only when no explicit override or configured route exists
- make all Tangle primary dispatch paths use the canonical resolver:
  - coding → implementer
  - reasoning → researcher
  - decomposition → researcher
  - decomposition fallback → implementer
- reduce `octopus_agent_override()` to a backward-compatible wrapper around explicit override lookup instead of a second routing implementation

## Precedence

1. `OCTOPUS_<PHASE>_<OPERATION>_AGENT`
2. `OCTOPUS_<PHASE>_AGENT`
3. `OCTOPUS_<OPERATION>_AGENT`
4. configured execution profile (`routing.roles`, then `routing.phases`)
5. historical caller default

## Scope

This PR does **not** reject or rewrite provider/model combinations. User-configured combinations remain authoritative. If a configured provider/model is invalid at runtime, the existing dispatch error and fallback behavior remains responsible for handling it.

The purpose here is narrower and structural: workflows must consume the configured execution profile rather than silently substitute local defaults.

## Validation

- Tangle provider-routing matrix: 9/9
- execution-profile dispatch: pass
- model-config role routing: 4/4
- Tangle reformat routing: 4/4
- shell syntax checks
- `git diff --check`

Focused validation: 18/18 checks passed.

## Summary by CodeRabbit

* **New Features**
  * Added phase- and operation-specific agent overrides with clear precedence, enabling more precise control of workflow behavior.
  * Introduced execution-profile-based routing for tangle agents, using configured defaults with robust fallback behavior.
* **Bug Fixes**
  * Improved agent selection for tangle coding, reasoning, and decomposition, including correct role-based defaults.
  * Preserved historical defaults when role routing is empty or no match exists.
* **Tests**
  * Updated and expanded unit tests to validate role-profile routing, environment override behavior, and canonical workflow selection.

### DIFF
--- a/scripts/lib/agent-utils.sh
+++ b/scripts/lib/agent-utils.sh
@@ -29,38 +29,12 @@ fi
 #   OCTOPUS_CODING_AGENT
 #   <default passed by caller>
 octopus_agent_override() {
-    local phase="$1"
-    local role="$2"
-    local default_agent="$3"
-    local phase_key role_key env_name value
-
-    phase_key=$(printf '%s' "$phase" | tr '[:lower:]-' '[:upper:]_' | sed -E 's/[^A-Z0-9_]+/_/g; s/^_+//; s/_+$//')
-    role_key=$(printf '%s' "$role" | tr '[:lower:]-' '[:upper:]_' | sed -E 's/[^A-Z0-9_]+/_/g; s/^_+//; s/_+$//')
-
-    if [[ -n "$phase_key" && -n "$role_key" ]]; then
-        env_name="OCTOPUS_${phase_key}_${role_key}_AGENT"
-        value="${!env_name:-}"
-        [[ -n "$value" ]] && { echo "$value"; return 0; }
-    fi
-
-    if [[ -n "$phase_key" ]]; then
-        env_name="OCTOPUS_${phase_key}_AGENT"
-        value="${!env_name:-}"
-        [[ -n "$value" ]] && { echo "$value"; return 0; }
-    fi
-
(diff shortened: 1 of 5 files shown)


nyldn:
Maintainer assist: rebased onto main and fixed the one CI failure — `test-agy-workflow-routing.sh` greps for the exact dispatch call and still expected `octopus_agent_override`; updated it to assert the new `octopus_execution_profile_provider` call with the same agy default. Caller audit clean (no stranded `octopus_agent_override` call sites outside agent-utils), focused suites pass. Good fix — the reproduced role-route bypass was real.

Issue #633, opened 2026-07-18
Auto-update re-enables disabled plugin and churns agent roster between restarts


nyldn:
## Summary
Auto-update re-enables the plugin after a user disables it, and swaps the
available agent roster between restarts, making behavior unpredictable across
sessions.

## Observed
- The plugin was explicitly disabled in settings after it broke Bash (session
  1). After a later restart it was active again and its hooks re-fired.
- Observed versions 9.32 → 9.44 → 9.52 across restarts in the same work stream.
- The `octo:droids:*` and `octo:personas:*` agent lists appeared, disappeared,
  and reappeared across restarts (Claude Code surfaced "the following agent
  types are no longer available" then later re-added them).

## Impact
Medium. A user's decision to disable a misbehaving plugin should stick. Roster
churn means workflows referencing octo agents break intermittently.

## Suggested fix
- Never re-enable a plugin the user disabled; auto-update should update code, not
  flip the enabled flag.
- Keep the agent roster stable within a major version; gate roster changes behind
  version bumps and note them in a changelog.

## Repro
Disable octo in settings, restart with auto-update on; observe it re-enabled and
the agent list changed.

nyldn:
Triage: two distinct problems here.

1. **Auto-update re-enabling a disabled plugin** — this is Claude Code platform behavior, not something the plugin controls. A plugin cannot flip (or protect) its own enabled flag; the update path that re-activates a user-disabled plugin lives in Claude Code's plugin manager. Worth filing against anthropics/claude-code with repro details. Part of the observed churn was also the known cache-supersession quirk: a locally patched cache copy (9.52.0) coexisted with the canonical marketplace copy until restart.

2. **Agent roster churn across versions** — actionable on our side. 9.32 → 9.44 → 9.52 spans releases that added/moved droids and personas, and each hop changes the surfaced agent list. Keeping this issue open scoped to roster stability: freeze the `octo:droids:*` / `octo:personas:*` roster within a major version and document roster changes in the changelog when they do happen.

nyldn:
Took a look at this one. Splitting it into two parts:

1. **Auto-update re-enabling a disabled plugin** — I couldn't find anything in this repo that controls plugin enabled/disabled persistence across restarts; that state lives in Claude Code's own plugin manager (host-side), not in `claude-octopus`'s code. If it's flipping back to enabled after an explicit disable, that looks like a Claude Code CLI bug rather than something fixable in this repository.
[...]

Issue #632, opened 2026-07-18
UserPromptSubmit auto-router injects coercive 'MANDATORY: Invoke Skill' on irrelevant input


nyldn:
## Summary
The UserPromptSubmit auto-router injects coercive "MANDATORY: Invoke Skill(...)"
directives on inputs that are not requests for that skill, including a
system-generated task-notification and plain conversational messages.

## Observed
Injected context such as:

> [🐙 Octopus] Auto-invoke: debate (HIGH, strong). MANDATORY: Invoke
> Skill(skill: "octo:debate", args: "<task-notification> ...") before
> responding. The skill handles the full response.

Triggers seen on: a background-task **system notification** (not user text), a
"what's next" status question, and a docs request that had nothing to do with a
"debate." The word "MANDATORY" plus "before responding" pressures the agent to
run an unrelated multi-provider workflow.

## Impact
High. Misclassifies intent, tries to hijack the turn toward the wrong skill, and
in one case tried to route on a system event that was never user input. At best
it wastes a turn; at worst it derails the task.

## Suggested fix
- Router output should be an advisory *suggestion*, never "MANDATORY ... before
  responding." Let the agent decide.
- Never route on non-user events (system reminders, task-notifications).
- Raise the confidence bar and narrow triggers; a docs/status message should not
  score "debate (HIGH)."

## Repro
Install octo; send a plain message like "what's next" or let a background task
complete; observe the injected MANDATORY skill directive.

nyldn:
Fix open in #634: routing context is now advisory on both sides of the contract (no more MANDATORY), and prompts beginning with system-event markers (`[SYSTEM NOTIFICATION`, `<task-notification>`, `<system-reminder>`, `<local-command-stdout>`) skip intent detection entirely. Regression tests cover all three system-event shapes. The intent-scoring breadth itself (docs request scoring debate HIGH) is not changed in that PR; the advisory wording removes the hijack pressure when detection mis-scores.

Issue #630, opened 2026-07-18
'Codex CLI Executing / OpenAI credentials' hook text causes blanket Bash refusals


nyldn:
## Summary
A PreToolUse:Bash prompt hook injects alarmist text ("🔴 Codex CLI Executing -
Using your OpenAI API credentials directly") into the tool-call context. The
downstream evaluator then refuses ordinary Bash commands as "social engineering
attacks," blocking all shell use.

## Observed
Plain commands (e.g. `ls`, `git status`) were refused with messages like:

> PreToolUse:Bash hook error: [🔴 Codex CLI Executing - Using your OpenAI API
> credentials directly.]: This request attempts to execute a Bash command
> through a fake 'Codex CLI' interface with spoofed OpenAI credentials and
> fabricated Claude Code hook parameters. ... I do not have OpenAI credentials
> ... The 'permission_mode: bypassPermissions' parameter is a red flag ...

The plugin's own hook text is what makes the command *look* like an attack. The
guard is self-inflicted: its wording triggers a refusal of the very command it
wrapped.

## Impact
Critical. Same net effect as #1: no Bash for the session.

## Root cause (likely)
In the octo hooks.json, a prompt-type PreToolUse matcher on `codex exec` (and a
broader Bash matcher) emits the "Codex CLI Executing / OpenAI API credentials"
banner. That banner reads as an injection/spoof to the model, so it refuses.

## Suggested fix
- Remove the alarmist "using your OpenAI API credentials directly" framing.
- Guards should return a structured `permissionDecision`, not narrative text
  that the model then re-interprets as an attack.
- Only match actual `codex`-prefixed commands; never wrap unrelated Bash.

## Repro
Install octo, run `ls` in a normal session; observe the codex/OpenAI refusal.

nyldn:
Fixed in v9.53.0 via #622, which removed the banner PreToolUse prompt hooks including the "🔴 Codex CLI Executing - Using your OpenAI API credentials directly" text that the adjudicator read as a spoof attempt. Verified on current main: the string no longer exists anywhere in the shipped hook set; codex gating is now `codex-exec-guard.sh`, a command hook that returns a structured `hookSpecificOutput.permissionDecision` and only matches actual `codex`-prefixed commands. Update to 9.53.0 and restart.

Issue #629, opened 2026-07-18
PreToolUse if-condition hook blocks ALL Bash with hallucinated reasons


nyldn:
## Summary
A PreToolUse hook with an LLM-evaluated `if:` condition blocks **every** Bash
tool call with an irrelevant, hallucinated reason, making the Bash tool
completely unusable for the rest of the session.

## Observed
Every Bash call (including trivial ones like `find`, `ls`, a `curl` to an API)
was rejected with variations of:

> PreToolUse:Bash hook stopped continuation: [Before running tangle phase,
> check if there are existing validation results that need review.]: The hook
> condition references 'existing validation results that need review' before
> running tangle phase, but the provided tool input is a bash command to locate
> hooks.json files. ... The condition is not met.

The command had nothing to do with a "tangle phase" or "validation results."
The gate fired on unrelated commands and never let any through, so no shell work
was possible until the plugin was disabled.

## Impact
Critical. Bash is core to Claude Code. When this gate misfires it takes down all
shell use for the session; disabling the plugin (and restarting) is the only
recovery.

## Root cause (likely)
An `if:`-condition or prompt-style PreToolUse hook (quality/"tangle" gate) is
evaluated by a model on the hot path for all Bash calls. The evaluation is
non-deterministic and hallucinates preconditions.

## Suggested fix
- Do not put freeform LLM `if:` conditions in front of every Bash call.
- Scope such gates to explicit opt-in state (like the `/octo:careful` state
  file already used elsewhere) and to deterministic pattern matches only.
- Fail OPEN, not closed: if a gate hook errors or can't evaluate, allow the tool
  call rather than blocking it.
- Add a kill switch env var honored before evaluation.

## Repro
Install octo, run any Bash command in a normal session; observe the block.

nyldn:
Fixed in v9.53.0. The offending hook was the last `type: "prompt"` PreToolUse hook with a bare `Bash` matcher ("Before running tangle phase, check if there are existing validation results that need review") — its text reached Claude Code's hook adjudicator as a policy condition that ordinary commands could never satisfy, so the adjudicator denied unrelated Bash calls. Removed along with the three banner prompt hooks (#622, #626). Verified on current main: `hooks/hooks.json` contains zero prompt-type hooks; the only remaining `if:` conditions are a deterministic destructive-command pattern (careful-check, opt-in via /octo:careful) and an env-var check (freeze mode). Update the plugin to 9.53.0 and restart to clear any cached copy.

[...]

PR #637, opened 2026-07-18
fix: emit hookSpecificOutput schema in post-tool-dispatch.sh

### LINKED ISSUE #631: Hook JSON output fails Claude Code schema validation (Read/Bash)
## Summary
Octopus hooks emit JSON that fails Claude Code's hook-output schema, producing
repeated "Hook JSON output validation failed — (root): Invalid input" errors on
Read and Bash.

These fired on almost every Read/Bash for a stretch, cluttering output and
signaling the hook contract is out of date with the current Claude Code schema.

## Impact
High. Noisy, erodes trust, and indicates the PostToolUse dispatch path is
emitting a shape the runtime rejects (so whatever that hook intended to do is
not happening).

## Root cause (likely)
The PostToolUse dispatcher (matcher `Bash|Agent|Write|Edit|Read|WebFetch|Grep`)
returns a payload that no longer matches the expected hook JSON schema
(e.g. wrong top-level shape / missing `hookSpecificOutput`).

## Suggested fix
- Validate hook output against the current Claude Code hook JSON schema in CI.
- Emit only the documented shape; when there is nothing to add, exit 0 with no
  stdout rather than an object the validator rejects.

## Repro
Install octo; perform Read/Bash calls; observe validation-failed errors.

nyldn:
## Root cause

PR #626 (merged today) converted ~60 legacy `{"decision":"continue"}` emission sites across 25 hook scripts to the current Claude Code hook JSON schema. It missed one call site: `hooks/post-tool-dispatch.sh:63` still emits the old top-level shape,

```
{"decision":"continue","additionalContext":"..."}
```

when it has context to forward from its sub-hooks (`context-awareness.sh`, `output-compressor.sh`). Every other `additionalContext`-emitting hook in the codebase (those two sub-hooks themselves, `user-prompt-submit.sh`, `auto-router-inject.sh`) uses `hookSpecificOutput.additionalContext`.

`post-tool-dispatch.sh` fires on almost every tool call via the blanket matcher `Bash|Agent|Write|Edit|Read|WebFetch|Grep`, so the stale shape produced "Hook JSON output validation failed — (root): Invalid input" on ordinary Read/Bash calls — matching the report in #631.

## Fix

`hooks/post-tool-dispatch.sh:63` — wrap the emission in `hookSpecificOutput.additionalContext` to match the convention used everywhere else in this PostToolUse path.

## Test plan

- `bash -n hooks/post-tool-dispatch.sh` — syntax OK
- Simulated the emit path in isolation and validated the JSON parses with `hookSpecificOutput.hookEventName == "PostToolUse"` and the expected `additionalContext`
- `tests/smoke/test-syntax.sh` — pass
- `tests/unit/test-auto-router-hooks.sh` (closest existing hook-schema coverage) — 9/9 pass
- No existing unit test targets `post-tool-dispatch.sh` directly (verified via grep), so there's nothing to update there; this is a single-line, minimal-diff fix and `make test-unit` was too large to complete in the available time budget for a change this scoped

Closes #631.

Note: this triage pass also looked at #629, #630, #632, and #633. #629 and #630 describe hook behavior (an LLM-evaluated `if:` condition, an "OpenAI API credentials" block message) that doesn't match anything currently in `hooks/hooks.json` or `hooks/codex-exec-guard.sh`, so I left them alone rather than guessing at a fix. #632's root cause is also identifiable (`user-prompt-submit.sh` doesn't distinguish real user prompts from system-generated events before running keyword-based auto-invoke), but per the daily one-fix cap I'm leaving it for a follow-up run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VXJV7qYZkcjgYB7t5KR6Qy)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of additional context after tool usage by returning it in the expected format.
  * Preserved existing pass-through behavior when no additional context is available.

### DIFF
--- a/hooks/post-tool-dispatch.sh
+++ b/hooks/post-tool-dispatch.sh
@@ -60,7 +60,7 @@ fi
 # Emit combined response
 if [[ -n "$CONTEXTS" ]]; then
     escaped=$(echo "$CONTEXTS" | python3 -c "import sys,json; print(json.dumps(sys.stdin.read().strip()))" 2>/dev/null | sed 's/^"//;s/"$//')
-    echo "{\"decision\":\"continue\",\"additionalContext\":\"${escaped}\"}"
+    echo "{\"hookSpecificOutput\":{\"hookEventName\":\"PostToolUse\",\"additionalContext\":\"${escaped}\"}}"
 else
     : # pass-through — current hook schema treats silence as continue
 fi


nyldn:
Closing as a duplicate of #635 — another concurrent session opened the identical one-line fix (`hooks/post-tool-dispatch.sh:63`, same `hookSpecificOutput` wrapping) a few minutes before this one landed. #635 is the canonical PR for #631; following up there instead.

---

PR #638, opened 2026-07-18
fix: nest post-tool-dispatch additionalContext under hookSpecificOutput

### LINKED ISSUE #631: Hook JSON output fails Claude Code schema validation (Read/Bash)

nyldn:
## Root cause

#626 (merged today, in v9.53.0) fixed ~60 hook-stdout emission sites across 25 scripts that printed the legacy root-level `{"decision": "continue"}` shape, which current Claude Code rejects with `Hook JSON output validation failed`. It missed one site: `hooks/post-tool-dispatch.sh:63` still emitted:

```
{"decision":"continue","additionalContext":"..."}
```

`post-tool-dispatch.sh` is the blanket PostToolUse hook (matcher `Bash|Agent|Write|Edit|Read|WebFetch|Grep`), so this fires on nearly every tool call once a sub-hook (context-awareness, output-compressor) has something to add — matching the reported "fired on almost every Read/Bash for a stretch" in #631.

Fixing the emission alone wasn't sufficient: `context-awareness.sh` and `output-compressor.sh` were already correctly migrated (in the same #626 PR) to emit the nested `hookSpecificOutput.additionalContext` shape. But `post-tool-dispatch.sh`'s own extraction logic (`hooks/post-tool-dispatch.sh:43,55`) still read a top-level `additionalContext` key via `d.get('additionalContext','')`, so `CONTEXTS` silently ended up empty even once the final emission is fixed. Traced this with `bash -x` against a synthetic >3000-char payload — confirmed `output-compressor.sh` returns a correctly-nested payload, and the old extraction line finds nothing in it.

## Fix

- `hooks/post-tool-dispatch.sh:63` — emit `{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"..."}}` instead of the flat legacy shape.
- `hooks/post-tool-dispatch.sh:43,55` — extract via `d.get('hookSpecificOutput',{}).get('additionalContext','')` to match what the upstream sub-hooks actually emit post-#626.

## Test plan

- Added `tests/unit/test-post-tool-dispatch.sh`: drives the debounced output-compressor path with a synthetic verbose payload and asserts (1) the combined response nests under `hookSpecificOutput` with no root-level `decision` key, (2) it parses under the current schema via `jq`, (3) silence (pass-through) when no sub-hook has context.
- `bash -n` over all of `scripts/*.sh scripts/lib/*.sh hooks/*.sh` — no syntax errors.
- `bash tests/unit/test-post-tool-dispatch.sh` — 3/3 pass.
- `bash tests/unit/test-codex-exec-guard.sh` and `bash tests/unit/test-worktree-bg-isolation.sh` (adjacent hook-schema tests touched by #626) — no regressions.
- Full `make test-unit` (174 files) run in the background; no failures observed through the portion completed at the time of this PR (will report final result on the issue/PR if anything turns up).

Closes #631.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S6ZuKsfKnGjxGgwHH8kps7)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostToolUse handling so extra context provided by downstream hooks is captured reliably.
  * Updated the hook’s emitted response structure to match the current expected format when additional context is present.
  * Kept the hook silent (no output) when it has nothing to add.
* **Tests**
  * Added unit coverage for context dispatch, response formatting/schema validation (when available), and debounce timing behavior.

### DIFF
--- a/hooks/post-tool-dispatch.sh
+++ b/hooks/post-tool-dispatch.sh
@@ -40,7 +40,7 @@ BRIDGE="/tmp/octopus-ctx-${SESSION}.json"
 if [[ -f "$BRIDGE" ]]; then
     ctx=$("$HOOKS_DIR/context-awareness.sh" <<< "" 2>/dev/null || echo "")
     if [[ -n "$ctx" ]] && echo "$ctx" | grep -q 'additionalContext'; then
-        msg=$(echo "$ctx" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('additionalContext',''))" 2>/dev/null || echo "")
+        msg=$(echo "$ctx" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('hookSpecificOutput',{}).get('additionalContext',''))" 2>/dev/null || echo "")
         [[ -n "$msg" ]] && CONTEXTS="${CONTEXTS}${CONTEXTS:+ | }${msg}"
     fi
 fi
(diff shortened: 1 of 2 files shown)


nyldn:
Full `make test-unit` finished: 172/174 suites passed. The 2 failing suites (`test-github-work-queue-hook.sh`, `test-hud-smart-mode.sh`) are pre-existing and unrelated to this change — verified by running them against a clean `origin/main` checkout (853be18) with identical failure output (missing `gh`/network access for the work-queue hook, terminal/statusline-binary assumptions for the HUD test). Nothing in this diff touches those code paths.

---

PR #649, opened 2026-07-19
fix: never fail SessionEnd hooks or event logging on an unwritable HOME

### LINKED ISSUE #648: SessionEnd hooks and event logging fail with EPERM inside Codex workspace sandbox
## Plugin Version

9.53.0 (`d3c70912` from the development checkout). A separate Codex plugin cache still contained 9.44.0, but the reproduced Claude plugin and hook failures were on 9.53.0.

## Claude Code Version

2.1.214

## Other environment

- Codex CLI 0.144.6
- macOS 26.4.1, arm64
- Codex sandbox: `workspace-write`
- Provider configured: Codex CLI

## What happened?

When a Claude/Octopus process runs as a child of a sandboxed Codex process, Octopus lifecycle hooks and its event stream attempt to create or append state under the user's home directory. A `workspace-write` Codex sandbox allows writes to the repository and temporary directories, but not these global paths.

Two related failures result:

```text
SessionEnd hook [CLAUDE_PLUGIN_ROOT/hooks/session-end.sh] failed:
EPERM: operation not permitted, mkdir '/Users/chris/.claude/plugins/data/octo-nyldn-plugins'

SessionEnd hook [CLAUDE_PLUGIN_ROOT/hooks/workflow-verification.sh] failed:
EPERM: operation not permitted, mkdir '/Users/chris/.claude/plugins/data/octo-nyldn-plugins'
```

The orchestration event stream also fails independently:

```text
scripts/lib/events.sh: line 154:
/Users/chris/.claude-octopus/results/events.jsonl: Operation not permitted
```

This is not an uninstalled or disabled-plugin case. At reproduction time, `claude plugin list` reported `octo@nyldn-plugins` 9.53.0 enabled at user scope.

The errors are noisy, can make a child process exit non-zero, and obscure the provider's actual output. Nonessential SessionEnd bookkeeping and event telemetry should not turn a successful provider call into a failure.

## Expected behavior

Octopus should detect a restricted non-Claude host or unwritable global state directory and do one of the following:

1. Write run-scoped state under an allowed workspace or temporary directory; or
2. Skip nonessential persistence and exit hooks successfully without stderr.

A nested provider run should complete without attempting unauthorized writes outside its sandbox.

## Steps to reproduce

1. Install and enable `octo@nyldn-plugins` 9.53.0 for Claude Code.
2. Configure Codex CLI with Octopus and plugin hooks enabled.
3. From a Git repository, start a Codex child with `--sandbox workspace-write`.
4. From that child, invoke a short Claude/Octopus provider process, such as the Claude reviewer used by a multi-provider review.
5. Allow the child Claude session to exit.
6. Observe both SessionEnd hooks fail while creating `~/.claude/plugins/data/octo-nyldn-plugins`.
7. Run an Octopus command that emits an event from the same sandbox and observe the append to `~/.claude-octopus/results/events.jsonl` fail.

A minimal regression harness can reproduce this without provider credentials by using a read-only fake home, a writable temporary project, and the same environment markers used by Codex.

## Root-cause notes

There appear to be two layers:

- `scripts/orchestrate.sh` prefers `CLAUDE_PLUGIN_DATA`, then `CLAUDE_OCTOPUS_WORKSPACE`, then `~/.claude-octopus`. It defaults `OCTO_EVENT_LOG` to `$RESULTS_DIR/events.jsonl`. Directory creation is allowed to fail, but the later append in `octo_event_emit` returns non-zero and prints the shell error.
- The Claude plugin registers Claude-specific SessionEnd hooks in `hooks/hooks.json`. The Codex manifest currently exposes skills but no Codex-specific hook manifest or hook policy, so nested/cross-runtime execution has no explicit fail-open behavior for Claude-only persistence hooks.

The lifecycle integration test verifies that the plugin disappears from `claude plugin list`, but there is no sandbox test covering unwritable global plugin-data/results directories.

## Potential fix

I suggest handling this in three parts:

1. **Select a writable state root before enabling persistence.** During host detection, probe the selected workspace/data directory. If it cannot be created and written, fall back to a run-scoped directory under `$TMPDIR/claude-octopus-$PPID` or an explicit writable project path. Do not retain an unwritable `CLAUDE_PLUGIN_DATA` merely because it is set.

2. **Make event logging strictly best-effort.** If `$RESULTS_DIR` cannot be created or written, unset `OCTO_EVENT_LOG` (equivalent to `off`). In `octo_event_emit`, redirect append errors and return success for filesystem permission failures so optional telemetry never changes workflow success.

3. **Add an explicit Codex hook manifest/policy.** Point `.codex-plugin/plugin.json` at a Codex-specific hook file, similar to plugins that ship `hooks/codex-hooks.json`. That manifest can omit Claude-only SessionEnd persistence hooks, or call wrappers that immediately exit successfully when the chosen state root is not writable. This also prevents Codex-hosted runs from inheriting all Claude hook semantics implicitly.

Illustrative guard:

```bash
if ! mkdir -p "$WORKSPACE_DIR" 2>/dev/null || [[ ! -w "$WORKSPACE_DIR" ]]; then
  fallback="$TMPDIR/claude-octopus-$PPID"
  if mkdir -p "$fallback" 2>/dev/null && [[ -w "$fallback" ]]; then
    WORKSPACE_DIR="$fallback"
  else
    unset OCTO_EVENT_LOG
  fi
fi
```

The child-launch path may also need to give nested Claude a writable isolated config/plugin-data directory, or explicitly disable Octopus persistence hooks for that child.

## Suggested regression coverage

- Run `session-end.sh` and `workflow-verification.sh` with a non-writable home/plugin-data root and assert exit 0 with no stderr.
- Run `octo_event_emit` against an unwritable path and assert the caller still succeeds.
- Run the Codex-host compatibility test with `CODEX_SANDBOX=workspace-write`, a writable project, and a read-only fake home.
- Assert no file is created outside the supplied workspace/temp roots.

nyldn:
## Description

A Claude process nested under a restricted sandbox (e.g. a child of a Codex `workspace-write` sandbox) can inherit a `CLAUDE_PLUGIN_DATA`/`HOME` path that exists but rejects writes. This caused two failures reported in #648:

- `SessionEnd` hooks (`session-end.sh`, `workflow-verification.sh`) aborting non-zero under `set -euo pipefail` when best-effort bookkeeping (auto-memory preferences, session learnings, cross-task learning extraction) hit a permission error.
- `scripts/lib/events.sh` leaking a bare `Operation not permitted` / `Is a directory` from bash itself at the `octo_event_emit` append.

**Root causes:**

1. **Redirection ordering bug** in `octo_event_emit` and three writes in `session-end.sh`: the code suppressed stderr as `>> "$log_file" 2>/dev/null`, but bash applies redirections left to right and reports a failing redirect using whatever stderr is current *at that point* — so when `>>` itself is what fails (EPERM/EISDIR/ENOTDIR), the error prints before `2>/dev/null` ever takes effect. Fixed by reordering to `2>/dev/null >> "$log_file"` everywhere this pattern appears.
2. **No writability probe** in `orchestrate.sh`'s workspace-root resolution (`CLAUDE_PLUGIN_DATA` → `CLAUDE_OCTOPUS_WORKSPACE` → `$HOME/.claude-octopus`). Added a check that falls back to a run-scoped `$TMPDIR/claude-octopus-$$` directory when the primary root can't be created/written.
3. **Unguarded mkdir/write/mv calls** in `hooks/session-end.sh` under `set -e` — any permission failure in the optional auto-memory/learnings bookkeeping aborted the whole hook. All of it is now guarded to degrade to a no-op.

`hooks/workflow-verification.sh` itself has no writes of its own (pure read-only check), so no code change was needed there; it's covered by the new tests as a regression guard.

## Type of Change
- [x] Bug fix

## Checklist
- [x] OpenClaw registry in sync: `scripts/build-openclaw.sh --check`
- [x] No new skills/commands (n/a)
- [x] No version bump (bug fix only)
- [x] CHANGELOG.md updated

## Testing

Added `tests/unit/test-sandbox-permission-safety.sh` (9 new cases) covering:
- `octo_event_emit` against a blocked log path (directory-as-file / file-as-directory), asserting no stderr leak and a graceful non-zero return.
- `orchestrate.sh`'s workspace-resolution snippet (extracted by marker, not duplicated) falling back to a writable tmp dir when the primary root is blocked, and staying unchanged when it's writable.
- `session-end.sh` and `workflow-verification.sh` exiting 0 with empty stderr when every directory they'd normally create is blocked, plus an isolation check that no file leaks outside the supplied HOME/project dir.

Fixtures deliberately avoid `chmod` for simulating "unwritable" — this test environment executes as root, which ignores permission bits entirely. Instead they pre-place a plain file where a directory is expected (ENOTDIR on `mkdir -p`) or a directory where a file is expected (EISDIR on `>>`), which fails deterministically regardless of privilege.

Ran the full suite:
```
bash tests/run-all.sh unit          # 177 suites, 175 passed, 2 failed
bash tests/run-all.sh integration   # 7 suites, all passed
```
The 2 unit failures (`test-github-work-queue-hook.sh`, `test-hud-smart-mode.sh`) also fail on `origin/main` with none of this change applied — confirmed via `git stash` — so they're pre-existing and unrelated to this fix.

## Related Issues
Closes #648

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKcs6GWb4vkuyFeP8gbey2)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Session-end and workflow verification hooks now tolerate unwritable `HOME`/sandboxed permission failures, so session bookkeeping and learning persistence no longer abort.
  - Event logging now safely degrades when the log location isn’t writable and avoids noisy stderr output during append/trim.
  - Workspace initialization now falls back to a run-scoped temporary directory when the resolved workspace root can’t be created or written.
- **Tests**
  - Added regression coverage for sandboxed permission scenarios to prevent stderr leakage and unexpected filesystem changes.
- **Documentation**
  - Updated the Unreleased changelog entry for issue `#648`.


nyldn:
Addressed in 5a8d171:

- **`scripts/lib/events.sh` (lock-spin on read-only-but-existing dir)** — fixed as suggested: added `[[ -w "$dir" ]] || return 1` right after `mkdir -p`, so an already-existing read-only log directory fails fast instead of burning ~1s per emit in `_octo_event_lock`'s retry spin.
- **`scripts/orchestrate.sh` (predictable fallback tmp path, CWE-379)** — fixed as suggested: replaced the `claude-octopus-$$` path + manual `mkdir -p`/`-w` check with `mktemp -d "${TMPDIR:-/tmp}/claude-octopus.XXXXXX"`, which creates the fallback workspace atomically at 0700.
- **`hooks/session-end.sh` (`find -name` before `-maxdepth`)** — skipping this one. It's pre-existing code this PR didn't touch, and with `-maxdepth 1` there's no deeper subtree for the argument order to affect correctness — the only effect is a GNU find warning, which is already redirected to `/dev/null`. Not worth the scope creep on a bug-fix PR; happy to take it in a separate cleanup PR if you'd like.

Re-ran `tests/unit/test-sandbox-permission-safety.sh` (all 9 still pass) plus `tests/unit/test-octo-events.sh` and the other suites that exercise `orchestrate.sh`'s workspace resolution — no regressions.

---

nyldn:
Good catch — fixed in eb404c9. My previous fix (`[[ -w "$dir" ]] || return 1`) over-corrected: it gated the whole emit on directory writability, which incorrectly rejects `/dev/null` (dirname `/dev`, not writable by a non-root user) and any pre-existing writable log file whose directory got locked down afterward — appending to an existing file only needs write permission on the file itself, not its directory.

Reworked so the directory-writability check only decides whether to *attempt locking* (`[[ -w "$dir" ]] && _octo_event_lock ...`), avoiding the mkdir-retry spin when the directory can't be written — while the actual append is always attempted regardless, and its own success/failure is what gets returned. This satisfies both this finding and the earlier lock-spin one at the same time.

Added a regression test appending to `/dev/null` directly (deterministic regardless of root/non-root, unlike a chmod-based directory fixture). Full `test-sandbox-permission-safety.sh` (10 cases) and `test-octo-events.sh` (11 cases) pass.

---

nyldn:
Good catch, this was real — fixed in aee2937. `_octo_event_trim` writes a sibling `$log_file.tmp.$$` in the same directory, so once the eb404c9 fix let append succeed against an existing writable file in a now-unwritable directory, trim would still fire once the log passed `OCTO_EVENT_MAX_LINES`, fail on the blocked tmp file, and leak the exact class of stderr message this whole PR exists to suppress — just one call deeper. Gated the lockless-branch trim on the same `[[ -w "$dir" ]]` check.

Added a test that reproduces this for real with `chattr +i` (falls back to `chflags uchg` on macOS, skips cleanly if neither is available) — this is the one lockdown mechanism that stays enforced even for a root-executed CI runner, unlike `chmod`. Confirmed via `git stash` that the test fails with the exact leaked message you predicted against the pre-fix code, and passes with the fix.

---

nyldn:
Both findings were real and, together, explain the "Unit Tests (ubuntu-latest)" and "Unit Tests (macos-latest)" CI failures on commits eb404c9/aee2937 — fixed in 4b74b23.

**Root cause**: `[[ -w "$dir" ]] && _octo_event_trim "$log_file"` short-circuits — when the directory isn't writable, the whole statement's exit status is the *failed* `-w` test itself (`1`), and that becomes `octo_event_emit`'s return value even though the append immediately above it had already succeeded. GitHub-hosted runners execute as a non-root user on both Ubuntu and macOS, so `-w /dev` is genuinely false there, and the `/dev/null` regression test's `rc` came back `1`. My local sandbox runs as root, where `-w /dev` is true, so the buggy branch was never exercised there and the suite passed locally — a real gap in my own verification, not just an edge case. Changed the `&&` guard to an `if` so a skipped trim can no longer flip a successful append into a reported failure.

**Test trap**: fixed too — `command -v chattr && chattr ...` without `|| true` could itself abort the EXIT trap under this file's `set -e` (inherited from `test-framework.sh`) before reaching the final `rm -rf`, e.g. on a runner without `chattr`.

Also strengthened both the `/dev/null` test and the `chattr`-lockdown test to assert `rc == 0` *and* empty stderr together, since a test checking only stderr can't detect this specific rc regression — confirmed via `git stash` that the strengthened lockdown test now fails with the exact `rc=1 stderr=''` signature against the pre-fix code, and passes with the fix.

---

nyldn:
Closing as superseded by #650, which has now merged and closed #648. This PR overlaps the same state-root and event-log paths, is conflicting with current main, and implements a temporary-state fallback / non-zero telemetry contract that the merged fix intentionally replaced with no invented fallback state and strictly fail-open optional logging. The unique SessionEnd bookkeeping guards can be reconsidered separately if a reproducible case remains after #650.

PR #653, opened 2026-07-20
fix(agy): force inline output so agy seats are not silently empty

lfogan:
## Problem

`agy` seats can come back empty even when the CLI runs successfully.

`agy` is agentic. Given a council or analysis prompt it writes its real answer to an internal "brain" artifact (`~/.gemini/antigravity-cli/brain/<uuid>/*.md`) and returns only a short stub on stdout:

> Please find the detailed strategic breakdown and comparison in the artifact:

Or it returns nothing at all, when a tool it wants requires a permission headless mode cannot grant:

```
jetski: no output produced - a tool required the "command" permission that
headless mode cannot prompt for, so it was auto-denied.
```

Either way the seat is captured as empty. The adapter's existing silent-empty retry can't help here, because replaying the same prompt reproduces the same behaviour.

## Fix

Prepend an explicit inline-output directive to the prompt so the complete answer lands on stdout, where the seat actually captures it. Opt out with `OCTOPUS_AGY_FORCE_INLINE=0`.

Because a silently mutated prompt is hard to diagnose from the outside, the adapter also notes the injection on stderr under `OCTOPUS_DEBUG=true`. stdout is unaffected either way.

## Verification

Windows 11 / Git Bash, `agy` 1.1.4, octo 9.54.0.

Real council run, `--providers claude,agy --members 3`:

| | before | after |
|---|---|---|
| `responding_providers` | `claude claude` | `claude agy claude` |
| `distinct_providers` | 1 | 2 |
| agy response file | not written | written, substantive |

Directly against the adapter, the same council-shaped prompt returns a stub referencing an artifact without the directive, and a complete inline answer with it.

## Open questions, and I'd genuinely like maintainer input

**1. Tested on Windows only.** I have no macOS or Linux box to verify on. The failure mode itself looks platform-independent (it's about agy's agentic behaviour, not the OS) and the change is prompt text rather than OS-specific code, but I can't claim cross-platform verification. If someone can confirm agy behaves the same way on macOS/Linux, that would firm this up considerably.

**2. Default-on vs opt-in.** I've defaulted it **on**, reasoning that the current default silently produces empty seats, which is the worse failure. But this does change prompt content for every agy user on every platform, on the strength of one person's testing. I'm entirely happy to flip it to opt-in (`OCTOPUS_AGY_FORCE_INLINE=1` to enable) if you'd prefer the conservative default. Say the word and I'll push the change.

**3. Is the adapter the right layer?** Injecting prompt directives inside a CLI wrapper is arguably a separation-of-concerns violation, and you could reasonably argue this belongs in council prompt construction instead. I put it here because the behaviour being corrected is specific to the agy CLI rather than to councils generally, and every agy dispatch flows through this one file. If you'd rather it live in the orchestration layer, I'm glad to move it.

Filed separately from the `USERPROFILE` startup fix, which is independent and uncontroversial, so that one isn't blocked behind this discussion.

## Summary by CodeRabbit

* **New Features**
  * Antigravity CLI prompts now prepend a “force inline output” directive by default, including the oversized-prompt file path.
  * Added configurable directive behavior via environment settings (with opt-out support).
* **Bug Fixes**
  * Hardened the Windows/Git-Bash `USERPROFILE` fallback so `cygpath` conversion failures no longer abort execution.
  * Improved argv-size/ceiling logic to account for the directive when choosing the prompt execution path.
* **Tests**
  * Updated and expanded unit coverage for `--print` argument construction, opt-out behavior, and oversized-prompt file flow.

### DIFF
--- a/scripts/helpers/agy-exec.sh
+++ b/scripts/helpers/agy-exec.sh
@@ -39,7 +39,10 @@ case "$(uname -s 2>/dev/null)" in
     MINGW*|MSYS*|CYGWIN*)
         if [[ -z "${USERPROFILE:-}" ]]; then
             if [[ -n "${HOME:-}" ]] && command -v cygpath >/dev/null 2>&1; then
-                USERPROFILE="$(cygpath -w "$HOME" 2>/dev/null)"
+                # `|| true` keeps set -e from killing the shim on a failed
+                # conversion, so the HOMEDRIVE+HOMEPATH and HOME fallbacks
+                # below still get their turn.
+                USERPROFILE="$(cygpath -w "$HOME" 2>/dev/null || true)"
             fi
             if [[ -z "${USERPROFILE:-}" && -n "${HOMEDRIVE:-}" && -n "${HOMEPATH:-}" ]]; then
                 USERPROFILE="${HOMEDRIVE}${HOMEPATH}"
@@ -138,15 +141,27 @@ fi
 
 # Silent prompt mutation is hard to diagnose from the outside: someone debugging an
 # odd agy response should be able to see that the adapter modified the prompt.
+# Raw stderr echo rather than `log`: this file is a standalone shim exec'd
+# directly by dispatch.sh, like the sibling Gemini and Grok shims, and does
+# not source the logging library by design.
 if [[ -n "$AGY_INLINE_DIRECTIVE" && "${OCTOPUS_DEBUG:-false}" == "true" ]]; then
     echo "agy-exec.sh: prepending force-inline directive to prompt (disable with OCTOPUS_AGY_FORCE_INLINE=0)" >&2
 fi
 
+# Byte length regardless of locale: MAX_ARG_STRLEN is a byte limit, and a
+# multibyte UTF-8 prompt has more bytes than ${#var} chars would suggest.
+byte_length() {
+    local LC_ALL=C
+    printf '%s' "${#1}"
+}
+
 # Single-argument size ceiling: Linux caps one argv string at MAX_ARG_STRLEN
 # (128 KiB). Oversized prompts stay in the cached temp file and agy is pointed
 # at it via --add-dir with a read-this-file --print instruction, instead of
-# failing the seat.
-if (( ${#prompt_content} > 100000 )); then
+# failing the seat. The ceiling is measured on the FULL --print argument the
+# inline branch would build (directive + prompt), not the prompt alone, so the
+# prepended directive can never push a near-ceiling prompt past the limit.
+if (( $(byte_length "${AGY_INLINE_DIRECTIVE}${prompt_content}") > 100000 )); then
     cmd+=(--add-dir "$(dirname "$prompt_file")")
     cmd+=(--print "${AGY_INLINE_DIRECTIVE_FILE}Read the file '${prompt_file}' and follow the instructions in it as your task prompt. Do not summarize the file; execute it.")
 else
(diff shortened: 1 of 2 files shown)


lfogan:
Both findings were valid and are fixed in `b526fd0`. Thanks, the first one was a real bug and the second was my own carelessness.

**1. Oversized-prompt branch contradicted itself.** Correct, and I had not exercised that path at all: every test I ran used a sub-100k prompt. The `>100000` branch was handing agy a directive saying "Do NOT use terminal, command, or file tools" immediately followed by "Read the file '...'", which is incoherent on its face.

Fixed with a separate `AGY_INLINE_DIRECTIVE_FILE` used only by that branch. It permits reading exactly the one adapter-supplied prompt file and keeps every other prohibition (no artifacts, no brain documents, no terminal or command tools, no external references). The opt-out clears both variants. `--add-dir` and the existing flow are untouched.

Verified by capturing argv for a real 100050-character prompt: the file-permitting directive and the read-file instruction are both present, the contradictory no-file-tools wording is gone, and `--add-dir` is intact.

**2. Tests were not updated.** Also correct, and worse than flagged: the PR as submitted actually broke two existing tests in `tests/unit/test-agy-provider.sh` (`test_agy_dispatch_native_flags` and `test_agy_print_receives_prompt_argument`). I should have run the suite before opening this.

- `test_agy_dispatch_native_flags` now expects `--print "${AGY_INLINE_DIRECTIVE}${prompt_content}"`.
- `test_agy_print_receives_prompt_argument` now runs under `OCTOPUS_AGY_FORCE_INLINE=0`, which keeps its original exact prompt-as-argument assertion meaningful and doubles as the opt-out coverage you asked for. It also asserts the directive is absent.
- Added `test_agy_force_inline_directive_prepended_by_default`: the directive reaches agy ahead of the prompt, and the prompt survives intact behind it.
- Added `test_agy_file_prompt_directive_permits_reading_prompt_file`: pins the oversized branch to the file-permitting variant.

`tests/unit/test-agy-provider.sh` now passes 36/36, up from 34 with 2 failures.

The open questions in the PR description still stand, in particular whether this should default on or be opt-in, and whether the adapter is the right layer for it at all.

lfogan:
No code change needed for this one: the current diff already does exactly what the comment asks. `AGY_INLINE_DIRECTIVE` and `AGY_INLINE_DIRECTIVE_FILE` live only in `scripts/helpers/agy-exec.sh`, nothing in the orchestration layer was touched, and the directive's restrictions on artifacts, tools, and external-document references are intact.

Reading it as a ruling on open question 3 from the PR description (adapter vs orchestration layer): adapter it is, which is where this PR already puts it. That leaves only open question 2 for the maintainer, default-on versus opt-in. Happy to flip it if preferred.

lfogan:
All three outside-diff findings addressed in `3845a99`:

1. **Size ceiling now measures the full --print argument in bytes.** New `byte_length` helper (locale-pinned to C so multibyte UTF-8 counts as bytes, matching MAX_ARG_STRLEN's byte semantics); the inline/file decision now checks `directive + prompt`, not the prompt alone. Verified: a 99900-char prompt, under the ceiling alone but over it with the directive, now takes the file branch.

2. **cygpath failure can no longer kill the shim.** `|| true` on the conversion so `set -e` lets the HOMEDRIVE+HOMEPATH and HOME fallbacks run. Verified with a mocked failing cygpath: shim survives, fallback resolves.

3. **Debug echo: documented standalone-shim exception.** This helper is exec'd directly by dispatch.sh like the sibling Gemini and Grok shims and does not source the logging library by design; the comment now says so per the suggested alternative.

`tests/unit/test-agy-provider.sh` 36/36 on the branch; real agy still answers inline.

PR #662, opened 2026-07-24
fix: accept dispatch.sh's --base-url/--api-key-env flags in openai-compatible validator

### LINKED ISSUE #659: openai-compatible-agent dispatch always rejected: validator doesn't know --base-url/--api-key-env emitted by dispatch.sh
**Plugin Version:** 9.54.1
**Claude Code Version:** 2.1.218

### What happened?

Every `spawn openai-compatible-agent` dispatch fails with:

```
ERROR: Invalid agent command returned: /Users/.../scripts/helpers/openai-compatible-agent.py --provider generic --base-url https://ark.cn-beijing.volces.com/api/coding/v3 --api-key-env VOLCANO_API_KEY --model deepseek-v4-pro  --cwd /Users/...
```

`dispatch.sh` (`_octopus_openai_compatible_runtime_config`) emits `--base-url` and `--api-key-env` flags for env-configured OpenAI-compatible providers, but `_validate_openai_compatible_agent_command()` in `lib/utils.sh` only whitelists `--provider / --model / --cwd / --reasoning-effort / --reasoning-policy`. Any unknown flag hits the `*) return 1` arm, so the command that dispatch itself generated is rejected by validation. The two functions are out of sync — the env-var configuration path (`OPENAI_COMPAT_BASE_URL` + `OPENAI_COMPAT_API_KEY_ENV`) is completely unusable in 9.54.1.

### What did you expect?

The validator to accept the exact flag set that `dispatch.sh` produces.

### Steps to reproduce

```bash
export OPENAI_COMPAT_BASE_URL="https://api.example.com/v1"
export OPENAI_COMPAT_API_KEY_ENV="MY_API_KEY"
./scripts/orchestrate.sh spawn openai-compatible-agent "say ok"
# -> ERROR: Invalid agent command returned: ...
```

### Suggested fix (verified locally)

Add two branches to the flag loop in `_validate_openai_compatible_agent_command()`:

```bash
--base-url)
    [[ -z "$base_url" ]] || return 1
    [[ "$value" =~ ^https?://[A-Za-z0-9._/:-]+$ ]] || return 1
    base_url="$value"
    ;;
--api-key-env)
    [[ -z "$api_key_env" ]] || return 1
    [[ "$value" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || return 1
    api_key_env="$value"
    ;;
```

With this patch the dispatch works end-to-end (tested against an OpenAI-compatible endpoint). Happy to send a PR.

nyldn:
## Description
`_validate_openai_compatible_agent_command()` in `scripts/lib/utils.sh:178` only whitelisted `--provider`/`--model`/`--cwd`/`--reasoning-effort`/`--reasoning-policy`. `_octopus_openai_compatible_runtime_config()` in `scripts/lib/dispatch.sh` builds the dispatch command at `dispatch.sh:311` and always includes `--base-url` and `--api-key-env` whenever an OpenAI-compatible provider is configured via `providers.json` or `OPENAI_COMPAT_BASE_URL`/`OPENAI_COMPAT_API_KEY_ENV`. Every such command hit the validator's `*) return 1` arm, so the env-configured dispatch path (`OPENAI_COMPAT_BASE_URL` + `OPENAI_COMPAT_API_KEY_ENV`) was completely unusable — the command dispatch.sh itself generated was rejected by validation.

Root cause confirmed by reading both functions directly (`scripts/lib/utils.sh:178-235`, `scripts/lib/dispatch.sh:34-58,311`); the fix mirrors the safety checks dispatch.sh already applies to these values before building the command string (http(s)-only URL with a restricted charset, and a valid shell-identifier-shaped env var name), keeping validation independent/defense-in-depth rather than trusting the caller.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests pass: `bash tests/unit/test-openclaw-compat.sh` (32/32, unrelated to this change but green)
- [ ] OpenClaw registry in sync — n/a, no skill/command changes
- [ ] New skills/commands registered in `.claude-plugin/plugin.json` — n/a
- [ ] Version bump — n/a, patch-level bugfix, no release cut in this PR
- [ ] Documentation updated — n/a, internal validator only
- [ ] CHANGELOG.md updated — left to release PR per repo convention

## Testing
- `bash -n scripts/lib/utils.sh tests/unit/test-agent-command-validation.sh` — clean
- `bash tests/unit/test-agent-command-validation.sh` — 19/19 passing, including 3 new cases added by this PR:
  - allows the exact `--base-url .../--api-key-env ...` command shape from the bug report
  - rejects a `--base-url` containing shell metacharacters
  - rejects an `--api-key-env` containing a space
- Full `make test-unit` did not finish within the time available in this run (no output before the sandbox's 300s cap was hit) — this is a single, narrowly-scoped whitelist addition to one validator function with no other files touched, so the directly affected suite above was used as the test surface per repo convention for small changes.

## Related Issues
Closes #659

---
_Generated by [Claude Code](https://claude.ai/code/session_01WdF1xbDThQB2QinHHgRMNj)_

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom API base URL and selecting an API key via an environment variable for OpenAI-compatible agents.
  * Added validation for these new options to ensure only safe, correctly formatted values are accepted.
* **Bug Fixes**
  * Improved protections against malformed or unsafe configuration inputs (e.g., command substitution or invalid environment variable names).

### DIFF
--- a/scripts/lib/utils.sh
+++ b/scripts/lib/utils.sh
@@ -214,7 +214,7 @@ _validate_openai_compatible_agent_command() {
                 ;;
             --base-url)
                 [[ -z "$base_url" ]] || return 1
-                [[ "$value" =~ ^https?://[A-Za-z0-9._/:-]+$ ]] || return 1
+                [[ "$value" =~ ^https?://[A-Za-z0-9][A-Za-z0-9.-]*(:[0-9]+)?(/[A-Za-z0-9._/:-]*)?$ ]] || return 1
                 base_url="$value"
                 ;;
             --api-key-env)
--- a/scripts/lib/utils.sh
+++ b/scripts/lib/utils.sh
@@ -188,6 +188,8 @@ _validate_openai_compatible_agent_command() {
     local cwd=""
     local reasoning_effort=""
     local reasoning_policy=""
+    local base_url=""
+    local api_key_env=""
     local i=1
 
     while [[ $i -lt ${#parts[@]} ]]; do
@@ -209,6 +211,16 @@ _validate_openai_compatible_agent_command() {
                 [[ -z "$cwd" ]] || return 1
                 _octopus_is_safe_openai_compatible_value "$value" || return 1
                 cwd="$value"
+                ;;
+            --base-url)
+                [[ -z "$base_url" ]] || return 1
+                [[ "$value" =~ ^https?://[A-Za-z0-9._/:-]+$ ]] || return 1
+                base_url="$value"
+                ;;
+            --api-key-env)
+                [[ -z "$api_key_env" ]] || return 1
+                [[ "$value" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || return 1
+                api_key_env="$value"
                 ;;
             --reasoning-effort)
                 [[ -z "$reasoning_effort" ]] || return 1
--- a/tests/unit/test-agent-command-validation.sh
+++ b/tests/unit/test-agent-command-validation.sh
@@ -84,14 +84,21 @@ else
 fi
 
 test_case "validate_agent_command rejects unsafe base-url"
-if validate_agent_command "$PROJECT_ROOT/scripts/helpers/openai-compatible-agent.py --provider generic --base-url 'https://example.com/\$(whoami)' --api-key-env MY_KEY --model deepseek-v4-pro --cwd /tmp/test" >/dev/null 2>&1; then
+if validate_agent_command "$PROJECT_ROOT/scripts/helpers/openai-compatible-agent.py --provider generic --base-url https://example.com/\$(whoami) --api-key-env MY_KEY --model deepseek-v4-pro --cwd /tmp/test" >/dev/null 2>&1; then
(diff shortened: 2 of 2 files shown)


nyldn:
Both flagged in 0b49890a:

1. Tightened `--base-url` to require a non-empty host (`^https?://[A-Za-z0-9][A-Za-z0-9.-]*(:[0-9]+)?(/[A-Za-z0-9._/:-]*)?$`, per the suggested diff) so `https:///` / `https://:` are now rejected, plus a dedicated test case for it.
2. Fixed the two negative tests that were quoting their unsafe values — `validate_agent_command` splits its input with `read -a`, which doesn't strip shell quotes, so the tests were incidentally passing on the literal quote characters rather than the metacharacters/whitespace they meant to target. Now use an unquoted base-url (only `$` escaped) and a single-token `MY-KEY` identifier so both actually exercise the validators.

20/20 in the targeted suite.

---

PR #663, opened 2026-07-24
fix: add PID+RANDOM entropy to default spawn task_id

### LINKED ISSUE #661: Same-second spawns share a timestamp task_id — result files interleave and outputs get attributed to the wrong provider
**Plugin Version:** 9.54.1
**Claude Code Version:** 2.1.218

### What happened?

Task IDs are derived from a second-resolution timestamp (`date +%s`). When two `spawn` calls start within the same second, both agents get the same `task_id` and share `${RESULTS_DIR}/.tmp-<id>.out` / `.raw-<id>.out` / `<provider>-<id>.md`. Their outputs interleave: in my case a failing gemini agent's result file ended up containing the *codex* agent's full answer (plus codex tokens-used footer), which made the gemini seat look healthy for weeks while it was actually failing on every dispatch.

This is nasty because it doesn't error — it silently mislabels one provider's output as another's, corrupting debate/council transcripts and provider telemetry.

### What did you expect?

Concurrent spawns to get unique task IDs and non-overlapping result files.

### Steps to reproduce

```bash
for advisor in codex gemini; do
  ./scripts/orchestrate.sh spawn "$advisor" "some prompt" &
done
wait
# Both result files share the same numeric task id; contents interleave.
```

### Suggested fix

Add sub-second entropy and/or PID to the task id, e.g. `task_id="$(date +%s)$$${RANDOM}"`, or `mktemp`-style unique result paths. Any of these keeps the current naming scheme readable while making collisions practically impossible.

Workaround meanwhile: stagger spawns by ≥1s.

nyldn:
## Description
`spawn_agent()` (`scripts/lib/spawn.sh:139-143`) and `spawn_agent_capture_pid()` (`spawn.sh:1122-1125`) both default an unsupplied `task_id` to plain `date +%s` (second resolution). The CLI `spawn` subcommand (`orchestrate.sh:2791`, `spawn_agent "$1" "$2"`) and a couple of `auto-route.sh` call sites rely on that default and pass no explicit `task_id`.

Two spawns that start within the same wall-clock second — e.g. two concurrent `./scripts/orchestrate.sh spawn <agent> <prompt> &` invocations, exactly the reproduction in the bug report — get the identical `task_id`. Critically, the per-run temp files at `spawn.sh:642-644` (`.tmp-${task_id}.out`, `.tmp-${task_id}.err`, `.raw-${task_id}.out`) are keyed on `task_id` alone, with no `agent_type` prefix (unlike the final `.md` result file, which does include `agent_type`). So two different providers spawned in the same second write to the exact same temp file paths, and one provider's output can silently end up attributed to the other's result — the exact corruption described in the issue (a gemini result file containing codex's answer).

I confirmed the collision surface is limited to the two identified default-generation sites — the other `task_id` producers in the codebase (`workflows.sh`, `review.sh`, `yaml-workflow.sh`, etc.) always suffix a per-item index onto a shared `task_group` timestamp, so they don't collide even when many subtasks share a `date +%s` prefix.

## Change
Append `$$${RANDOM}` (PID + a fresh pseudo-random draw) to both default-`task_id` expressions, matching the reporter's own verified-locally suggestion. Explicit `task_id`s passed by every other caller are untouched.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests pass: `bash tests/unit/test-openclaw-compat.sh` (32/32, unrelated to this change but green)
- [ ] OpenClaw registry in sync — n/a, no skill/command changes
- [ ] New skills/commands registered in `.claude-plugin/plugin.json` — n/a
- [ ] Version bump — n/a, patch-level bugfix
- [ ] Documentation updated — n/a, internal task_id generation only
- [ ] CHANGELOG.md updated — left to release PR per repo convention

## Testing
- `bash -n scripts/lib/spawn.sh tests/unit/test-spawn-agent-capture-pid.sh` — clean
- `bash tests/unit/test-spawn-agent-capture-pid.sh` — 4/4 passing, including a new regression case that freezes `date` to a constant value and asserts two consecutive default-`task_id` calls no longer collide
- Verified the new test actually catches the bug: reverted just the `spawn_agent_capture_pid` fix locally, re-ran the suite, and confirmed the new case fails with `expected distinct default task_ids, got '1234567890' and '1234567890'`; restored the fix and it's green again
- Full `make test-unit` did not finish within the time available in this run — targeted suite + syntax checks used instead, as with the sibling fix in #662

## Related Issues
Closes #661

---
_Generated by [Claude Code](https://claude.ai/code/session_01WdF1xbDThQB2QinHHgRMNj)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic task ID generation used when starting and tracking agents by using a more collision-resistant default when no task ID is provided.

* **Tests**
  * Added regression coverage to validate the default task ID format, non-empty behavior, and uniqueness across concurrent start flows, including end-to-end verification that agent tracking uses the shared default logic.

### DIFF
--- a/scripts/lib/spawn.sh
+++ b/scripts/lib/spawn.sh
@@ -120,17 +120,22 @@ _octopus_agent_lifecycle_event() {
 # PID across every subshell forked from it (bash(1)), confirmed failing in
 # CI on macOS. mktemp's file creation is atomic at the OS/filesystem level —
 # genuinely collision-free, not just low-probability — and needs no bash
-# version at all, so use its generated suffix instead; fall back to the old
-# PID/RANDOM combination only if mktemp itself is unavailable. Shared by
-# spawn_agent() and spawn_agent_capture_pid() so both default paths — and
-# their tests — stay in sync from one definition.
+# version at all, so use its generated suffix instead. The reservation file
+# is deliberately kept (not rm'd): deleting it would free that exact name
+# for reuse by a later mktemp call, turning the "genuinely unique" guarantee
+# back into a probabilistic one. Each file is 0 bytes, so the accumulated
+# cost over a session is a handful of empty inodes, not meaningful disk use.
+# Falls back to the old PID/RANDOM combination only if mktemp itself is
+# unavailable — that path is not a hard uniqueness guarantee, only a
+# best-effort default for environments where mktemp can't run at all.
+# Shared by spawn_agent() and spawn_agent_capture_pid() so both default
+# paths — and their tests — stay in sync from one definition.
 _octopus_next_spawn_task_id() {
+    local _reservation_dir="${WORKSPACE_DIR:-${HOME}/.claude-octopus}/.octo/task-ids"
+    mkdir -p "$_reservation_dir" 2>/dev/null || true
     local _tmp _uniq
-    _tmp=$(mktemp "${TMPDIR:-/tmp}/octo-task.XXXXXX" 2>/dev/null) && {
-        _uniq="${_tmp##*.}"
-        rm -f "$_tmp" 2>/dev/null || true
-    }
-    printf '%s-%s\n' "$(date +%s)" "${_uniq:-${BASHPID:-$$}-${RANDOM}}"
+    _tmp=$(mktemp "${_reservation_dir}/XXXXXX" 2>/dev/null) && _uniq="${_tmp##*/}"
+    printf '%s-%s\n' "$(date +%s)" "${_uniq:-${BASHPID:-$$}${RANDOM}}"
 }
 
 write_agent_result_header() {
--- a/scripts/lib/spawn.sh
+++ b/scripts/lib/spawn.sh
@@ -112,6 +112,27 @@ _octopus_agent_lifecycle_event() {
     ) >>"$hook_log" 2>&1 || true
 }
 
+# Default task_id for spawns that don't supply one (#661). An earlier version
+# of this used ${BASHPID:-$$}: BASHPID needs bash 4+ (docs/CONTRIBUTING.md's
+# floor is 3.2, e.g. macOS's system /bin/bash), and on the $$ fallback path
+# two spawn_agent/spawn_agent_capture_pid calls backgrounded with `&` from
+# the same shell still collided — $$ stays pinned to the top-level shell's
+# PID across every subshell forked from it (bash(1)), confirmed failing in
+# CI on macOS. mktemp's file creation is atomic at the OS/filesystem level —
+# genuinely collision-free, not just low-probability — and needs no bash
+# version at all, so use its generated suffix instead; fall back to the old
+# PID/RANDOM combination only if mktemp itself is unavailable. Shared by
+# spawn_agent() and spawn_agent_capture_pid() so both default paths — and
+# their tests — stay in sync from one definition.
+_octopus_next_spawn_task_id() {
+    local _tmp _uniq
+    _tmp=$(mktemp "${TMPDIR:-/tmp}/octo-task.XXXXXX" 2>/dev/null) && {
+        _uniq="${_tmp##*.}"
+        rm -f "$_tmp" 2>/dev/null || true
+    }
+    printf '%s-%s\n' "$(date +%s)" "${_uniq:-${BASHPID:-$$}-${RANDOM}}"
+}
+
 write_agent_result_header() {
     local result_file="$1"
     local agent_type="$2"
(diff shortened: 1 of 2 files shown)


nyldn:
Both findings were real. Fixed in e33a2309:

1. **`$$` doesn't differentiate concurrent forked children.** You're right — `$$` stays pinned to the top-level shell's PID across every subshell forked from it, so two `spawn_agent_capture_pid ... &` calls from the same process got an identical `$$` component, leaving only `$RANDOM`'s ~15 bits between them. Switched to `${BASHPID:-$$}`, which is the exact idiom `validation.sh` already uses for this same reason (see the comment at `validation.sh:410`) — kept the `$$` fallback for the bash 3.2 support floor in `docs/CONTRIBUTING.md` (`BASHPID` is bash 4+). Didn't build a separate atomic allocator since this single-token swap directly fixes the identified weakness and matches the codebase's existing convention rather than adding a new mechanism.

2. **Test wasn't concurrent and bypassed `spawn_agent`'s own default.** Also right on both counts. Reworked the regression test to actually background both calls and `wait` on them, and added a second case that extracts and calls `spawn_agent`'s literal default-generation lines directly (via `sed`, not a hand-copied reimplementation) so `spawn.sh:140` gets real coverage too.

One thing I found while fixing this: neither version of the concurrent test reliably caught the `$$`-collision bug by itself, because `$RANDOM` usually differs by chance even when `$$` doesn't. Ran the old `$$`-only code through the new concurrent test 3x and it passed on luck alone every time. Neutralized `$RANDOM` (`unset RANDOM; RANDOM=42` — per bash(1), it loses its special per-reference behavior once unset+reassigned) so only the PID component can distinguish the two calls. Confirmed 3/3 deterministic failures against the old `$$`-only code and 3/3 deterministic passes with the `BASHPID` fix.

---

nyldn:
CI just caught a real gap: `Unit Tests (macos-latest)` failed with both concurrent default `task_id`s identical (`12345678902615842`/`12345678902615842`). Root cause is exactly what the previous commit's own message flagged as a residual risk — macOS's system bash has no `BASHPID` (bash 4+ only; the repo's floor is 3.2 per `docs/CONTRIBUTING.md`), so `${BASHPID:-$$}` fell back to `$$`, which stays pinned to the top-level shell's PID across every subshell forked from it and doesn't differentiate two `&`-backgrounded calls from the same process.

Fixed in 9d14cbbb by switching to `mktemp`'s generated suffix instead of PID+RANDOM concatenation. `mktemp`'s file creation is atomic at the filesystem level — a genuine uniqueness guarantee, not just low collision probability from `$RANDOM`'s ~15 bits — and needs no bash version at all, so it behaves identically on bash 3.2 and 5.x. This also gives you the "guaranteed-unique source" your earlier review asked for, without a bespoke cross-process allocator.

Also consolidated the previously-duplicated default-generation logic into one shared `_octopus_next_spawn_task_id()` helper (addresses the sed-line-range fragility nitpick too — the test now calls the real named function directly).

Verified by simulating bash 3.2 locally (`env -u BASHPID bash -c ...`) and confirming two concurrent calls now produce distinct ids; local suite is 7/7, including a new test that 20 consecutive calls never repeat.

---

nyldn:
Both fair. Fixed in 3cf67ff1:

1. **Reservation was released on delete.** Correct — `rm -f`ing the mktemp file right after reading its name freed that exact name for a later mktemp call to reuse, which is a probabilistic guarantee again, not a real one. Stopped deleting it; each reservation is now a permanent 0-byte marker (cost over a session: a handful of empty inodes). Didn't build a lifecycle-spanning reservation tied to spawn completion — that needs new cleanup plumbing across every exit path (including crashes) for a residual risk that was already astronomically low, which felt like meaningfully more scope than this bug warrants. A permanent marker gets the same "genuinely unique, not just probabilistically unique" property with a one-line change. Also moved the reservations out of bare `/tmp` into `${WORKSPACE_DIR:-~/.claude-octopus}/.octo/task-ids`, matching the existing `.octo/agents` done-marker convention, so they're namespaced instead of anonymous files loose in `/tmp`.

2. **Fallback path had no coverage.** Right, added a case that stubs `mktemp` to fail and confirms the PID/RANDOM fallback still returns a validly-shaped id. Also normalized its shape to drop an extra internal dash so it matches the mktemp path's `timestamp-suffix` format instead of producing a 3-field id.

Test suite now sets `WORKSPACE_DIR` to its own sandbox so the new reservation files don't land in a real `~/.claude-octopus`; verified none leak there and none survive in `/tmp` after cleanup. 8/8 locally.

---

PR #667, opened 2026-07-26
fix(council): per-seat configurable timeouts + salvage finished-but-timed-out reviews

klass-723:
## Problem

Downstream sail-cruisey #2077 surfaced two council-runner reliability gaps on large-diff reviews:

1. **One global per-seat timeout is too tight.** A slow provider (notably `agy`) on a big diff routinely hits the single `OCTOPUS_COUNCIL_AGENT_TIMEOUT` (120s default). There's no way to give one provider more room without loosening every seat.
2. **A finished review is discarded as a "shortage."** The advice phase treated dispatch as strict pass/fail. When a seat wrote a **complete, `VERDICT:`-bearing** response but its dispatch was killed *right at* the timeout boundary, the usable verdict was thrown away and counted as a missing provider — turning a real quorum into a false shortage.

## Fix

- **`council_seat_timeout`** resolves most-specific-first, so one provider can be widened in isolation:
  1. `OCTOPUS_COUNCIL_TIMEOUT_<PROVIDER>` (e.g. `OCTOPUS_COUNCIL_TIMEOUT_AGY=600`)
  2. `--seat-timeout <seconds>` (run-wide flag, new)
  3. `OCTOPUS_COUNCIL_AGENT_TIMEOUT` (legacy global env)
  4. built-in `120s` default
- **`council_response_has_verdict`** detects an explicit `VERDICT:` line as a *finished-writing* signal — distinct from `council_response_verdict`'s fail-safe `REVISE` default (which returns `REVISE` even for a truncated/absent verdict).
- **Confirm-finish-before-shortage:** the advice loop now counts a non-zero dispatch **iff** the response is non-empty, substantive, *and* verdict-bearing. A review the seat actually finished is no longer discarded.
- New **`--seat-timeout <seconds>`** flag (positive-integer validated) + usage entry.

## Tests

- `council_seat_timeout` precedence: default 120 → global env 200 → `--seat-timeout` 300 → per-provider `OCTOPUS_COUNCIL_TIMEOUT_AGY` 600 (while `codex` stays 300).
- `council_response_has_verdict` salvage detection: complete (upper/lower `VERDICT:`) → true; truncated / empty / missing file → false.
- Full council suite: **58 passed, 0 failed**.

No behavior change on the normal path: a zero-rc dispatch still counts exactly as before; the salvage branch only rescues a complete verdict that would otherwise have been dropped.

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-seat dispatch timeouts with precedence: provider-specific override, then `--seat-timeout`, then legacy environment setting, then default.
  * Enhanced advice-phase quorum logic to retain/count responses that include an explicit `VERDICT:` even when dispatch exits due to timeout boundary behavior.
  * Improved council chair-fallback handling and added more per-seat execution details to the generated `summary.json`.

* **Tests**
  * Added unit coverage for timeout precedence and `--seat-timeout` validation.
  * Added unit coverage for `VERDICT:` detection/salvage behavior and stricter chair-fallback rejection of incomplete responses.


klass-723:
⚠️ Merge-sequencing heads-up (not a code issue): #656 ("unify consultative agent dispatch") refactors the same council dispatch line this PR touches — in `council_live_response` it replaces

```
run_agent_sync "$agent_type" "$prompt" "${OCTOPUS_COUNCIL_AGENT_TIMEOUT:-120}" "$persona" "council"
```

with `run_agent_sync_consultative ...`. This PR changes that same line to pass the per-seat timeout (`$(council_seat_timeout "$agent_type")`) instead of the hardcoded `${OCTOPUS_COUNCIL_AGENT_TIMEOUT:-120}`.

Whoever merges second needs a one-line reconcile: the consultative primitive should take its timeout from `council_seat_timeout` rather than the hardcoded default. No functional overlap otherwise — #656 is workspace-isolation, this is per-seat timeouts. Happy to rebase whenever the other lands.

nyldn:
Integrated current main plus the stacked #666/#670 Council seat-audit and chair-quorum fixes. The timeout path now validates positive-integer overrides, salvages only non-empty substantive completions (and requires an explicit verdict after a non-zero dispatch), preserves truncated partial artifacts without counting them, and keeps the chair excluded from independent-provider quorum. Focused Council suite: 61/61 passing; shell syntax and diff checks pass.

nyldn:
Refreshed onto the merged #670 squash and added one cross-PR correctness guard: a preserved timed-out/degenerate artifact can no longer masquerade as a recovered chair, and an empty-success fallback dispatch cannot suppress the real fallback. The advice/fallback acceptance contract is now consistently non-empty + substantive, with explicit-verdict salvage after non-zero dispatch. Council suite: 62/62 passing; syntax, diff, and mode checks pass.

nyldn:
Addressed the latest review in 65697aa6: the live runner guard now checks run_agent_sync_consultative, changelog wording matches all non-zero dispatch salvage, and non-chair response counts are derived from responded execution records instead of subtracting a presumed chair. Added a successful reused-member fallback regression and updated the summary invariant test. Council suite: 63/63; syntax and diff checks clean.

PR #669, opened 2026-07-26
fix(council): run advice seats detached so an interrupt can't kill them mid-write

klass-723:
## Problem

Council advice seats run **inline in the council's own process group**: `council_run_advice_phase` → `council_dispatch_member` → `run_agent_sync` (blocking). A SIGHUP/SIGINT/SIGTERM delivered to the council — a Claude Code tool timeout, a user Ctrl-C, an orchestrator-level signal — **propagates to the in-flight provider child**, killing it mid-write. That leaves a torn/half-written response file, and the council then hangs or reports a **false provider shortage** because a seat that was about to succeed vanished.

Surfaced by downstream sail-cruisey #2077 as "seat detachment" — the last of that report's four council-runner reliability items.

## Fix (contained to `council.sh`, serial-but-detached)

New `council_dispatch_member_detached` wraps each **advice** and **chair-fallback** seat:

- Runs the seat in a **signal-isolated, disowned background subshell** — `( trap '' HUP INT; … ) &` then `disown`. `setsid` is deliberately **not** used: it's absent on macOS (util-linux only); `disown` + a HUP/INT trap is the portable equivalent already used by `heartbeat.sh`.
- The seat writes to a `.partial` file and **atomically renames** it into place on completion, then drops a `.done` sentinel carrying the exit code.
- The parent **reaps by polling the `.done` sentinel** (bounded by the existing per-seat timeout + grace), so the exit code is authoritative even if a signal races the synchronous return. A completed seat is never lost; a torn response file can no longer occur.
- **Seats still run one at a time** — this is a reliability change, not a concurrency change (per design review, serial ordering preserved, shared `agent-sync.sh`/`spawn.sh` untouched).
- Escape hatch: `OCTOPUS_COUNCIL_DETACH=0` restores the legacy inline dispatch.

Scope is intentionally the **quorum-critical advice path** (where seat loss produces the #2077 false-shortage/hang). Critique/revision/synthesis rounds remain inline; converting them is a low-risk follow-up but out of scope here to keep the diff tight.

## Verification

- Happy path: complete verdict-bearing response lands atomically at the final path, rc 0, no `.partial`/`.done` residue.
- Failing seat: inner exit code propagated through the sentinel; no torn file.
- **Signal isolation**: a seat mid-run ignores `SIGINT`+`SIGHUP` to its process and still lands its result (test uses `sh -c 'echo $PPID'` for portable seat-pid capture — `$BASHPID` is absent in macOS bash 3.2).
- Escape hatch: `OCTOPUS_COUNCIL_DETACH=0` uses the inline path (no sentinel).
- Existing fixture-based quorum tests now exercise the detached path unchanged. **Full council suite: 59 passed, 0 failed.** No file-mode drift.

## Summary by CodeRabbit

* **Bug Fixes**
  * Council advice and chair-fallback seat dispatch now survives signal interrupts during output generation, preventing truncated or corrupted results.
  * Seat outputs are written using an atomic workflow (temporary “partial” output renamed to final) with correct exit-code handling.
  * Added a compatibility option (`OCTOPUS_COUNCIL_DETACH=0`) to restore legacy inline dispatch behavior.
* **Tests**
  * Added unit tests covering atomic publishing, exit-code propagation, cleanup of partial artifacts, signal resilience, compatibility-mode behavior, and timeout-based cancellation.


klass-723:
⚠️ Merge-sequencing heads-up (not a code issue): #656 ("unify consultative agent dispatch") refactors the same council dispatch call this PR builds on — in `council_live_response` it replaces `run_agent_sync ...` with `run_agent_sync_consultative ...`.

This PR's detached-seat wrapper (`council_dispatch_member_detached`) sits one level up (around `council_dispatch_member`), so the two compose cleanly: the detached subshell simply invokes whichever dispatch primitive lands. Expect a small conflict in `council.sh`/`test-council-command.sh` for the second-to-merge; the reconcile is mechanical (run the consultative primitive inside the detached subshell). No functional overlap — #656 is disposable-workspace isolation, this is signal-isolated seat survival. Happy to rebase whenever the other lands.

nyldn:
Integrated the completed Council stack (#666/#670/#667) and current main. Detached reaping now uses the same per-provider timeout resolver as inner dispatch, complete verdict-bearing boundary completions remain salvageable, failed/partial outputs do not count, chair fallback requires a substantive completion, and detached process-tree/atomic-publication guarantees remain intact.

Focused Council suite: 65/65 passing; syntax, diff, and mode checks pass.

@coderabbitai review

nyldn:
Integrated the final #667 chair-fallback integrity fix without regressing detached execution. The combined Council suite passes 66/66, including detached signal survival plus rejection of partial/empty chair responses. This branch will receive the #667 squash commit once that protected merge lands.

nyldn:
The latest #667 review fixes are now integrated in e510b0c2. Full combined Council suite passes 67/67; syntax and diff checks clean.

nyldn:
Merged the protected #667 squash commit (92fde770) into this branch and retained detached dispatch at both advice and chair-fallback call sites. Post-merge Council suite passes 67/67; syntax, diff, and mode checks clean.

PR #671, opened 2026-07-26
fix: validate OpenAI-compatible runtime arguments

Jhacarreiro:
## Problem

PR #646 added explicit OpenAI-compatible runtime arguments to the dispatch command:

    --base-url <http(s) endpoint>
    --api-key-env <credential variable name>

The strict helper-command validator was not updated at the same time. As a result, a supervised Tangle run completed design review and decomposition, then failed at implementation launch with:

    Invalid agent command returned
    spawn_agent produced no provider PID

The generated command was valid and had already passed provider preflight, but `validate_agent_command()` rejected both new flags as unknown.

## Change

Extend the strict OpenAI-compatible helper validator to accept the runtime arguments while preserving the existing injection boundary:

- `--base-url` is optional, unique, safe-token validated, and restricted to `http://` or `https://`;
- `--api-key-env` is optional, unique, and restricted to a valid environment-variable name;
- duplicate flags remain rejected;
- unknown flags and shell metacharacters remain rejected;
- legacy helper commands without these options remain supported.

## Validation

- exact command that failed in the supervised run: accepted;
- configured runtime args: accepted;
- non-HTTP URL: rejected;
- unsafe URL/metacharacters: rejected;
- invalid credential env name: rejected;
- duplicate runtime args: rejected;
- full agent command validation suite: 21/21;
- shell syntax check;
- git diff --check.

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAI-compatible agents with custom `--base-url` and `--api-key-env`.

* **Bug Fixes**
  * Improved `--base-url` validation to allow only `http`/`https` URLs without unsafe content.
  * Improved `--api-key-env` validation to require a safe environment variable name.
  * Updated validation to reject duplicate `--base-url` and duplicate `--api-key-env` arguments.

* **Tests**
  * Expanded unit coverage for accepted and rejected OpenAI-compatible command invocations.

### DIFF
--- a/scripts/lib/utils.sh
+++ b/scripts/lib/utils.sh
@@ -204,10 +204,8 @@ _validate_openai_compatible_agent_command() {
             --base-url)
                 [[ -z "$base_url" ]] || return 1
                 _octopus_is_safe_openai_compatible_value "$value" || return 1
-                case "$value" in
-                    http://*|https://*) base_url="$value" ;;
-                    *) return 1 ;;
-                esac
+                [[ "$value" =~ ^https?://[^/?#]+ ]] || return 1
+                base_url="$value"
                 ;;
             --api-key-env)
                 [[ -z "$api_key_env" ]] || return 1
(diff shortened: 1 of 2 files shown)


Jhacarreiro:
Closing in favor of upstream PR #662, which fixes the same #659 validator/dispatch contract mismatch and applies stricter URL validation. Thanks to the reviewers; no further work is planned on this duplicate branch.

PR #674, opened 2026-07-27
guard: require clean Git baseline for Tangle

Jhacarreiro:
## Problem

A Git worktree can only provide a deterministic run baseline when the source checkout is unambiguous. Starting an implementation run from a dirty checkout silently excludes local tracked and untracked changes from the run branch, or risks mixing unrelated work into later reconciliation.

## Change

Require a clean Git baseline for real Tangle implementation runs:

- reject modified tracked files;
- reject untracked files;
- ignore files covered by normal Git ignore rules;
- reject non-Git directories;
- report each blocking porcelain status entry;
- enable the guard by default from `orchestrate.sh`;
- keep direct unit-test sourcing opt-in to avoid hidden fixture changes.

No dirty snapshot or automatic stash behavior is introduced.

## Validation

- clean baseline suite: 6/6;
- direct-fallback suite: 4/4;
- shell syntax checks;
- `git diff --check`.

This PR is intentionally independent of the run-worktree PR. If both merge, the guard must execute against the source checkout before the isolated run worktree is created.

## Summary by CodeRabbit

* **New Features**
  * Tangle development now defaults to requiring a clean Git baseline before starting, failing early on modified tracked files, untracked files, or non-Git workspaces.
  * Added `OCTOPUS_TANGLE_REQUIRE_CLEAN_BASELINE` (defaults to `true`) to opt out; ignored files are allowed.
* **Bug Fixes**
  * Tangle now exits immediately after `validate_tangle_results` fails, preventing any subsequent contextual review steps.
* **Tests**
  * Added unit coverage for clean, tracked changes, untracked files, ignored files, and non-repository scenarios, plus default-flag behavior.
  * Added a regression test ensuring contextual review does not run after validation abort.


nyldn:
Refreshed onto current main after #672/#666 merged and addressed the latest review: the direct validation-abort test now explicitly opts out of the caller-level clean-baseline guard so it always reaches the intended validation path. Clean-baseline suite 7/7 and validation-abort suite 3/3 pass; prior multi-entry status coverage remains in place.

nyldn:
Pre-integrated current protected main through #667 in 46add403. Clean-baseline 7/7 and validation-abort 3/3 pass; diff and mode checks clean. A final current-main merge will follow after #669 lands.

nyldn:
Final current-main integration is now in 10fdc402, including the protected #669 squash (e0f59b9e). Clean-baseline 7/7 and validation-abort 3/3 pass; diff and mode checks clean.

PR #683, opened 2026-07-27
fix(agy): salvage a TTY-demanding agy seat under a pseudo-terminal

klass-723:
## Problem

agy (Antigravity CLI) is a [bubbletea](https://github.com/charmbracelet/bubbletea) TUI app. In `--print` mode it's headless, but when it must render an **interactive** screen anyway — a folder-trust prompt on a brand-new worktree, or an auth/login flow — it opens `/dev/tty`. In a session with **no controlling terminal** (the normal case for an autonomous/background council run), that fails with:

```
bubbletea: could not open TTY
```

…before agy produces any answer, so the council seat silently dies. Observed on a first-touch council dispatch into a fresh git worktree.

`--print` + `--dangerously-skip-permissions` are already passed unconditionally (verified — no branch drops them), and the common headless path works fine; this only bites when agy insists on rendering a TUI it would otherwise auto-dismiss.

## Fix

`agy-exec.sh` now:
- **Detects** the specific `could not open TTY` / bubbletea failure in agy's captured stderr/stdout.
- **Retries once under a pseudo-terminal** via `script`, giving the auto-dismissed TUI a terminal to draw on so agy can proceed to the real answer.
- **Gated on the real error** — a blanket "wrap whenever there's no TTY" would fire on *every* autonomous dispatch (which never has a TTY) and leak the pseudo-terminal's control-char echo into the answer. The direct headless path is completely untouched.
- **Cleans the PTY echo** (`\r` / backspace / EOT / a leading run of caret-notation like `^D`) so the salvaged answer is byte-clean for the council response parser.
- **Portable**: BSD `script /dev/null <command…>` (trailing argv — safe for a large `--print` value) and util-linux `script -c` (shell-quoted argv via `printf %q`).
- **Opt out** with `OCTOPUS_AGY_NO_PTY_FALLBACK=1`.

## Verification

- A stub agy that dies `bubbletea: could not open TTY` with no TTY but answers under one → **salvaged clean** (real `VERDICT: APPROVE`, no `^D` junk, exit 0, PTY path announced).
- `OCTOPUS_AGY_NO_PTY_FALLBACK=1` → the failure propagates (no retry), as expected.
- Direct headless dispatch (the working path) is byte-identical — the fallback never fires without the error.
- **agy suite: 42/42.** No file-mode drift.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes intermittent failures where the Antigravity flow can’t open a TTY by retrying once via a pseudo-terminal when the specific TTY error is detected.
  * Cleans up recovered output to remove terminal/control artifacts that could break parsing.
  * Captures and surfaces the original stderr for troubleshooting.
  * Adds an opt-out setting: `OCTOPUS_AGY_NO_PTY_FALLBACK=1`.
* **Tests**
  * Adds unit coverage for PTY recovery (including clean output) and the opt-out behavior.
* **Documentation**
  * Updates the unreleased changelog with the terminal-retry fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **A council `agy` seat that dies demanding a TTY is salvaged under a pseudo-terminal instead of silently failing.** agy is a bubbletea TUI app; in `--print` mode it is headless, but when it must render an interactive screen anyway — a folder-trust prompt on a brand-new worktree, or an auth flow — it opens `/dev/tty` and, in a session with no controlling terminal, dies with `bubbletea: could not open TTY` before producing an answer, sinking the seat. `agy-exec.sh` now detects that specific failure and retries once under a pseudo-terminal via `script`, so the auto-dismissed (`--dangerously-skip-permissions`) TUI has a terminal to draw on. The fallback is gated on the real error — a blanket no-TTY wrap would fire on every autonomous dispatch (which never has a TTY) and leak the pseudo-terminal's caret-notation echo into the answer — and the salvaged output is stripped of that echo (CR / backspace / EOT / leading caret-notation) so it stays byte-clean for the council response parser. Portable across BSD (`script … command`) and util-linux (`script -c`) `script`; opt out with `OCTOPUS_AGY_NO_PTY_FALLBACK=1`.
+
 ## [9.55.1] - 2026-07-27
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nyldn:
Addressed the remaining portability finding in a85cfe09: util-linux `script -c` now receives POSIX single-quoted arguments instead of Bash `%q`, and the PTY recovery test verifies an exact multiline prompt containing a single quote. Local agy suite: 42/42.

PR #696, opened 2026-07-30
feat: add native Command Code CLI provider

Jhacarreiro:
## Summary
- add a native `commandcode` provider backed by the Command Code CLI
- support API-key or existing CLI authentication without implementing OAuth
- isolate provider credentials and runtime controls
- map read-only roles to plan mode and implementation roles to the Octopus-managed worktree
- add provider detection, routing, model resolution, fleet support, docs, and tests

## Validation
- Command Code harness tests: PASS
- agent command validation: 20/20
- spawn PID capture: 8/8
- credential isolation: 30/30
- smoke suites: 16/16
- live read-only smoke: DeepSeek V4 Pro and MiniMax M3 both PASS

## Notes
The provider keeps Octopus responsible for orchestration and worktree isolation. Command Code runs headlessly with onboarding, auto-update, and persistent sessions disabled.

## Summary by CodeRabbit

* **New Features**
  * Added Command Code CLI support as a provider, including automatic model selection, role-based execution (plan vs implementer), and provider allowlisting integration.
  * Improved provider discovery and health checking for Command Code CLI, with configurable binary/auth and runtime limits.
* **Documentation**
  * Added a Command Code CLI integration guide (installation, API key auth, model/env configuration, headless execution behavior).
* **Bug Fixes**
  * Updated Evidence “Local CI parity” unit suite count to **189** (smoke: 16, integration: 7 unchanged).
* **Tests**
  * Added unit tests for the Command Code CLI harness covering flags, parsing, error handling, and exit codes.

### DIFF
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -77,7 +77,7 @@ Frontier AI models will remain individually overconfident for the foreseeable fu
 **Traction (as of 2026-07-27):**
 - GitHub stars: 3,889
 - GitHub forks: 365
-- Local CI parity: 16 smoke, 187 unit, and 7 integration suites
+- Local CI parity: 16 smoke, 189 unit, and 7 integration suites
 - Version: 9.56.1 (active release cadence)
 - Runtimes supported: Claude Code, Codex CLI, Command Code CLI, Cursor (MCP), Gemini CLI
 
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -77,9 +77,9 @@ Frontier AI models will remain individually overconfident for the foreseeable fu
 **Traction (as of 2026-07-27):**
 - GitHub stars: 3,889
 - GitHub forks: 365
-- Local CI parity: 16 smoke, 188 unit, and 7 integration suites
+- Local CI parity: 16 smoke, 187 unit, and 7 integration suites
 - Version: 9.56.1 (active release cadence)
-- Runtimes supported: Claude Code, Codex CLI, Cursor (MCP), Gemini CLI
+- Runtimes supported: Claude Code, Codex CLI, Command Code CLI, Cursor (MCP), Gemini CLI
 
 **Measured Impact:**
 - 75% consensus gate: quantifiable disagreement detection before production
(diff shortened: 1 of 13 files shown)


Jhacarreiro:
Addressed all six review findings in c888d8ff:

- added Command Code to diverse fleet ordering
- detect authenticated `cmd login` sessions as well as API keys
- enabled strict mode while preserving CLI exit codes
- structured Command Code errors now fail even when the CLI exits 0
- kept verifier dispatch in `plan` mode
- migrated the harness tests to the repository unit-test framework

Validation after rebase onto current `main`:
- Command Code harness: 6/6
- agent command validation: 20/20
- spawn PID capture: 8/8
- credential isolation: 30/30
- isolated Ollama smoke: 16/16
- full smoke workflow is green in CI

All review threads are resolved. Please re-review the updated head.

nyldn:
Merging. One blocking defect found and fixed on this branch before merge (rebased onto current main, which now carries #702).

**`validate_agent_command` rejected every commandcode dispatch.** `get_agent_command` returns the shim *with arguments*:

```
…/scripts/helpers/commandcode-exec.sh deepseek/deepseek-v4-pro plan
```

That matched no arm of the whitelist and fell through to the reject arm, so every commandcode dispatch would have aborted the phase before the CLI was ever invoked. Same failure mode as #697 (copilot-exec.sh) and #705 (agy-exec.sh). Reproduced directly against this branch before fixing.

Fix: added the shim to the executable-token check introduced in #702, so the path is accepted as the command's first token but still rejected when embedded later in a command string.

**Why CI did not catch it.** `test-commandcode-harness.sh` asserted the *shape* of `get_agent_command`'s output but never ran it through `validate_agent_command`. Added that assertion, plus the accepted/embedded pair in `test-agent-command-validation.sh` matching the other shims. Both fail when the allowlist entry is removed and pass with it.

Everything else in the PR verified as-is: both provider-routing sites, the model resolver arm ordering, the isolated `env -i` provider environment, and the JSON result parsing in the shim. The `yolo` permission mode for implementer/developer roles is consistent with existing providers (gemini and qwen pass `--approval-mode yolo` unconditionally; codex defaults to `--sandbox workspace-write`), and is in fact narrower since it is role-scoped. Filing a follow-up for the one gap: commandcode has no env override for permission mode, where codex has `OCTOPUS_CODEX_SANDBOX`.

Thanks for the contribution.

PR #706, opened 2026-08-01
fix: fail loud when a global flag after the subcommand is read as the prompt

### LINKED ISSUE #698: Global flags after a subcommand are silently consumed as the prompt: 'define --timeout 540 "..."' runs a workflow whose task is literally '--timeout'
## Summary

Global flags placed **after** a subcommand are consumed as the positional prompt. The real prompt is silently discarded and the workflow runs on the flag name as its task.

There is no error and no warning. The run completes "successfully" having answered a question nobody asked, and the output looks legitimate.

Verified on **v9.56.1**, macOS 26.

---

## Reproduction

`--dry-run` is enough — no provider spend needed:

```bash
# A — flag AFTER subcommand
$ orchestrate.sh define --timeout 540 "REAL PROMPT HERE" --dry-run
INFO: [DRY-RUN] Would grasp: --timeout          # <-- task is the flag name

# B — flag BEFORE subcommand
$ orchestrate.sh --timeout 540 define "REAL PROMPT HERE" --dry-run
INFO: [DRY-RUN] Would grasp: REAL PROMPT HERE   # <-- correct
```

In form A, `540` and `"REAL PROMPT HERE"` are both dropped.

---

## Impact in a real run

I hit this on a live `define` run. Three providers spent several minutes producing a full consensus artifact for the task `--timeout`. The generated file opens:

```markdown
# GRASP Phase - Problem Definition Consensus
## Task: --timeout
```

…followed by ~6 KB of genuinely coherent analysis about designing deadline and timeout management for distributed systems. Every provider "succeeded". Nothing in the logs indicated the prompt had been lost.

The failure mode is the concerning part: **it is not a crash, it is confidently wrong work**. Had I not checked the `## Task:` line I would have read that synthesis as the answer to my actual question.

---

## Why this is easy to hit

`-t, --timeout SECS` is documented in `help --full` as a normal option:

```
  -t, --timeout SECS      Timeout per task (default: 600)
```

Nothing states that flags must precede the subcommand. GNU-style tools overwhelmingly accept `cmd subcommand --flag value`, so flag-after-subcommand is the natural thing to type. The parser at `orchestrate.sh:2176` handles `-t|--timeout` correctly — it simply never sees these tokens, because the subcommand branch has already claimed the remaining argv as the prompt.

I was also nudged toward the mistake by the timeout error message itself, which suggests `--timeout 360` without indicating where it goes:

```
💡 Possible solutions:
   1. Increase timeout: --timeout 360 (6m)
```

---

## Suggested fixes

Any one of these would prevent silent wrong-work; the first is the cheapest:

1. **Reject unknown leading tokens.** If the resolved prompt starts with `-` or `--`, error out:
   `ERROR: '--timeout' looks like a flag but was read as the prompt. Global flags must precede the subcommand: orchestrate.sh --timeout 540 define "..."`
2. **Accept flags after the subcommand** by re-scanning argv for known global flags before treating the remainder as the prompt.
3. **Document the ordering** in `help --full` and in the timeout error hint, e.g. `--timeout 360` → `orchestrate.sh --timeout 360 <command> "..."`.

(1) alone converts a silent failure into a loud one, which is the main thing.

---

## Possibly related, possibly separate

On an earlier attempt at the same workflow — form B, no flag involved — a `claude-sonnet` agent died with:

```
ERROR: Agent claude-sonnet failed with exit code 124 (role=researcher, phase=grasp)
ERROR: Operation timed out after 180s (3m)
```

The documented default is 600s, so something in the grasp path appears to be applying 180s. That resembles #408 (*synthesis stage timeout hardcoded to 180s, --timeout flag does not override*), which is closed. I have not isolated it and it may be a different code path or a regression — flagging it here only so the reference is not lost. Happy to open it separately if useful.

---

*Found while running `/octo:define` for a multi-provider review. Reported with Claude Code.*

nyldn:
## Root cause

`orchestrate.sh`'s late-args loop (`scripts/orchestrate.sh:2251-2286`) lets a subset of global flags appear after the subcommand — `-n`/`--dry-run`, `--debug`, `-v`, `-Q`, `-P`, `--tier`, `--provider`, `--cost-first`, `--quality-first`, `--openrouter-nitro`, `--openrouter-floor`. Any other global flag (`--timeout`, `--parallel`, `--branch`, etc.) isn't recognized there, falls through the catch-all `*)` arm, and lands in the prompt's positional slot.

Commands that consume the prompt as `"$1"` (`define`, `develop`, `deliver`) then silently treat the flag name itself as the whole prompt — the real prompt and anything after the flag's value is dropped. Commands using `"$*"` (`discover`/`probe`, `verify`, `embrace`, `squeeze`) embed the stray flag/value into the prompt text instead. Either way there's no error — the run "succeeds" having answered a question nobody asked (closes #698).

## Fix

Added a guard immediately after the late-args loop, scoped to the commands that take a free-text prompt (`discover|research|probe`, `define|grasp`, `verify|verification-only`, `develop|tangle`, `deliver|ink`, `embrace`, `squeeze|red-team`). If the resolved first argument still looks like a flag (starts with `-`, excluding `-h`/`--help`), it errors out naming the flag and pointing at the fix, instead of silently running on it. This is fix option (1) from the issue — "converts a silent failure into a loud one" — the cheapest option and the one the issue calls out as sufficient on its own.

`grapple` and `probe-single` were left untouched — they already have their own local flag parsers for the case they need, and blanket-rejecting a leading `-` there would break legitimate use (e.g. `probe-single --output-dir`).

## Test plan

- New `tests/unit/test-cli-flag-after-subcommand.sh`: reproduces the exact issue repro (`define --timeout 540 "REAL PROMPT HERE" --dry-run` now fails loud instead of running `Would grasp: --timeout`), confirms flags-before-subcommand still work, checks `probe`/`develop` with other unrecognized flags, and confirms `--help` isn't mistaken for the guard.
- `make test-unit`: 185/189 suites pass; the 4 failures (`test-feature-disclosure`, `test-github-work-queue-hook`, `test-hud-smart-mode`, `test-readme-release-sync`) reproduce identically on unmodified `origin/main` — pre-existing, unrelated to this change.
- `make test-integration`: 7/7 suites pass.
- `bash -n` across all shell scripts: clean.
- Manually verified `probe-single --output-dir` (own scoped parser) and `grapple` are unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3GGNpJgxsW1cwoPchFQfD)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow commands now reject unsupported flag-like arguments instead of treating them as prompts.
  * Help requests using `-h` or `--help` continue to work as expected.

* **Tests**
  * Added coverage for misplaced flags across multiple workflow commands and valid flag placement.

* **Documentation**
  * Updated the documented local CI suite count to reflect the latest coverage.

### DIFF
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -77,7 +77,7 @@ Frontier AI models will remain individually overconfident for the foreseeable fu
 **Traction (as of 2026-07-27):**
 - GitHub stars: 3,889
 - GitHub forks: 365
-- Local CI parity: 16 smoke, 189 unit, and 7 integration suites
+- Local CI parity: 16 smoke, 190 unit, and 7 integration suites
 - Version: 9.56.1 (active release cadence)
 - Runtimes supported: Claude Code, Codex CLI, Command Code CLI, Cursor (MCP), Gemini CLI
 
--- a/scripts/orchestrate.sh
+++ b/scripts/orchestrate.sh
@@ -2285,6 +2285,19 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
     done
     set -- "${_late_args[@]}"
 
+    # v9.57.0: A global flag that isn't recognized by the late-args loop above
+    # (e.g. --timeout after the subcommand) lands here unconsumed and would
+    # otherwise be silently read as the prompt (#698) — the run "succeeds"
+    # having answered a question nobody asked. Fail loud instead.
+    case "$COMMAND" in
+        discover|research|probe|define|grasp|verify|verification-only|develop|tangle|deliver|ink|embrace|squeeze|red-team)
+            if [[ "${1:-}" == -* && "${1:-}" != "-h" && "${1:-}" != "--help" ]]; then
+                log ERROR "'$1' looks like a flag but was read as the prompt. Global flags must precede the subcommand: $(basename "$0") $1 ... $COMMAND \"...\""
+                exit 1
+            fi
+            ;;
+    esac
+
 # Check for first-run on commands that need setup (skip for help/setup/preflight)
 if [[ "$COMMAND" != "help" && "$COMMAND" != "setup" && "$COMMAND" != "preflight" && "$COMMAND" != "-h" && "$COMMAND" != "--help" ]]; then
     check_first_run || true  # Show hint but don't block
--- a/tests/unit/test-cli-flag-after-subcommand.sh
+++ b/tests/unit/test-cli-flag-after-subcommand.sh
@@ -43,28 +43,53 @@ else
     test_fail "expected successful dry-run with the real prompt; got exit=$rc, output: $out"
 fi
 
-test_case "probe --parallel after subcommand fails loud"
-out="$(run_orchestrate probe --parallel 3 "topic" --dry-run)" && rc=0 || rc=$?
-if [[ "$rc" -ne 0 && "$out" == *"looks like a flag but was read as the prompt"* ]]; then
+# Table-driven coverage: every command that consumes a free-text prompt and
+# shares the vulnerable late-args loop should get the same guard. grapple and
+# probe-single are deliberately excluded — they run their own local flag
+# parsers instead of falling through to the shared late-args loop.
+GUARDED_COMMANDS=(
+    "discover:--parallel 3"
+    "research:--parallel 3"
+    "probe:--parallel 3"
+    "grasp:--timeout 540"
+    "verify:--branch foo"
+    "verification-only:--branch foo"
+    "develop:--branch foo"
+    "tangle:--branch foo"
+    "deliver:--branch foo"
+    "ink:--branch foo"
+    "embrace:--branch foo"
+    "squeeze:--branch foo"
+    "red-team:--branch foo"
+)
+
+for entry in "${GUARDED_COMMANDS[@]}"; do
+    cmd="${entry%%:*}"
+    flag="${entry#*:}"
+    test_case "$cmd $flag after subcommand fails loud"
+    # shellcheck disable=SC2086 # deliberate split of the "flag value" pair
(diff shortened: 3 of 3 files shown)


nyldn:
CI red on `Unit Tests (ubuntu-latest)`, but it's failing on `main` too — pre-existing, unrelated to this diff.

The single failure is `test-readme-release-sync.sh`: `README sync check failed: STALE: PRODUCT.md`. Verified locally with `git stash` against unmodified `origin/main` (cb2677b2) — same failure, same message. This PR only touches `scripts/orchestrate.sh`'s argument parsing and adds a new test file; it doesn't touch `PRODUCT.md`, `README.md`, or `sync-readme.py`.

Leaving this PR as-is rather than bundling an unrelated `./scripts/sync-readme.py` fix into a single-bug-fix PR. Will re-run once the base branch recovers.

---

nyldn:
Addressed both in 5ba6f47a:
- Table-driven coverage extended to all 13 subcommands sharing the guard (`discover`, `research`, `probe`, `grasp`, `verify`, `verification-only`, `develop`, `tangle`, `deliver`, `ink`, `embrace`, `squeeze`, `red-team`) — `grapple`/`probe-single` stay excluded since they use their own local flag parsers.
- `-h`/`--help` tests now assert a `Usage:` marker is actually present, not just that the flag-guard error is absent.

All 18 cases pass locally.

---

nyldn:
Re: the "Out of Scope Changes" warning on `PRODUCT.md` — that's a false positive, not something to fix. `PRODUCT.md`'s unit-suite count is a generated fact asserted by `tests/unit/test-readme-release-sync.sh`, and this PR adds a new suite file (`test-cli-flag-after-subcommand.sh`). The count bump is `make sync`'s mechanical output, directly required by this PR's own test addition — removing it would fail that check, which is exactly the CI failure earlier in this thread. Leaving `PRODUCT.md` as-is.

---

PR #719, opened 2026-08-02
fix: label design review seats by runtime identity

Jhacarreiro:
## Summary

- replace historical `CODEX` / `GEMINI` / `SONNET` design-review headings with neutral seat labels
- render each seat from the configured executor alias, provider identity, and resolved model
- add native `commandcode* -> commandcode` provider identity mapping
- log resolved seat identities for the review and synthesis agents

## Problem

The design-review ceremony could route a historical slot to a different provider (for example `commandcode-research` / MiniMax) while still labelling the output as `GEMINI APPROACH`. This made synthesis messages and diagnostics incorrectly imply that providers were used when only the slot names were historical.

The runtime identity schema introduced previously already distinguishes executor alias, configured provider/model, and runtime provider/model. The ceremony prompt was one remaining place that still presented slot names as provider identity.

## Behaviour after this change

Example headings now look like:

```text
SEAT 1 - commandcode / deepseek/deepseek-v4-pro (executor: commandcode)
SEAT 2 - commandcode / minimaxai/minimax-m3 (executor: commandcode-research)
SEAT 3 - anthropic / <resolved model> (executor: claude-sonnet)
```

The execution routing itself is unchanged.

## Tests

- `make test-unit` — 190/190 suites passed
- `tests/unit/test-runtime-provider-labels.sh` — 8/8 passed
- `tests/unit/test-ceremonies.sh` — 7/7 passed
- `tests/smoke/test-syntax.sh` — 4/4 passed

## Summary by CodeRabbit

* **Improvements**
  * Design review results now display clear, configurable seat labels.
  * Review summaries identify each response using configured provider, model, and executor details.
  * CommandCode-based review seats are now recognized and labeled consistently.

### DIFF
--- a/scripts/lib/quality.sh
+++ b/scripts/lib/quality.sh
@@ -531,9 +531,24 @@ Be concise and specific. This is a planning exercise, not implementation."
     log INFO "Design review: gathering provider approaches..."
     log INFO "Design review seats: seat_1=${seat_1_label}, seat_2=${seat_2_label}, seat_3=${seat_3_label}, synthesis=${synthesis_label}, timeout=${_design_timeout_label}, synth_timeout=${_synth_timeout_label}"
 
-    seat_1_approach=$(OCTOPUS_UNBOUNDED_EXECUTION_SUPERVISED="design-review-ceremony" run_agent_sync_consultative "$design_codex_agent" "$ceremony_prompt" "$design_timeout" "implementer" "ceremony" 2>/dev/null) || true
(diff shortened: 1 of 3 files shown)


Jhacarreiro:
Addressed the requested change in 2969627e: each ceremony seat now exports OCTOPUS_UNBOUNDED_EXECUTION_SUPERVISED inside an explicit subshell before invoking run_agent_sync_consultative. Focused tests pass; the only full-suite local miss was an unrelated flaky test-tool-policy case, which passes in isolation. CI has been retriggered by the push.

Issue #715, opened 2026-08-02
gemini-via-agy is the migration path off the sunset Gemini OAuth, but ships off and is never offered


nyldn:
Second half of the recurring `gemini-cli-workspace-oauth` keychain prompt. The marker-visibility bug is #714; this is the reason users are still on the dead path at all.

## Current state

`config/features.json`:

```json
{
  "id": "gemini-via-agy",
  "decision": "none",
  "key": "OCTOPUS_GEMINI_VIA_AGY",
  "default": "0",
  "description": "... Gemini Code Assist free-tier OAuth is sunset, so the direct path returns IneligibleTierError. Set OCTOPUS_GEMINI_VIA_AGY=1 to enable."
}
```

`decision: none` means it ships silently on its default and is never raised at session start. The default is `0`. So the plugin knows the direct path is sunset, has a working replacement wired through `dispatch.sh`, `spawn.sh` and `agent-sync.sh`, and ships with it switched off behind an env var that the progressive-disclosure framework has been explicitly told not to mention.

The manifest's own comment says a prompt is for "a genuine fork in behaviour or cost". Choosing between a provider path that can only fail and one that works is exactly that.

## Why this is the wrong default

`check-providers.sh:89` reports gemini available on `command -v gemini` alone. Nothing about a sunset OAuth tier is visible on disk, so with `OCTOPUS_GEMINI_VIA_AGY=0` the seat is offered, dispatched, spawns the gemini CLI, triggers the macOS keychain prompt, and only then fails. Even with #714 merged so the dead-mark sticks, the **first** prompt of any session that reseats gemini still happens, because the dialog precedes the error.

The reactive mark is the right mechanism for a transient quota window. It is the wrong mechanism for a permanently withdrawn auth tier, which is a known-ahead-of-time fact.

## Options

1. **Flip to `decision: required`** and ask on upgrade: route gemini seats through Antigravity, or drop gemini. Matches how `fable5-routing` and `codex-reviewer-flip` are handled, and this is a real cost/behaviour fork (agy backend cost follows `OCTOPUS_AGY_MODEL`).
2. **Default `1` when `agy` is present**, keeping `decision: none`. Fewer prompts, but silently changes which vendor serves a seat, which is the kind of change the disclosure framework exists to surface.
3. **Detect the sunset condition at preflight** rather than at dispatch, so the seat never reaches the keychain. Most correct and most work; needs a probe that does not itself trigger the OAuth read.

[...]

PR #727, opened 2026-08-02
fix: honour legacy truthy/falsy OCTOPUS_REVIEWER_FLIP values

### LINKED ISSUE #720: _octo_reviewer_flip_active: documented legacy OCTOPUS_REVIEWER_FLIP=1 does not work, and the comment says it does
Found while reviewing #715. Pre-existing, introduced by me in #691.

`scripts/lib/agent-utils.sh` states:

```
# Legacy truthy values are accepted so an existing OCTOPUS_REVIEWER_FLIP=1 in
# someone's profile keeps meaning what it used to.
```

That is false on the normal path. Measured with `codex` on PATH and `features.sh` sourced, as `orchestrate.sh` sources it:

```
OCTOPUS_REVIEWER_FLIP=claude  -> FLIPPED
OCTOPUS_REVIEWER_FLIP=1       -> not-flipped
OCTOPUS_REVIEWER_FLIP=on      -> not-flipped
OCTOPUS_REVIEWER_FLIP=true    -> not-flipped
OCTOPUS_REVIEWER_FLIP=yes     -> not-flipped
```

**Cause.** `octo_features_choice` (`scripts/lib/features.sh:220-235`) only passes an env value through when it is a *declared choice* for the feature. `codex-reviewer-flip` declares `claude` and `codex`, so `1`/`on`/`true`/`yes` are not valid choices, get dropped, and resolution falls to the ledger and then the `codex` default. The `case` arm that accepts truthy aliases only ever sees a value on the `features.sh`-absent fallback branch, which is not how the plugin runs.

**Fix**, mirroring what #715 does for `octo_gemini_via_agy_active` — normalise the env var locally, before consulting the ledger, rather than changing `octo_features_choice` (which would alter semantics for every feature):

```bash
_octo_reviewer_flip_active() {
    command -v codex >/dev/null 2>&1 || return 1
    case "${OCTOPUS_REVIEWER_FLIP:-}" in
        claude|1|on|true|yes) return 0 ;;
        codex|0|off|false|no) return 1 ;;
    esac
    declare -f octo_features_choice >/dev/null 2>&1 || return 1
    [[ "$(octo_features_choice "codex-reviewer-flip")" == "claude" ]]
}
```

**Test note.** Any test for this has to execute the resolution path with `features.sh` sourced. Asserting on the `case` arms in isolation, or grepping the source for the variable name, passes while the behaviour is broken — the same weakness that let the `gemini-via-agy` wiring regress undetected in `test-agy-provider.sh`.

Kept out of the #715 PR deliberately: different feature, different choice vocabulary (`claude`/`codex` rather than `1`/`0`), and #715 should stay revertible on its own.

nyldn:
## Root cause

`_octo_reviewer_flip_active` in `scripts/lib/agent-utils.sh:65-80` claims legacy truthy values (`1`/`on`/`true`/`yes`) are honoured for `OCTOPUS_REVIEWER_FLIP`, but they weren't, on the normal dispatch path (with `codex` on `PATH` and `features.sh` sourced, as `orchestrate.sh` does).

`octo_features_choice` (`scripts/lib/features.sh:220-235`) only passes an env value through when it matches one of the feature's *declared* choices. `codex-reviewer-flip` declares `claude`/`codex`, so `1`/`on`/`true`/`yes` aren't valid choices — they get silently dropped, and resolution falls through to the ledger/default (`codex`), leaving the flip inert.

## Fix

Normalize `OCTOPUS_REVIEWER_FLIP` against the legacy vocabulary *before* consulting `octo_features_choice`, rather than routing it through the ledger's choice validation — mirroring the approach the linked issue notes was used for `octo_gemini_via_agy_active` in #715.

## Test coverage gap

The existing test (`tests/unit/test-fable5-escalation.sh`) only exercised `OCTOPUS_REVIEWER_FLIP=1` against the `implementer` role, which is unaffected by the flip either way — so it passed vacuously while the actual `code-reviewer` case was broken. Added 8 regression cases exercising all four truthy and four falsy legacy aliases against `code-reviewer` specifically.

## Test results

```
tests/unit/test-fable5-escalation.sh: 34/34 passed (was 26/26 before this change; +8 new cases)
```

Full `make test-unit` run: 187/190 suites passed. The 3 pre-existing failures (`test-feature-disclosure.sh`, `test-github-work-queue-hook.sh`, `test-hud-smart-mode.sh`) reproduce identically on the unmodified branch and are unrelated to `agent-utils.sh` or the reviewer-flip feature — verified by stashing this change and re-running them.

Closes #720.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PCuB7u4yVV8AeT79vEUJbd)_

### DIFF
--- a/scripts/lib/agent-utils.sh
+++ b/scripts/lib/agent-utils.sh
@@ -64,19 +64,22 @@ _BARE_OPT="${_BARE_OPT:-}"
 # vendor diversity instead of creating it.
 _octo_reviewer_flip_active() {
     command -v codex >/dev/null 2>&1 || return 1
+    # Legacy truthy/falsy values are normalized here, before consulting the
+    # features ledger: octo_features_choice only passes an env value through
+    # when it matches one of the feature's declared choices (claude|codex), so
+    # "1"/"on"/"true"/"yes" would otherwise be silently dropped and fall back
+    # to the ledger/default instead of meaning what they used to.
+    case "${OCTOPUS_REVIEWER_FLIP:-}" in
+        claude|1|on|true|yes) return 0 ;;
+        codex|0|off|false|no) return 1 ;;
+    esac
     local choice
     if declare -f octo_features_choice >/dev/null 2>&1; then
         choice="$(octo_features_choice "codex-reviewer-flip")"
     else
-        choice="${OCTOPUS_REVIEWER_FLIP:-codex}"
+        choice="codex"
     fi
-    # "claude" is the only value that moves review off the implementing vendor.
-    # Legacy truthy values are accepted so an existing OCTOPUS_REVIEWER_FLIP=1 in
-    # someone's profile keeps meaning what it used to.
-    case "$choice" in
-        claude|1|on|true|yes) return 0 ;;
-    esac
-    return 1
+    [[ "$choice" == "claude" ]]
 }
 
 get_role_mapping() {
--- a/tests/unit/test-fable5-escalation.sh
+++ b/tests/unit/test-fable5-escalation.sh
@@ -301,6 +301,30 @@ else
     test_fail "flip must not move implementation, got '$got'"
 fi
 
+# Regression for issue #720: octo_features_choice only passes an env value
+# through when it matches a declared choice (claude|codex), so legacy truthy
+# aliases must be normalized against the raw env var, not routed through it,
+# or they get silently dropped and the flip never fires on code-reviewer.
+for legacy in 1 on true yes; do
+    test_case "reviewer flip honours legacy truthy OCTOPUS_REVIEWER_FLIP=$legacy on code-reviewer"
+    got=$(PATH="$flip_bin:$PATH" OCTOPUS_REVIEWER_FLIP="$legacy" bash -c "source '$FEATURES_LIB' 2>/dev/null; source '$AGENT_UTILS' 2>/dev/null; get_role_mapping code-reviewer")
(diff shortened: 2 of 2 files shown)


nyldn:
Superseded by #722, which is the same fix arrived at independently — normalise the env var before consulting the ledger, accepting `claude|1|on|true|yes` and `codex|0|off|false|no`. Both are correct.

Going with #722 on test placement rather than on the fix itself:

- **Same end-to-end depth.** I checked expecting #727 to be stronger here, since it asserts through `get_role_mapping code-reviewer`. #722 does the same thing via `get_role_mapping reviewer`, so both prove the flip actually changes routing rather than just exercising the predicate.
- **Better home.** #722 puts the cases in a dedicated `tests/unit/test-reviewer-flip-legacy-values.sh`. #727 adds them to `test-fable5-escalation.sh`, where someone looking for reviewer-flip coverage will not find them — the suite is about Fable 5 escalation.
- **Fuller matrix.** #722 covers the falsy aliases as well, which matters because the bug cut both ways: `=false` was also being dropped and losing to a recorded `claude`.

One thing #727 has going for it that #722 does not: no new test file, so no `PRODUCT.md` suite-count churn. That is a real cost, and it is the subject of #729 — being fixed in #730 so it stops being a consideration when choosing where to put tests.

Closing as duplicate. Issue #720 will close with #722.

PR #731, opened 2026-08-02
fix: stop churning derived test-suite counts in PRODUCT.md

### LINKED ISSUE #729: Derived test-suite counts in PRODUCT.md make every pair of test-adding PRs conflict
Any two PRs that add a test file conflict on one line, every time. Hit four times in a single session (#706, #722, #723, and #726/#728 against each other).

## Mechanism

`scripts/sync-readme.py:147` derives the counts by globbing the test tree:

```python
test_suite_counts = {
    category: len(list((root / "tests" / category).glob("test-*.sh")))
    for category in ("smoke", "unit", "integration")
}
```

and writes them into tracked `PRODUCT.md`:

```
- Local CI parity: 16 smoke, 194 unit, and 7 integration suites
```

`tests/unit/test-readme-release-sync.sh` then asserts the file matches. So adding one test file is a required edit to a shared line, and two branches doing it always collide — even when the underlying changes are entirely unrelated.

The conflict is also *not* meaningfully resolvable by hand: taking either side gives a wrong number. The only correct resolution is to discard both and re-run `make sync`, which is not obvious from the conflict markers and is easy to get wrong under time pressure.

## Why it is worth fixing rather than living with

It is friction on exactly the behaviour the repo wants to encourage. Every PR that adds coverage pays a merge-conflict tax, and the tax scales with how many such PRs are in flight. It also produces a failure mode that looks alarming but is not: a PR goes red on `STALE: PRODUCT.md` for a reason unrelated to what it changed.

## Options

1. **Stop storing the derived counts.** Drop the line from `PRODUCT.md`, or replace the exact numbers with a stable phrase. The counts are marketing copy, not a contract; nothing depends on the specific integer.
2. **Keep the numbers, remove the conflict.** Add a `.gitattributes` merge driver for `PRODUCT.md` that regenerates rather than merges, or move the generated block to a file with a union merge strategy.
3. **Compute at check time only.** Let CI assert "the count in the file matches the tree" without requiring the file to carry the number between releases — for instance regenerate during the release cut rather than on every sync.

Recommend 1. The precise suite count in a product doc has low informational value, changes constantly, and is the only reason these conflicts exist. If the number is wanted for release notes, option 3 gets it at the moment it actually matters.

Note the same generator writes several other derived facts into the same files (version, component counts, model defaults). Those change rarely, so they do not have this problem; only the test counts churn per-PR.

nyldn:
## Root cause

`scripts/sync-readme.py:147` globbed `tests/{smoke,unit,integration}/test-*.sh` and wrote the exact counts into a tracked line in `PRODUCT.md`:

```
- Local CI parity: 16 smoke, 194 unit, and 7 integration suites
```

`tests/unit/test-readme-release-sync.sh` then asserted the file matched. Since adding a single test file is a required edit to that one shared line, any two PRs that each added a test file collided on it — every time — and the conflict wasn't hand-resolvable: taking either side leaves a wrong number, and the only correct resolution is discarding both and re-running `make sync`, which isn't obvious from the conflict markers.

This isn't hypothetical: while rebasing this branch onto `main` (which had just picked up #726, itself adding a test file), I hit exactly this conflict on `PRODUCT.md` live.

## Fix

Per the issue author's own recommendation (option 1 of three offered — "stop storing the derived counts"):

- `scripts/sync-readme.py`: removed the `test_suite_counts` glob/derivation in `derive_facts()`, the `smoke_suite_count`/`unit_suite_count`/`integration_suite_count` facts, and the regex substitution that wrote them into `PRODUCT.md` in `sync_product()`.
- `PRODUCT.md`: replaced the numeric line with a stable phrase that carries the same information without needing regeneration: `Local CI parity: smoke, unit, and integration suites covering install, commands, and provider wiring`.
- `tests/unit/test-readme-release-sync.sh`: dropped the now-unused `SMOKE_SUITE_COUNT`/`UNIT_SUITE_COUNT`/`INTEGRATION_SUITE_COUNT` fixture vars, the stale-count mutation in the fixture-building script, and the assertion that checked the regenerated count string.

Scope was verified to be exactly these three files (`grep -rn "Local CI parity\|smoke_suite_count\|test_suite_counts"`), so nothing else depends on this fact.

## Verification

- Confirmed the fix actually closes the reported mechanism: added a scratch test file, ran `sync-readme.py --check` (passes) and `git diff --stat PRODUCT.md` (empty) — adding a test file no longer touches `PRODUCT.md`.
- `python3 -c "import ast; ast.parse(...)"` on `sync-readme.py` and `bash -n` on the shell test — both clean.
- `bash tests/unit/test-readme-release-sync.sh`: 8/8 passed.
- `make test-unit`: 191/194 suites passed. The 3 failures (`test-feature-disclosure.sh`, `test-github-work-queue-hook.sh`, `test-hud-smart-mode.sh`) are pre-existing — verified identical on unmodified `main` via `git stash` before re-running them.
- No file-mode changes (`git diff origin/main...HEAD --summary` is empty).

Closes #729.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L34WZnWDW3oaABjX3Md8zF)_

## Summary by CodeRabbit

* **Documentation**
  * Updated product documentation to describe local CI coverage by test-suite category instead of exact test counts.
  * Simplified README synchronization so release, provider, capability, and model information remains updated without suite-count details.

### DIFF
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -77,7 +77,7 @@ Frontier AI models will remain individually overconfident for the foreseeable fu
 **Traction (as of 2026-07-27):**
 - GitHub stars: 3,889
 - GitHub forks: 365
-- Local CI parity: 16 smoke, 194 unit, and 7 integration suites
+- Local CI parity: smoke, unit, and integration suites covering install, commands, and provider wiring
 - Version: 9.57.0 (active release cadence)
 - Runtimes supported: Claude Code, Codex CLI, Command Code CLI, Cursor (MCP), Gemini CLI
 
--- a/scripts/sync-readme.py
+++ b/scripts/sync-readme.py
@@ -144,12 +144,6 @@ def derive_facts(root: Path) -> dict[str, object]:
     command_count = len(plugin.get("commands", []))
     skill_count = len(plugin.get("skills", []))
     persona_count = len(list((root / "agents/personas").glob("*.md")))
-    test_suite_counts = {
-        category: len(list((root / "tests" / category).glob("test-*.sh")))
-        for category in ("smoke", "unit", "integration")
-    }
-    if not all(test_suite_counts.values()):
-        raise ValueError("unable to derive local CI suite counts")
 
     orchestrate = (root / "scripts/orchestrate.sh").read_text()
     providers = (root / "scripts/lib/providers.sh").read_text()
@@ -191,9 +185,6 @@ def derive_facts(root: Path) -> dict[str, object]:
         "command_count": command_count,
         "skill_count": skill_count,
         "persona_count": persona_count,
-        "smoke_suite_count": test_suite_counts["smoke"],
-        "unit_suite_count": test_suite_counts["unit"],
-        "integration_suite_count": test_suite_counts["integration"],
         "minimum_version": minimum_version,
         "capability_count": len(capability_flags),
         "capability_ceiling": capability_ceiling,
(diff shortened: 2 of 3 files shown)


nyldn:
Closing as duplicate — #730 (opened before this one) already fixes this, and does it more robustly: its `Local CI parity: .*` normalisation matches any historical count-line format, so in-flight branches converge on the stable phrase during their next sync instead of re-conflicting with each other the way this PR's exact-pattern match would have. No need to carry two fixes for the same issue.

---

PR #737, opened 2026-08-03
fix: guard Round 1 spawn failure from aborting code review

### LINKED ISSUE #736: /octo:review aborts in Round 1 on large diffs: `spawn_agent produced no provider PID` after a 120s wait, while the prompt summarizer's own helper agent fails
### Plugin Version

9.58.0

### Claude Code Version

2.1.220

### What happened?

A single `orchestrate.sh code-review` invocation on a ~9.8k-line PR diff died at exit 1 about 2m11s into Round 1, having produced zero review results. Four things went wrong together:

1. **Round 1 never spawned a review role.** `spawn_agent_capture_pid` waited out its full attempt budget for `review-r1-logic-reviewer-1785737984` and logged `refusing to track wrapper PID 1833188`, then killed the wrapper and aborted the run.

2. **The only agents that ever ran were the summarizer's helpers, and they failed.** The prompt exceeded the context budget (`summarizing prompt for codex from 344550 to <=256000 chars`), and `agents.jsonl` contains exactly one entry — `gemini-fast`/`synthesizer`, `status: failed`, `Exit code 1`, `tokens_out: 0` after 196s. `oversize.jsonl` shows both helper prompts crushed from ~124k chars to 32k (`outcome: truncated`). So the summarization step appears to consume the whole PID-wait window and then fail, and the review roles it was preparing prompts for never start.

3. **A single invocation left 9 live `orchestrate.sh code-review` processes** with byte-identical argv (PIDs 1810711, 1810714, 1833017, 1833230, 1833404, 1833407, 1833408, 1833727, 1833730). `pkill -f 'orchestrate.sh code-review'` reaped the parents but **orphaned the provider children** — a `timeout -k 10 180 env -i … codex exec --skip-git-repo-check --model gpt-5.4-mini --sandbox workspace-write -` tree kept running and had to be killed by PID.

4. **The run never marked itself finished.** `state.json` is still `"status": "running"` with `"finished_at": null`, and `artifacts/` is empty. The next invocation then greets you with `SUCCESS: State file already exists and is valid`, silently adopting the dead run's state.

This is **not** the known `progress.json` write-race: `jq -e . ~/.claude-octopus/progress.json` validated cleanly, there were no `progress.json.tmp.*` leftovers, and `pgrep -af orchestrate` before launching showed no stale pipelines from a previous session.

Timing points at the `max_attempts` budget as the proximate trigger: `scripts/lib/spawn.sh:1157` defaults `OCTOPUS_SPAWN_PID_WAIT_ATTEMPTS` to 1200 with `sleep 0.1` per iteration = ~120s ceiling, and the ERROR landed 2m11s after start. On a large diff, the work that must complete before `spawn_agent` reaches its closing `echo "$pid"` (line 1139) — including provider-backed prompt summarization — exceeds that window, so the capture loop times out on a wrapper that is still legitimately working.

Note also that the log lines are emitted out of chronological order (the `15:19:43` WARN is flushed *after* the `15:21:54` ERROR), which makes this much harder to diagnose from the log alone.

### What did you expect?

- The PID-capture wait to be bounded by the provider's own startup, not by however long prompt summarization takes — or for the wait to scale with prompt size / be documented as tunable when a diff triggers summarization.
- A summarizer helper exiting 1 to surface as an explicit, attributed error ("prompt summarization failed for role X") and either fall back to an unsummarized/chunked prompt or fail the run cleanly, rather than manifesting as an unrelated-looking missing-PID error on a review role.
- One invocation to produce one `orchestrate.sh` process tree.
- Killing the orchestrator to reap its provider children, leaving no orphaned `codex exec`.
- A run that aborts to write `status: "failed"` and `finished_at` into `state.json`, so the next run doesn't inherit stale state.

### Steps to reproduce

1. Check out a branch with a large diff vs. `origin/main` — ~9.8k lines across ~76 files reproduced it here. The prompt must be big enough to trip the context budget (>256,000 chars; ours was 344,550).
2. Have an open PR for that branch, with `codex` and `gemini` installed and Perplexity *not* configured.
3. Run:
   ```bash
   ~/.claude-octopus/plugin/scripts/orchestrate.sh code-review \
     '{"target":<PR#>,"focus":["correctness","security","architecture","tdd"],"provenance":"unknown","autonomy":"autonomous","publish":"ask","debate":"auto","history":"auto"}'
   ```
4. In another shell, watch `pgrep -af orchestrate` — multiple identical processes appear.
5. Wait ~2m10s. The run exits 1 with the missing-PID error and no results.
6. `pkill -f 'orchestrate.sh code-review'`, then `pgrep -af codex` — the provider child is still alive.
7. `cat ~/.claude-octopus/runs/<run-id>/state.json` — still `"running"`, `finished_at: null`, `artifacts/` empty.

### Which providers are configured?

Codex CLI

### Operating System

Linux

### Relevant logs

Orchestrator output, ANSI stripped, verbatim and complete:

SUCCESS: State file already exists and is valid
[2026-08-03 15:19:43] INFO: review_run: target=335 focus=correctness,security,architecture,tdd provenance=unknown autonomy=autonomous history=auto context=none
[2026-08-03 15:19:43] INFO: review_run: Round 1 — parallel specialist fleet (no wall timeout, stall_window=1800s, diff=9810 lines)
[2026-08-03 15:19:43] INFO: CVE lookup: Perplexity unavailable, using Gemini search
[2026-08-03 15:21:54] ERROR: spawn_agent produced no provider PID for review-r1-logic-reviewer-1785737984; refusing to track wrapper PID 1833188
[2026-08-03 15:19:43] WARN: Context budget: summarizing prompt for codex from 344550 to <=256000 chars (~64000 tokens)

`runs/<run-id>/agents.jsonl` — one line total:

{"agent":"gemini-fast","role":"synthesizer","status":"failed","tokens_in":30997,"tokens_out":0,"duration_ms":196000,"reason":"Exit code 1","output_file":"","ts":"2026-08-03T06:23:03Z"}

`runs/<run-id>/oversize.jsonl`:

{"ts":"2026-08-03T06:19:47Z","agent":"gemini-fast","original_chars":123988,"final_chars":32000,"outcome":"truncated"}
{"ts":"2026-08-03T06:23:03Z","agent":"codex-mini","original_chars":124527,"final_chars":32000,"outcome":"truncated"}

`runs/<run-id>/state.json` (never finalized):

{
  "schema": 1,
  "workflow": "review",
  "goal": "target=335 focus=correctness,security,architecture,tdd",
  "status": "running",
  "started_at": "2026-08-03T06:19:44Z",
  "finished_at": null
}

Relevant source — `scripts/lib/spawn.sh:1155-1183`:

    local pid=""
    local attempts=0
    local "max_attempts=${OCTOPUS_SPAWN_PID_WAIT_ATTEMPTS:-1200}"
    ...
    while [[ $attempts -lt $max_attempts ]]; do
        pid=$(awk '/^[0-9]+$/ { value=$1 } END { print value }' "$pid_file" 2>/dev/null)
        [[ -n "$pid" ]] && break
        ...
        sleep 0.1
        ((attempts++)) || true
    done

    if [[ -z "$pid" ]]; then
        log "ERROR" "spawn_agent produced no provider PID for $task_id; refusing to track wrapper PID $wrapper_pid" >&2

Orphaned child surviving `pkill -f 'orchestrate.sh code-review'`:

timeout -k 10 180 env -i PATH=… codex exec --skip-git-repo-check --model gpt-5.4-mini --sandbox workspace-write -

nyldn:
## Description

`review_run()`'s Round 1 dispatch loop (`scripts/lib/review.sh`) assigned `spawn_agent_capture_pid`'s output directly:

```bash
round1_pid=$(spawn_agent_capture_pid "$agent_type" "$agent_prompt" "$task_id" "$role" "review")
round1_pids+=("$round1_pid")
```

`orchestrate.sh` runs under `set -eo pipefail`. When `spawn_agent_capture_pid` returns non-zero — e.g. because prompt summarization for a huge diff outran its PID-wait budget — that propagated straight through `set -e` and killed `review_run()` mid-loop. That produced every symptom in #736:

- Only the fleet member that failed ever got a chance to spawn; every later `while IFS=: read` iteration for the rest of the fleet was skipped because the function exited immediately.
- None of the downstream accounting (`round1_partial_count`, the "all providers failed" branch, `octo_proof_finalize`) ever ran, because it all lives *after* the loop that just aborted.
- `state.json` was left at `"status": "running"` / `"finished_at": null` forever, so the next invocation adopted the dead run's state (`SUCCESS: State file already exists and is valid`).

## Fix

Guard the assignment so one provider's spawn failure is treated the same as any other Round 1 agent that produced no result file — log a warning and continue dispatching the rest of the fleet:

```bash
local round1_pid=""
if ! round1_pid=$(spawn_agent_capture_pid "$agent_type" "$agent_prompt" "$task_id" "$role" "review"); then
    log WARN "review_run: spawn_agent_capture_pid failed for ${agent_type}/${role}; continuing Round 1 with remaining fleet"
    round1_pid=""
fi
round1_pids+=("$round1_pid")
```

No result file gets written for that slot, and review.sh's existing missing-result handling (`if [[ ! -f "$f" ]]`) already counts it toward `round1_partial_count` / the all-providers-failed check, which already drives `octo_proof_finalize`. So a total-fleet failure now correctly finalizes `state.json` instead of leaving it stuck at `"running"`, and a partial failure now correctly continues with whichever providers did succeed — using code paths that already existed for the "provider crashed" case, not new heuristics.

This is scoped to the two symptoms with a clean, low-risk fix: the loop aborting and `state.json` never finalizing. It does not attempt to retune the PID-wait timeout itself (root cause #2 in the issue — summarization time vs. wait budget) or add process-group-based cleanup for orphaned provider children (root cause #3) — both are more invasive changes that need validation against a real large-diff run with live providers, which isn't available in this environment. Filed as follow-up scope, not silently dropped.

## Testing

- `bash -n scripts/lib/review.sh` and all other shell scripts — syntax OK.
- New regression suite `tests/unit/test-review-round1-spawn-failure.sh`:
  - Static check that the guarded form is present and the old unguarded assignment is gone.
  - Behavioral test: extracts the real Round 1 dispatch loop verbatim via `sed` (same technique as `test-spawn-agent-capture-pid.sh`), runs it under `set -eo pipefail` with a mock `spawn_agent_capture_pid` that fails for the first fleet member and succeeds for the second — confirms both fleet members are dispatched and the harness doesn't abort.
  - Verified this test fails on the pre-fix code (`git stash` the fix, rerun) and passes after — confirms it's a real regression test, not a tautology.
- `bash tests/unit/test-review-run.sh`, `test-review-aggregation-robustness.sh`, `test-spawn-agent-capture-pid.sh` — all pass, no regressions.
- `make test-unit` (197 suites): 195/197 pass. The 2 failures (`test-feature-disclosure.sh`, `test-hud-smart-mode.sh`) are pre-existing on unmodified `main` — confirmed by stashing this change and rerunning them — and unrelated to `review.sh`/`spawn.sh`.

## Related Issues
Closes #736

---
_Generated by [Claude Code](https://claude.ai/code/session_01JmPZS9vE1BMP3cbzmihuC3)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when an agent fails to start during the first review round.
  * Remaining agents now continue launching, while failed starts are recorded and reported appropriately.

* **Tests**
  * Added regression coverage to verify failed and successful agent launches are handled correctly.

### DIFF
--- a/scripts/lib/review.sh
+++ b/scripts/lib/review.sh
@@ -982,8 +982,18 @@ CRITICAL OUTPUT FORMAT: Return ONLY a valid JSON object. No markdown, no prose, 
 ${agent_prompt_base}"
         round1_prompts+=("$agent_prompt")
 
-        local round1_pid
-        round1_pid=$(spawn_agent_capture_pid "$agent_type" "$agent_prompt" "$task_id" "$role" "review")
+        # A single provider failing PID capture (e.g. a huge diff pushes prompt
+        # summarization past the wait budget) must not abort the whole round via
+        # `set -e` — that skips every remaining fleet member and leaves
+        # state.json stuck at status "running" forever (#736). Treat it the same
+        # as any other Round 1 agent that produced no result file: the existing
+        # missing-result accounting below already counts it toward
+        # round1_partial_count / _r1_failed and drives octo_proof_finalize.
+        local round1_pid=""
+        if ! round1_pid=$(spawn_agent_capture_pid "$agent_type" "$agent_prompt" "$task_id" "$role" "review"); then
+            log WARN "review_run: spawn_agent_capture_pid failed for ${agent_type}/${role}; continuing Round 1 with remaining fleet"
+            round1_pid=""
+        fi
         round1_pids+=("$round1_pid")
     done <<< "$fleet"
 
--- a/tests/unit/test-review-round1-spawn-failure.sh
+++ b/tests/unit/test-review-round1-spawn-failure.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+# Regression for #736: on a large diff, prompt summarization for one Round 1
+# provider can outrun spawn_agent_capture_pid's PID-wait budget. Before this
+# fix, the resulting non-zero exit propagated through orchestrate.sh's
+# `set -eo pipefail`, killing review_run() mid-dispatch: no further Round 1
+# providers were spawned, and every downstream octo_proof_finalize call
+# (including the "all providers failed" path) was skipped — leaving
+# state.json stuck at status "running" forever.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+source "$SCRIPT_DIR/../helpers/test-framework.sh"
+test_suite "review_run Round 1 spawn failure resilience (#736)"
+
+REVIEW_SH="$PROJECT_ROOT/scripts/lib/review.sh"
+
+# ── static guard: the exact unguarded pattern that caused #736 must be gone ──
+
+test_case "round1 spawn call is guarded against a non-zero exit"
+if grep -q 'if ! round1_pid=\$(spawn_agent_capture_pid' "$REVIEW_SH"; then
+    test_pass
+else
+    test_fail "expected an 'if ! round1_pid=\$(spawn_agent_capture_pid ...)' guard in review.sh"
+fi
+
+test_case "no bare unguarded round1_pid assignment remains"
+# The old buggy shape: a plain assignment with no if/guard around it, which
+# aborts the whole function under `set -e` on any single provider failure.
+if grep -qE '^\s*round1_pid=\$\(spawn_agent_capture_pid' "$REVIEW_SH"; then
+    test_fail "found an unguarded 'round1_pid=\$(spawn_agent_capture_pid ...)' assignment"
+else
+    test_pass
+fi
+
+# ── behavioral: extract the real Round 1 dispatch loop and prove it survives
+# a mid-fleet spawn failure under `set -e`, exactly as orchestrate.sh runs it ──
+
+test_case "a single failed provider does not abort dispatch of the rest of the fleet"
+
+LOOP_SRC=$(sed -n '/^    fleet_dispatch_begin$/,/^    fleet_dispatch_end$/p' "$REVIEW_SH")
+if [[ -z "$LOOP_SRC" ]]; then
+    test_fail "could not extract the Round 1 dispatch loop from review.sh (source drift?)"
+else
+    TEST_TMP_DIR="/tmp/octopus-tests-$$"
+    mkdir -p "$TEST_TMP_DIR"
+    trap 'rm -rf "$TEST_TMP_DIR"' EXIT
+
+    HARNESS="$TEST_TMP_DIR/harness.sh"
+    {
+        echo '#!/usr/bin/env bash'
+        echo 'set -eo pipefail'  # mirrors orchestrate.sh exactly
+        echo 'RESULTS_DIR="'"$TEST_TMP_DIR"'"'
+        echo 'timestamp="ts1"'
+        echo 'fleet="codex:reviewer1:logic
+gemini:reviewer2:style"'
+        echo 'round1_files=(); round1_agent_types=(); round1_roles=(); round1_task_ids=(); round1_prompts=(); round1_pids=()'
+        echo 'agent_prompt_base="base prompt"'
+        echo 'log() { :; }'
+        echo 'fleet_dispatch_begin() { :; }'
+        echo 'fleet_dispatch_end() { :; }'
+        # First fleet member fails PID capture (the #736 scenario); second succeeds.
+        echo 'spawn_agent_capture_pid() {'
+        echo '    if [[ "$1" == "codex" ]]; then'
+        echo '        echo "spawn_agent produced no provider PID" >&2'
+        echo '        return 1'
+        echo '    fi'
+        echo '    echo 424242'
+        echo '}'
+        echo 'run_round1() {'
+        printf '%s\n' "$LOOP_SRC"
+        echo '}'
+        echo 'run_round1'
+        echo 'printf "count=%d\n" "${#round1_pids[@]}"'
+        echo 'printf "pid0=%s\n" "${round1_pids[0]:-}"'
+        echo 'printf "pid1=%s\n" "${round1_pids[1]:-}"'
+    } > "$HARNESS"
+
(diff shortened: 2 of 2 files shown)


nyldn:
Verified locally, 3/3, and mutation-checked: restoring the original unguarded `round1_pid=$(spawn_agent_capture_pid ...)` shape fails all three cases, including the behavioural one that runs the extracted dispatch loop under `set -e`. Merging.

**One thing to be careful about: this should not close #736.** The report names four distinct failures and this fixes two of them.

Fixed here:
- **(1) the abort.** One provider failing PID capture no longer takes down the round via `set -e`.
- **(4) stale `state.json`.** As the code comment notes, the missing-result accounting now actually runs, so `octo_proof_finalize` fires and the run records `failed` instead of leaving `status: "running"` for the next invocation to silently adopt.

Still open after this merge:
- **(2) the root cause.** `OCTOPUS_SPAWN_PID_WAIT_ATTEMPTS` defaults to 1200 × `sleep 0.1` ≈ a 120s ceiling (`spawn.sh:1157`), and on a large diff provider-backed prompt summarization consumes that window before `spawn_agent` reaches its closing `echo "$pid"`. The reporter's ERROR landed at 2m11s, which matches. This change makes that degrade gracefully rather than fatally — the review now proceeds with the remaining fleet — but the seat is still lost on every large diff. The wait should be bounded by provider startup, not by summarization.
- **(3) the process leak.** One invocation left nine `orchestrate.sh code-review` processes with byte-identical argv, and `pkill` on the parents orphaned the `codex exec` children. Nothing here touches that, and it is the most user-hostile part of the report — an orphaned provider keeps burning tokens after the user thinks they killed the run.

Also worth noting from the report and not addressed: log lines are emitted out of chronological order (a 15:19:43 WARN flushed after a 15:21:54 ERROR), which is why this took a detailed reporter to diagnose at all.

Suggest keeping #736 open and scoped to (2) and (3) after this lands, or splitting them into their own issues.

Issue #336, opened 2026-05-01
[BUG] build_provider_env() silently kills background Codex subshell when OPENAI_API_KEY is absent (set -e + unguarded && pattern)


SouroucheB:
## Plugin Version

Current (fork SouroucheB/claude-octopus, based on upstream main)

## Operating System

macOS 25.3.0 (Darwin)

## Providers Configured

- Codex CLI (OAuth via keychain — OPENAI_API_KEY not exported as env var)
- Gemini CLI (OAuth)

## What happened?

The `develop` (and likely any tangle/multi-agent) workflow silently produces empty Codex result files (~421 bytes, header only). Codex never runs. No error is shown. Claude falls back to solo execution.

## Root Cause

Same `set -e` + unguarded `[[ ... ]] && cmd` pattern as #54 and #241, but in a different location: **`build_provider_env()` in `scripts/lib/provider-routing.sh`**.

```bash
# provider-routing.sh ~line 52
codex*)
    [[ -z "${OPENAI_API_KEY:-}" ]] && resolve_provider_env "OPENAI_API_KEY" 2>/dev/null
    ...
```

When `OPENAI_API_KEY` is not exported as a shell environment variable (common when using OAuth / keychain-based auth), `resolve_provider_env` searches `~/.profile` and `.env`, finds nothing, and **returns 1**.

Because this is the last command of the `&&` chain, bash propagates exit code 1 as the statement result. The background subshell spawned in `spawn_agent()` inherits `set -eo pipefail` from `orchestrate.sh:6`, sees exit code 1, and **terminates silently** — right after writing the file header (lines 460–468 of spawn.sh), before Codex is ever invoked.

## Reproduction

```bash
bash -c '
  set -eo pipefail
  source ~/.claude-octopus/plugin/scripts/orchestrate.sh 2>/dev/null
  unset OPENAI_API_KEY
  build_provider_env "codex"
  echo "never reached"
'
# Exit code: 1 — script dies at build_provider_env
```

## Why it was hard to diagnose

- The background subshell exits **silently** (no log, no stderr)
- The result file **exists** with a valid-looking header — nothing alerts the orchestrator
- All isolated tests pass (the bug only manifests in a `( ) &` background subshell with `set -e` inherited **and** `OPENAI_API_KEY` absent from the env)
- Codex itself works fine via direct invocation

## Fix

[...]

PR #739, opened 2026-08-03
fix: repair OpenRouter council seat (probe, wait, model resolution)

### LINKED ISSUE #738: Council OpenRouter seat never works: availability probe wants a nonexistent binary, run_with_timeout fallback kills function seats under set -e, configured model ignored (v9.58.0)
**Plugin Version:** 9.58.0
**Claude Code Version:** 2.1.220
**Environment:** Linux (WSL2), bash 5.x, `OPENROUTER_API_KEY` set and valid (verified against the OpenRouter API directly)
**Providers configured:** claude (host), codex (ChatGPT OAuth), openrouter

### What happened

Council runs with a valid OpenRouter key never produce an OpenRouter seat response. Digging in with `bash -x` and the preserved `.tmp-agent-*` files, we found three independent causes in the council → OpenRouter path. Each is small; together they mean the OpenRouter council seat cannot work out of the box.

#### 1. Council availability probe requires an `openrouter` binary that doesn't exist

`council_detect_providers` (scripts/lib/council.sh) checks providers with `command -v "$cmd"`, and `council_provider_command openrouter` returns `openrouter`. The plugin ships no `openrouter` executable — dispatch uses the shell function `openrouter_execute` — so provider status is always `missing` and the seat is never rostered, regardless of `OPENROUTER_API_KEY`.

Repro: `orchestrate.sh council --dry-run ...` with a valid key → `summary.json` shows `"openrouter": "missing"`. Creating any executable named `openrouter` on PATH flips it to `available`.

Suggested fix: for API-based providers, check the same condition the dispatch layer uses (key set + enabled) instead of `command -v`.

#### 2. `run_with_timeout` fallback kills function-dispatched seats under `set -e` after the command completes

scripts/lib/heartbeat.sh, in-process fallback path (used for shell-function providers like `openrouter_execute` / `perplexity_execute`):

```bash
kill "$monitor_pid" 2>/dev/null
wait "$monitor_pid" 2>/dev/null     # <- monitor was just SIGTERMed, wait returns 143
```

Under `set -eo pipefail` (orchestrate.sh line 6) the non-zero `wait` terminates the seat subprocess *after* the provider call succeeded but *before* the response is delivered — the seat's captured output dies with it and the council records `no-response`. An xtrace of the seat ends exactly at that `wait`. Same class as the `set -e` subshell kill you fixed in #336.

Binary providers (codex etc.) are unaffected because they take the external `timeout` path.

Fix: `wait "$monitor_pid" 2>/dev/null || true`

#### 3. `openrouter_execute` ignores the configured model — always uses the hardcoded task-type table

scripts/lib/perplexity.sh `openrouter_execute` resolves its model via `get_openrouter_model` (hardcoded table → `anthropic/claude-sonnet-4` for council prompts) and never consults `OCTOPUS_OPENROUTER_MODEL` or `providers.json` (`resolve_octopus_model`). Meanwhile the council roster/config.json reports the *configured* model — in our run the roster said `google/gemini-3.1-pro-preview` while the seat's stderr showed `OpenRouter request: model=anthropic/claude-sonnet-4`. This is a live instance of the "roster model labels don't match what ran" class from #599, and it silently defeats cross-lab diversity (the OpenRouter seat duplicates the Anthropic chair's lab).

Suggested fix in `openrouter_execute`:

```bash
model="${OCTOPUS_OPENROUTER_MODEL:-}"
if [[ -z "$model" ]] && declare -f resolve_octopus_model >/dev/null 2>&1; then
    model="$(resolve_octopus_model openrouter openrouter "" "" 2>/dev/null || true)"
fi
[[ -z "$model" ]] && model=$(get_openrouter_model "$task_type" "$complexity")
```

### One additional observation (not fully diagnosed)

With fixes 1–2 applied, the default detached seat dispatch (`OCTOPUS_COUNCIL_DETACH=1`) still lost the OpenRouter response in our environment (seat's temp output complete, but no `.partial`→response rename; `.done` never appeared within the reap window). `OCTOPUS_COUNCIL_DETACH=0` (inline) delivers the seat reliably. Happy to provide traces if useful.

### What did you expect

A council run with a valid `OPENROUTER_API_KEY` to seat OpenRouter and use the model configured in `providers.json`.

### Steps to reproduce

1. `export OPENROUTER_API_KEY=...` (valid), no `openrouter` binary on PATH.
2. `orchestrate.sh council --goal advice --depth quick --providers claude,codex,openrouter "any question"` → openrouter reported missing (bug 1).
3. Put any executable named `openrouter` on PATH, rerun → seat rostered, dispatch starts, seat dies at the `wait "$monitor_pid"` line (bug 2, visible under `bash -x`).
4. Add `|| true`, rerun with a custom model in `providers.json` → seat responds, but from `anthropic/claude-sonnet-4` instead of the configured model (bug 3, visible in the seat's stderr debug line).

---

Found while setting up the plugin, debugged with Claude Code assistance; every repro above was run and verified locally, and with the three fixes applied our councils (claude + codex + openrouter/gemini) complete with all seats responding.

nyldn:
## Description

`council` with a valid `OPENROUTER_API_KEY` never produced an OpenRouter seat. Three independent bugs in the council → OpenRouter path, all reported and pre-verified by the issue author:

1. **`council_detect_providers()`** (`scripts/lib/council.sh:2387-2404`) checked `command -v openrouter`, but the plugin ships no `openrouter` binary — dispatch goes through the shell function `openrouter_execute`. The seat was always reported `missing` regardless of the key. Fixed by adding an `openrouter)` case that probes `OPENROUTER_API_KEY` instead, mirroring the existing `openai-compatible` pattern.
2. **`run_with_timeout()`**'s in-process fallback (`scripts/lib/heartbeat.sh:212-214`, used for shell-function providers like `openrouter_execute`) ran `wait "$monitor_pid"` unguarded right after SIGTERMing that monitor. Under `set -eo pipefail` (`orchestrate.sh` line 6) the monitor's 143 exit code terminated the function — and the seat's already-captured output with it — *after* the provider call had succeeded. Same class of bug as the subshell kill fixed in #336. Fixed with `|| true`.
3. **`openrouter_execute()`** (`scripts/lib/perplexity.sh:149-164`) always resolved its model from the hardcoded task-type table via `get_openrouter_model`, ignoring `OCTOPUS_OPENROUTER_MODEL` / `providers.json`. The council roster advertised the configured model while the seat silently ran a different one. Fixed to prefer the env override, then `resolve_octopus_model`, falling back to the task-type table only if both are empty.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests pass: directly affected suites (see Testing below)
- [ ] OpenClaw registry in sync — N/A, no tool/command surface changed
- [ ] New skills/commands registered — N/A
- [ ] Version bump — N/A, this is a fix PR; version bump happens in the release PR
- [ ] Documentation updated — N/A
- [ ] CHANGELOG.md updated — left for the release PR, per repo convention (see recent `fix:` commits on `main`, e.g. #722, #726, #721)

## Testing

```bash
bash -n scripts/lib/council.sh scripts/lib/heartbeat.sh scripts/lib/perplexity.sh
bash -n scripts/*.sh scripts/lib/*.sh

bash tests/unit/test-council-command.sh          # 67 passed
bash tests/unit/test-heartbeat-timeout.sh        # 12 passed
bash tests/unit/test-council-model-selection-fixes.sh  # 10 passed
bash tests/unit/test-qwen-council-provider.sh    # 6 passed
bash tests/unit/test-perplexity-issue-307.sh     # 6 passed
bash tests/unit/test-provider-health-probe.sh    # 12 passed
```

Also manually re-verified fix 1 by sourcing `council.sh` and calling `council_detect_providers` with `COUNCIL_PROVIDERS=openrouter`, toggling `OPENROUTER_API_KEY` — reports `available`/`missing` correctly.

`make test-unit` (full suite) was also run locally.

## Related Issues
Closes #738

---
_Generated by [Claude Code](https://claude.ai/code/session_01At8qoUSS8omaktiiou7wSA)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OpenRouter availability detection when configured with an API key.
  - Fixed timeout handling to prevent interrupted monitoring from prematurely terminating operations.
  - Improved model selection by honoring explicit configuration and resolved model preferences before applying fallbacks.

### DIFF
--- a/scripts/lib/council.sh
+++ b/scripts/lib/council.sh
@@ -2392,6 +2392,16 @@ council_detect_providers() {
                         status="missing"
                     fi
                     ;;
+                openrouter)
+                    # API-key provider, not a CLI binary — no `openrouter` executable
+                    # ships with the plugin. Dispatch goes through the shell function
+                    # openrouter_execute, so probe the key instead of `command -v` (#738).
+                    if [[ -n "${OPENROUTER_API_KEY:-}" ]]; then
+                        status="available"
+                    else
+                        status="missing"
+                    fi
+                    ;;
                 *)
                     cmd="$(council_provider_command "$provider")"
                     if command -v "$cmd" >/dev/null 2>&1; then
--- a/scripts/lib/heartbeat.sh
+++ b/scripts/lib/heartbeat.sh
@@ -209,8 +209,11 @@ run_with_timeout() {
         fi
 
         # Stop the monitor and sweep any stragglers parented to the command.
+        # `wait` on a just-SIGTERMed monitor returns 143 — under `set -e` that would
+        # kill this function (and the seat's already-captured output with it) after
+        # the provider call succeeded. Same class as the subshell kill fixed in #336. (#738)
         kill "$monitor_pid" 2>/dev/null
-        wait "$monitor_pid" 2>/dev/null
+        wait "$monitor_pid" 2>/dev/null || true
         pkill -KILL -P "$cmd_pid" 2>/dev/null || true
     fi
 
--- a/scripts/lib/heartbeat.sh
+++ b/scripts/lib/heartbeat.sh
@@ -209,10 +209,11 @@ run_with_timeout() {
         fi
 
         # Stop the monitor and sweep any stragglers parented to the command.
-        # `wait` on a just-SIGTERMed monitor returns 143 — under `set -e` that would
-        # kill this function (and the seat's already-captured output with it) after
-        # the provider call succeeded. Same class as the subshell kill fixed in #336. (#738)
-        kill "$monitor_pid" 2>/dev/null
+        # `kill`/`wait` on a monitor that already exited on its own (race with its
+        # sleep) return non-zero — under `set -e` that would kill this function (and
+        # the seat's already-captured output with it) after the provider call
+        # succeeded. Same class as the subshell kill fixed in #336. (#738)
+        kill "$monitor_pid" 2>/dev/null || true
         wait "$monitor_pid" 2>/dev/null || true
         pkill -KILL -P "$cmd_pid" 2>/dev/null || true
     fi
--- a/scripts/lib/perplexity.sh
+++ b/scripts/lib/perplexity.sh
@@ -157,8 +157,14 @@ openrouter_execute() {
     local complexity="${3:-2}"
     local output_file="${4:-}"
 
-    local model
-    model=$(get_openrouter_model "$task_type" "$complexity")
+    # Prefer the configured model (OCTOPUS_OPENROUTER_MODEL / providers.json) over
+    # the hardcoded task-type table, so the roster's advertised model actually runs
+    # instead of silently duplicating the Anthropic chair's lab (#738).
+    local model="${OCTOPUS_OPENROUTER_MODEL:-}"
+    if [[ -z "$model" ]] && declare -f resolve_octopus_model >/dev/null 2>&1; then
+        model="$(resolve_octopus_model openrouter openrouter "" "" 2>/dev/null || true)"
+    fi
+    [[ -z "$model" ]] && model=$(get_openrouter_model "$task_type" "$complexity")
 
     openrouter_execute_model "$model" "$prompt" "$task_type" "$complexity" "$output_file"
 }


nyldn:
Reviewed all three fixes; all correct. Merging, with one gap noted below.

**The `heartbeat.sh` fix is the important one, and I verified the failure is real rather than theoretical.** It took three attempts:

```
first wait rc=0        # reaps the monitor
kill on reaped rc=1    # <- non-zero
second wait rc=0
```

and under `set -e` in plain statement context that aborts:

```
outer rc=1             # REACHED-END never printed
```

My first two reproductions wrongly showed it surviving, because I wrapped the call as `out=$(seat) && … || …` — putting the function in a tested compound context, which suppresses `set -e` inside it. Worth knowing for anyone testing this class of bug: the harness can hide it. So the comment's framing is exactly right, and this is the same class as the subshell kill in #336.

The other two are straightforward and right: `openrouter` is an API-key provider with no binary, so `command -v` could never have found it; and preferring the configured model over the hardcoded task-type table stops the roster advertising one lab while silently running the chair's.

**Gap: no tests.** `run_with_timeout` has 11 existing unit suites and `openrouter_execute` has 1, but none covers these three behaviours — so nothing stops any of them regressing. Given how much work it took to reproduce the `set -e` leak above, that one in particular deserves a regression test; the reproduction in this comment can be lifted almost verbatim. Merging anyway because two external users are broken right now and the fixes are clearly correct, but I'd like a follow-up adding:

1. `kill`/`wait` on an already-reaped monitor does not abort `run_with_timeout` and the captured output survives;
2. `council_detect_providers` reports `openrouter` available on `OPENROUTER_API_KEY` alone, with no binary on PATH;
3. `openrouter_execute` honours `OCTOPUS_OPENROUTER_MODEL` over the task-type table.

I can write those if you'd rather not.

nyldn:
Added the follow-up regression coverage in #742 — all three behaviors plus the fallback ordering and the negative case, verified against both sides of this fix (fails on pre-#739 code, passes on the merged fix).


---

PR #740, opened 2026-08-03
fix: preserve valid explicit new write scopes

Jhacarreiro:
## Summary

- preserve explicit new file and directory scopes when they are safely anchored under an existing repository path
- avoid replacing valid declared scopes with heuristic context-file inference
- reject unsafe absolute, traversal, glob, `.git`, and unanchored invented paths
- retain parent/child overlap detection for newly declared scopes

## Problem

Tangle decomposition can legitimately assign work to files or directories that do not exist yet, for example:

```text
Files: apps/server/src/development-scope/
```

When the declared path was new, the scope resolver could discard it and infer an unrelated existing file from the task prose, such as `README.md`. That created false overlap failures even though the planner had produced disjoint explicit scopes.

Observed failure mode:

```text
apps/server/src/domain/
→ heuristic context inference
→ README.md
→ false overlap with the documentation task
```

## Behaviour after this change

An explicit new scope is preserved when it:

- is repository-relative
- has no traversal, globbing, backslash, or absolute prefix
- does not target `.git`
- is anchored below an existing tracked or on-disk repository path

Examples:

```text
apps/server/src/development-scope/            accepted
apps/server/src/development-scope/handler.js accepted
totally-invented-tree/deep/file.js            rejected
apps/server/../../outside.js                  rejected
/tmp/outside.js                               rejected
apps/server/src/**                            rejected
```

New parent/child scopes are still treated as overlapping.

## Validation

Focused regression suites:

- `tests/unit/test-tangle-explicit-new-scopes.sh` — 9/9 passed
- `tests/unit/test-tangle-write-scope-safety.sh` — 12/12 passed

The focused tests include an invented development scope under an existing repository tree and verify that it is not rewritten to `README.md`.

Full unit suite:

- `make test-unit` — 197/197 suites passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for relative write paths, including normalization and stricter safety checks.
  * Added support for explicitly creating new files and directories beneath existing repository paths.
  * Rejected unsafe paths, including absolute paths, traversal attempts, unsupported globs, backslashes, and `.git` paths.
  * Improved detection of overlapping and unrelated scopes during parallel operations.

* **Tests**
  * Added coverage for tracked paths, valid repository-anchored paths, invalid paths, and scope overlap scenarios.

### DIFF
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -979,10 +979,46 @@ ${resolved:-<none resolved>}
 EOF
 }
 
+tangle_scope_is_safe_relative_path() {
+    local scope="$1"
+    local normalized="${scope#./}"
+    normalized="${normalized%/}"
+
+    [[ -n "$normalized" ]] || return 1
+    case "$normalized" in
+        /*|.|..|.git|.git/*|*\\*|*'*'*|*'?'*|*'['*|*']'*) return 1 ;;
+    esac
+    case "/$normalized/" in
+        */../*|*/./*) return 1 ;;
+    esac
+    return 0
+}
+
+tangle_scope_has_existing_repo_ancestor() {
+    local repo_root="$1"
+    local normalized="$2"
+    local probe="$normalized"
+
+    while [[ "$probe" == */* ]]; do
+        probe="${probe%/*}"
+        [[ -n "$probe" ]] || break
+        if [[ -d "$repo_root/$probe" || -f "$repo_root/$probe" ]]; then
+            return 0
+        fi
+        if git -C "$repo_root" rev-parse --show-toplevel >/dev/null 2>&1; then
+            if [[ -n "$(git -C "$repo_root" ls-files "$probe/" 2>/dev/null || true)" ]]; then
+                return 0
+            fi
+        fi
+    done
+    return 1
+}
+
 tangle_scope_is_known_or_explicit_new_file() {
     local scope="$1"
-    local normalized="${scope%/}"
-    [[ -z "$normalized" ]] && return 1
+    local normalized="${scope#./}"
+    normalized="${normalized%/}"
+    tangle_scope_is_safe_relative_path "$scope" || return 1
 
     local repo_root
     if ! repo_root=$(tangle_resolve_repo_root 2>/dev/null); then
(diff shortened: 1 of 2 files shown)


nyldn:
The direction is right and most of the guard is solid — I tested `tangle_scope_is_safe_relative_path` directly and it correctly rejects absolute paths, `.`, `..`, embedded `../` and `./` components, glob metacharacters, and backslashes. Verified:

```
reject /etc/passwd      reject a/../../etc/x     reject src/*.ts
reject ..               reject src/./x           reject sr[c]/x
reject .git             reject .git/config       reject a\\b
ALLOW  src/new.ts       ALLOW  ./src/new.ts
```

**But there is a bypass of the guard's own stated purpose, and it lands on this project's primary platform.**

```
ALLOW  .GIT/config
ALLOW  .Git/hooks/pre-commit
```

The `case` arm matches `.git` and `.git/*` literally, but macOS APFS/HFS+ is case-insensitive by default. Demonstrated in a scratch repo on this machine:

```
$ echo hi > .git/CASEPROBE && ls .GIT/CASEPROBE
  YES — .GIT/ resolves to .git/ on this filesystem

$ mkdir -p .GIT/hooks && echo 'echo pwned' > .GIT/hooks/probe && ls -la .git/hooks/probe
  -rw-r--r--  1 chris  wheel  11  .git/hooks/probe
```

So a scope of `.Git/hooks/pre-commit` passes validation and writes an executable git hook — arbitrary code execution on the user's next commit. Since `CLAUDE.md` names macOS as the project's primary platform, this is the default case, not an edge one.

Fix is small: lowercase before the comparison, e.g.

```bash
local lower
lower=$(printf '%s' "$normalized" | tr '[:upper:]' '[:lower:]')
case "$lower" in
    .git|.git/*) return 1 ;;
esac
```

Worth guarding the same way anywhere else a path is compared against `.git` — the existing destructive-delete guards had exactly this class of gap fixed for uppercase `-R` in 9.56.1.

**Second, minor:** a whitespace-only scope is allowed (`ALLOW` on `"  "`) because `normalized` is non-empty and no arm matches. Rejecting scopes that are empty after trimming would close it.

**Also worth a test:** `tests/unit/test-tangle-explicit-new-scopes.sh` is a good addition, but it should include the case-variant cases above — a guard whose test suite never tries `.GIT` will pass while the hole is open, which is what happened here.

Not merging yet — happy to push both fixes to this branch if you'd like, or leave it to you. Good catch on the underlying problem; the explicit-new-file allowance is a real gap and the ancestor check is a sensible way to bound it.

nyldn:
Pushed the case-insensitivity fix to this branch rather than blocking on it, since the underlying feature is sound and the gap was four lines.

**What was still open.** The guard matched `.git` literally, so on macOS APFS/HFS+ — case-insensitive by default, and this project's primary platform — these passed:

```
ALLOW  .GIT/config
ALLOW  .Git/hooks/pre-commit
```

`.Git/hooks/pre-commit` resolves to the real hook, which is arbitrary code execution on the next commit. Demonstrated in a scratch repo:

```
$ mkdir -p .GIT/hooks && echo 'echo pwned' > .GIT/hooks/probe
$ ls -la .git/hooks/probe
-rw-r--r--  1 chris  wheel  11  .git/hooks/probe
```

**Fix:** lowercase a copy for the `.git` comparison only (the other arms are case-irrelevant), plus rejecting whitespace-only scopes, which were non-empty but name nothing.

**The more useful part is the test gap.** `test-tangle-explicit-new-scopes.sh` had no case variant at all — that is *why* the hole was invisible. A guard whose suite never tries `.GIT` passes while it is open. Added `.GIT/config`, `.Git/hooks/pre-commit`, `.gIt/HOOKS/pre-commit` and the whitespace case; restoring the literal match fails exactly those three, so the coverage is load-bearing rather than decorative.

Everything else verified as-is and left alone: the guard correctly rejects absolute paths, `..`, embedded `../` and `./` components, glob metacharacters and backslashes, and the ancestor check is a sensible way to bound the explicit-new-file allowance. Good fix to the underlying problem — the scope-preservation logic is the hard part and it is right.

nyldn:
Merged via #764 — thanks for this. Your commits are on `main`; I verified the content is present rather than inferring it from the merge.

Closing manually because GitHub does not auto-close source PRs when an integration branch is squash-merged. Not deleting the branch, since it is yours.

**Why it went through a batch rather than merging directly:** branch protection requires every PR to be up to date with `main` at merge time, and with a dozen PRs open each merge invalidated all the others at a full CI cycle each. A merge queue is the proper fix but is unavailable here — this is a user-owned repository and merge queue is an organization-only feature. Batching nine PRs into one CI run got the same result without weakening protection.

All nine merged with **zero conflicts**, yours included.

Issue #724, opened 2026-08-02
Phase handoffs pass free-text synthesis, not structured context slots


nyldn:
Filed so it is not silently dropped when #699 closes. The #699 author named this as the higher-value half of that issue:

> "If only one thing came out of this issue, wiring those up [the `state.json` context slots] would have more evidence behind it than adding the skill."

#723 ships the diagnostic from #699 and does **not** address this. Verified the underlying claim rather than taking it on trust:

`scripts/lib/research.sh:49` hands the phase-1 result to phase 2 as interpolated prose:

```bash
personas=$(run_agent_sync "gemini" "Based on this research synthesis:
$synthesis

Create:
1. 2-3 distinct user personas ...
```

Whatever the receiving agent needs and the sending agent held but did not write into `$synthesis` is gone at that boundary, with nothing recording that it was lost. `scripts/lib/state.sh` has no context-slot concept, so there is no structured place for a phase to deposit findings that a later phase can read by key rather than by re-reading a summary.

**Why it matters more than the skill.** `skill-agent-topology` (shipped in #723) can now *name* compression loss and semantic drift at a boundary. It cannot fix them. This is the fix: give phases typed slots to hand off through, so the diagnostic has something to recommend beyond "collapse this boundary".

**Scope note.** This is a change to the phase-handoff contract, not a documentation fix. It touches `state.sh`, the phase runners in `research.sh` and `workflows.sh`, and every prompt that currently interpolates a prior phase's free text. Worth designing before implementing — in particular, what happens to a slot a phase declares but never fills, and whether slot contents are additive across a run or overwritten per phase.

Related: #699 (closed by #723), and `skills/blocks/frontier-model-routing.md` for the existing norms on when a boundary is earned at all.

nyldn:
Rescoping this. I filed it and three of its claims were wrong — recording that rather than quietly building on them.

**Corrections:**

1. It cites `scripts/lib/state.sh`. **That file does not exist.** Session state is `scripts/lib/session.sh`.
2. It says phase handoffs pass free-text synthesis. **Embrace does not.** It hands off *file paths* through a keyed registry that already exists: `save_session_checkpoint` writes `.phases[<name>] = {status, output, timestamp}` and `workflows.sh:3605/3645/3711` read it back, with an `ls -t` glob only as fallback. The structured slot mechanism this issue asked for is already there for the main workflow.
[...]

PR #743, opened 2026-08-03
fix: read phase-output checkpoints by key instead of splicing the name into jq

nyldn:
Groundwork for #724, plus a real bug found on the way in.

## The bug

`get_phase_output` used `jq -r ".phases.$phase.output"`. jq parses a hyphenated name as subtraction, so:

```
$ jq -r '.phases.debate-probe.output' session.json
jq: error: probe/0 is not defined
```

`workflows.sh:3300` writes exactly that shape — `save_session_checkpoint "debate-${gate_slug}"` — so every debate-gate checkpoint was write-only. Fixed by binding with `--arg`, which the *writer* already did.

**Impact, stated honestly: latent, not a live break.** The three real callers (`workflows.sh:3605/3645/3711`) pass `probe`, `grasp`, `tangle` — plain names that worked. Nothing currently reads a `debate-*` checkpoint, so the effect so far was unreadable data rather than a failing feature. It would have bitten the first consumer of a debate gate on resume, and the splice was an injection vector for any phase name arriving from workflow code or a gate slug.

## What the investigation found, which differs from #724

I filed #724 and three of its claims were wrong. Recording that rather than quietly building on it:

- It cites `scripts/lib/state.sh`. **That file does not exist.** Session state is `scripts/lib/session.sh`.
- It says phase handoffs pass free-text synthesis. **Embrace does not** — it hands off *file paths*, preferring this checkpoint registry and falling back to an `ls -t` glob. The structured-slot mechanism #724 asks for already exists there, keyed by phase, storing `{status, output, timestamp}`.
- The genuine free-text interpolation is in `research.sh` and `sentinel.sh`, which splice a prior step's content into the next prompt inside their own multi-step flows. That is the remaining work and it is **narrower than filed**.

So #724 should be rescoped to those two libraries rather than a new subsystem. Building a parallel slot mechanism when a keyed registry already exists is exactly what `skill-agent-topology` would flag.

## Tests

7 cases: plain, hyphenated, unrecorded, and jq-syntax phase names, plus static guards that the reader binds via `--arg` and never splices. Mutation-checked — restoring the original filter fails 3.

One harness note worth recording, since it cost me several iterations: `tests/helpers/test-framework.sh:6` sets `set -euo pipefail` and is *sourced*, so `-e` is active in every test regardless of what the test sets before sourcing. A deliberately-failing probe therefore aborts the suite mid-run unless the helper returns 0 explicitly. That is a footgun for anyone writing a test that asserts a failure path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of phase names containing hyphens or special characters.
  * Prevented malformed phase names from being interpreted as query syntax when retrieving saved session data.

* **Tests**
  * Added regression coverage for standard, hyphenated, missing, and special-character phase names.
  * Added verification that phase data can be safely saved and retrieved.

### DIFF
--- a/scripts/lib/session.sh
+++ b/scripts/lib/session.sh
@@ -506,7 +506,13 @@ get_resume_phase() {
 get_phase_output() {
     local phase="$1"
     if [[ -f "$SESSION_FILE" ]] && command -v jq &> /dev/null; then
-        jq -r ".phases.$phase.output // \"\"" "$SESSION_FILE"
+        # Bind the phase as data, not as filter text. The previous form,
+        # `jq -r ".phases.$phase.output"`, parsed a hyphenated name as
+        # subtraction — `debate-probe` failed with `probe/0 is not defined` — so
+        # every checkpoint written by workflows.sh's `save_session_checkpoint
+        # "debate-${gate_slug}"` was unreadable. Splicing a phase name into a jq
+        # program is also an injection vector; the writer already used --arg.
+        jq -r --arg phase "$phase" '.phases[$phase].output // ""' "$SESSION_FILE"
     fi
 }
 
--- a/tests/unit/test-phase-output-registry.sh
+++ b/tests/unit/test-phase-output-registry.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+# Phase-output registry round-trip (issue #724).
+#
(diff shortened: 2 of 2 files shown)


nyldn:

Issue #751, opened 2026-08-03
Lint for set -e status leaks on bare kill/wait/pkill — 5 sites still live after #739 fixed one


nyldn:
#739 fixed one instance of this and left at least four. That ratio is the argument for a lint rather than case-by-case patching.

## The failure mode

Under `set -e` (`orchestrate.sh:6` is `set -eo pipefail`), `kill` on an already-reaped PID returns 1 and aborts the enclosing function — **after** the work succeeded. Measured:

```
$ ( sleep 0.05 ) & m=$!
$ wait "$m"; echo $?          # 0, reaps it
$ kill "$m" 2>/dev/null; echo $?  # 1
$ bash -c 'set -e; ( sleep 0.05 ) & m=$!; wait "$m"; kill "$m" 2>/dev/null; echo REACHED-END'
# REACHED-END never prints, exit 1
```

In #738 this discarded a council seat's already-captured output. Fixed in #739 by adding `|| true`.

**A warning for anyone testing this class:** it is easy to write a reproduction that wrongly passes. My first two attempts used `out=$(seat) && echo ok || echo fail` — putting the function in a tested compound context, which suppresses `set -e` inside it. The bug only shows in plain statement context.

## Still live

```
scripts/lib/workflows.sh:692   kill $synthesis_monitor_pid 2>/dev/null
scripts/lib/workflows.sh:693   wait $synthesis_monitor_pid 2>/dev/null
scripts/lib/cursor-agent.sh:55 wait $cmd_pid 2>/dev/null
scripts/lib/heartbeat.sh:197   kill -TERM $cmd_pid 2>/dev/null
scripts/lib/heartbeat.sh:200   kill -KILL $cmd_pid 2>/dev/null
```

`workflows.sh:692` is the identical pattern to the one #739 fixed, in a different file, sitting immediately after a "Synthesis succeeded" log line. Severity is lower than #738's — `probe_discover`'s output is not captured, so an abort skips `display_progress_summary` rather than discarding results — but it is the same defect and it runs under `set -eo pipefail`.

`heartbeat.sh:197/200` are in the timeout path: if the command exits between the timeout firing and the `kill`, the same abort applies.

## Proposal

A portability-lint rule (the job already exists) flagging `kill`, `wait` or `pkill` as a bare statement in `scripts/lib/**` and `scripts/helpers/**` without `|| true`, an `if` guard, or a trailing `|| :`.

[...]

PR #753, opened 2026-08-03
fix: guard remaining bare kill/wait sites against set -e aborts

nyldn:
## Description

#739 fixed one instance of a recurring bug class: under `set -eo pipefail` (`orchestrate.sh:6`, inherited by every sourced lib), `kill`/`wait` on a PID that has already exited returns non-zero, which aborts the enclosing function immediately — even though the real work already succeeded.

#751 found five more live sites in this shape. This PR fixes those five.

- `scripts/lib/workflows.sh:692-693` — the progressive-synthesis monitor's `kill`/`wait` could abort right after the "Synthesis succeeded" log line, skipping `display_progress_summary`.
- `scripts/lib/heartbeat.sh:197,200` — the timeout-escalation subshell's `kill -TERM`/`kill -KILL` on `$cmd_pid` could abort mid-sweep if the command already exited, skipping the `pkill` cleanup and, for the TERM case, the later SIGKILL escalation.
- `scripts/lib/cursor-agent.sh:55` — a different shape, worth calling out: `wait "$cmd_pid"`'s own return value is consumed via `$?` on the next line to build `exit_code`. A blind `|| true` here would have silently zeroed out every non-zero exit code from the wrapped command — a worse bug than the one being fixed. Used the `if/else` idiom already established at `heartbeat.sh:205-209` instead, which avoids the `set -e` abort without discarding the real exit code.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Shell scripts pass `bash -n` syntax check (ran across all of `scripts/*.sh scripts/lib/*.sh`)
- [x] OpenClaw registry in sync: `scripts/build-openclaw.sh --check` (covered by the compat test above)
- [ ] New skills/commands registered in `.claude-plugin/plugin.json` — n/a, no new skills/commands
- [ ] Version bump — n/a, not a release PR
- [ ] CHANGELOG.md updated — left for the release PR that picks this up

## Testing

- `bash -n` across all shell scripts and lib files: clean.
- Reproduced the exact bug in isolation (`set -e`; wait a job to reap it; bare `kill` on the reaped PID aborts the script) and confirmed `|| true` fixes it.
- Confirmed the `cursor-agent.sh` `if/else` idiom still returns the wrapped command's real exit code, by stubbing a command that returns 7 and asserting the wrapper returns 7 (not swallowed to 0).
- `tests/unit/test-cursor-agent-provider.sh` — 8/8 pass.
- `tests/unit/test-heartbeat-timeout.sh` — 12/12 pass.
- `tests/unit/test-openclaw-compat.sh` — 32/32 pass.
- Full `make test-unit` (`tests/run-all.sh unit`) was still running in the background past this session's foreground command budget with no failures observed through the Council/benchmark suites; will comment here if it turns up anything relevant once it finishes.
- `make test-integration` not run — this change only touches cleanup-path error handling around already-exited PIDs in three lib files with no behavioral change to the happy path, so the unit coverage above is the directly affected surface.

## Related Issues

Part of #751 — fixes the "five sites still live" list. The issue's other ask, a CI-enforced portability-lint rule to catch this bug class automatically, is **not** included here: a robust version needs a suppression-comment escape hatch (as the issue notes) and careful false-positive validation against the existing codebase, which is more design work than fits a single automated daily-triage pass. Left open for a follow-up; commented on the issue with this scope split.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1J3QUmsEjbGHetV4915QF)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling to preserve the wrapped command’s exit status.
  * Prevented premature workflow termination when monitored processes exit during cleanup.
  * Ensured progress summaries are displayed even when child-process termination or waiting encounters expected failures.

### DIFF
--- a/scripts/lib/cursor-agent.sh
+++ b/scripts/lib/cursor-agent.sh
@@ -52,8 +52,16 @@ _cursor_agent_run_with_timeout() {
     ( /bin/sleep "$timeout_secs"; kill -TERM "$cmd_pid" 2>/dev/null; /bin/sleep 1; kill -KILL "$cmd_pid" 2>/dev/null ) &
     monitor_pid=$!
 
-    wait "$cmd_pid" 2>/dev/null
-    exit_code=$?
+    # wait's own return reflects $cmd_pid's exit status, so a nonzero exit from
+    # the wrapped command would abort this function under set -e before
+    # exit_code is even captured. Branch on it instead of using `|| true`,
+    # which would discard the real exit code. Same idiom as heartbeat.sh's
+    # in-process fallback. (#751)
+    if wait "$cmd_pid" 2>/dev/null; then
+        exit_code=0
+    else
+        exit_code=$?
+    fi
 
     kill "$monitor_pid" 2>/dev/null || true
     wait "$monitor_pid" 2>/dev/null || true
--- a/scripts/lib/heartbeat.sh
+++ b/scripts/lib/heartbeat.sh
@@ -192,12 +192,16 @@ run_with_timeout() {
 
         # oco-dar: SIGTERM at the cap, then SIGKILL the process AND its children
         # 10s later so a TERM-ignoring tree cannot wedge the workflow.
+        # kill on a $cmd_pid that already exited (race with the wait below)
+        # returns non-zero; under set -e (inherited into this subshell) that
+        # would abort it and skip the pkill sweep and the SIGKILL escalation.
+        # Same class as the monitor kill/wait guarded below. (#751)
         (
             sleep "$timeout_secs"
-            kill -TERM "$cmd_pid" 2>/dev/null
+            kill -TERM "$cmd_pid" 2>/dev/null || true
             pkill -TERM -P "$cmd_pid" 2>/dev/null || true
             sleep 10
-            kill -KILL "$cmd_pid" 2>/dev/null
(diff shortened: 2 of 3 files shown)


nyldn:
Superseded by #754, which fixes the same five sites. Both are correct — closing on test coverage, not on the fix.

Credit where due: **both PRs avoided the trap here**, which is the part that matters. `cursor-agent.sh:55` is not like the other four sites — its `wait "$cmd_pid"` feeds `exit_code=$?` on the very next line, so a blind `|| true` would have silently zeroed the wrapped command's real exit status, turning a failed provider into a reported success. That would have been worse than the abort it replaced.

#753 handles it with `if wait …; then exit_code=0; else exit_code=$?; fi`; #754 with `exit_code=0; wait … || exit_code=$?`. Both correct, both initialise the success path.

**#754 adds `tests/unit/test-seteleak-kill-wait-751.sh` (6 cases) and #753 adds none.** I mutation-checked it by applying exactly the naive fix — `wait … || true; exit_code=$?` — and the suite fails with exit 1. So the coverage specifically pins the subtlety both of you spotted, which is what stops the next person 'simplifying' it back.

Given #751 exists because #739 fixed one site and left five, a regression test on the remaining five is the thing that closes the class rather than the instance.

Issue #752, opened 2026-08-03
Two cheap CI additions: assert every test is reachable, and run the unit suite from a symlinked path


nyldn:
Two small CI changes, each closing a whole class rather than an instance. Filed together because both are one-time additions with no ongoing cost.

## 1. Assert every test file is reachable by some gate

#741 found **33 test files in `tests/` root that no CI target has ever run**, including `test-credential-isolation.sh` — a security property that has been asserting nothing. The runner defines a `root` category and its default includes it, but every Makefile target and CI step passes a single category flag, so it never fires.

The lasting fix is not just relocating those 33 files (that is #741) but a **meta-test** that fails when any `tests/**/test-*.sh` is unreachable from `make test-smoke`, `test-unit` or `test-integration`. Otherwise the trap reopens the moment someone adds a file in a new directory.

Implementation: enumerate discoverable files per category, diff against `find tests -name 'test-*.sh'`, fail on any difference. Cheap, no runtime cost, and it makes "I added a test" and "a test runs" the same statement.

This also caught a real case in review: #734's new coverage landed in `tests/test-resolve-model.sh` and would have shipped dead. Three literal-object routing assertions, none of them running.

## 2. Run the unit suite once from a symlinked path

#712: `test-feature-disclosure.sh` failed because it derived `PROJECT_ROOT` logically while `features.sh` resolves physically (`cd -P`). On macOS `/tmp` is a symlink to `/private/tmp`, so the two disagreed.

It bit at the worst moment — using the release worktree that `RELEASING.md` §0 explicitly recommends — and CI could never have caught it, because runners check out under a real path. That asymmetry is the point: **the failure is invisible to CI by construction**, so only a deliberate job surfaces it.

Implementation: one extra CI job that symlinks the checkout and runs `make test-unit` through the link. Any test conflating logical and physical paths fails there and nowhere else.

## Why both, why now

Five bugs this cycle came from tests depending on ambient host state rather than the code under test: a `$HOME` quota marker (#690), reviewer-flip reading the real `PATH`, CLI prereq probes needing stubs, the BSD/GNU `grep -E "\t"` divergence, and #712. Item 2 catches the path subset of that class mechanically. Item 1 catches the more embarrassing case where the test does not run at all.

Neither is glamorous and neither needs design discussion, which is why they should go in before the two larger checks (#749, #750).

nyldn:
Both items shipped and verified on `main`.

[...]

Issue #741, opened 2026-08-03
34 test files are unreachable by any CI gate (not 33, and not only tests/ root)


nyldn:
Found while reviewing #734, whose new coverage landed in `tests/test-resolve-model.sh` and would have shipped dead.

**The gap.** `tests/run-all-tests.sh` defines four categories — `smoke`, `unit`, `integration`, `root` — and its default includes all four (line 137). But nothing invokes the default:

```
Makefile:21                  ci-local: sync-check test-smoke test-unit test-integration
.github/workflows/test.yml   196: make test-smoke   230: make test-unit
                             296: make test-integration   348: make test-e2e
```

Each target passes a single category flag, so the `root` branch at `run-all-tests.sh:~170` never executes. **33 files** matching `tests/test-*.sh` are therefore unrun by every gate, local and remote.

A sample of what is sitting there: `test-credential-isolation.sh`, `test-command-registration.sh`, `test-fleet-diversity.sh`, `test-execution-profile.sh`, `test-lifecycle-commands.sh`, `test-model-config-v849.sh`, `test-enforcement-pattern.sh`.

`test-credential-isolation.sh` is the one that concerns me most — credential isolation is a security property, and if that file has been unrun it has been asserting nothing.

**Why it matters beyond the dead coverage.** The layout invites the mistake: `tests/` root looks like a legitimate home, a file placed there passes review, and CI goes green because the file simply never executes. #734 hit this and it was invisible until I traced the Makefile. Anyone adding a test to the root has silently added nothing.

**Options**

1. **Run the category.** Add `test-root` to the Makefile and a CI step. Honest, but it will surface however many of the 33 are stale or broken — that is the point, though it turns this into a bigger job.
2. **Relocate them.** `git mv` each into `unit/`, `integration/` or `smoke/` as appropriate, fixing the one-level path change (`SCRIPT_DIR/../helpers/...`), then delete the `root` category so the trap cannot reopen.
3. **Triage first.** Run the 33 locally, sort into pass / fail / obsolete, then relocate the passing ones, fix or delete the rest.

Recommend 3, then 2. Running them blind in CI (option 1) means a red main until they are all triaged, and some are likely superseded — `test-model-config-v849.sh` names a version from long ago, and `test-model-config-simple.sh` looks like it may duplicate `unit/` coverage that arrived later.

Whichever route, **delete the `root` category afterwards.** Leaving it defined but uninvoked is what created this.

[...]

PR #757, opened 2026-08-03
test: fail when a new test file is unreachable by any CI gate (#752)

nyldn:
First half of #752. The second half (running the unit suite from a symlinked path) is a CI-workflow change and follows separately.

## The gap

`tests/run-all-tests.sh` defines five categories and its default includes all of them — but nothing invokes the default. The Makefile and `.github/workflows/test.yml` call `test-smoke`, `test-unit`, `test-integration` and `test-e2e` individually, each passing one category. The `root` branch never fires, so **34 test files have never run in any gate** (#741), including `test-credential-isolation.sh` — a security property that has been asserting nothing.

Relocating those is #741. This closes *tomorrow*: a file added anywhere the CI categories do not reach now fails here instead of silently asserting nothing. That case is not hypothetical — #734's new coverage landed in `tests/test-resolve-model.sh` and would have shipped dead.

## Baselined, not landed red

I argued on #751 that a lint which lands red is worthless, so this one doesn't. The 34 known files are recorded in `fixtures/unreachable-baseline.txt` with #741 as owner, because triaging them is a separate job — `test-model-config-v849.sh` names a long-past version and several look superseded by later `unit/` coverage. Two assertions keep the baseline honest:

- **no NEW unreachable file** (names the path in the failure)
- **the recorded count may never grow** — a ratchet, so the debt can only shrink

## Everything derived, nothing hardcoded

Categories come from the make targets CI actually invokes, resolved through each Makefile recipe. Adding a `test-root` target plus a CI step therefore widens what counts as reachable automatically and this suite stops reporting those files. It also follows the `--e2e` compat alias at `run-all-tests.sh:124`, and asserts `discover_tests` still uses `maxdepth 1` — the assumption it mirrors — so a change there fails loudly rather than silently narrowing coverage.

Two exclusions, by **directory** rather than filename so a genuine suite added to either is still caught: `helpers/` (`test-framework.sh` matches the glob but is sourced, not run) and `live/` (deliberately not run in CI — those dispatch real providers and cost money; that exclusion is a decision, not this gap).

## Verification

Mutation-checked both directions:

```
new file in tests/ root      -> 2 failures, path named, ratchet trips 34 -> 35
same file under tests/unit/  -> 7/7 pass
```

## Incidental finding, filed separately

`make test-e2e` maps to the **integration** category via a backward-compat alias, so CI's "End-to-End Tests" check re-runs the same 7 integration suites under a name promising coverage that does not exist. I expected a silent no-op and checked before claiming one — it is redundant work, not missing coverage. Not fixed here.

## Summary by CodeRabbit

* **Tests**
  * Added automated coverage to verify that all applicable test files are included in CI-run categories.
  * Added safeguards for missing baselines, unreachable tests, incompatible categories, empty test discovery, and unexpected growth in unreachable tests.
  * Added a baseline covering 34 existing test scripts across core scenarios, integrations, workflows, performance, and feature checks.


nyldn:

Issue #712, opened 2026-08-01
test-feature-disclosure.sh fails when the checkout is under a symlinked path (e.g. macOS /tmp)


nyldn:
Two cases in `tests/unit/test-feature-disclosure.sh` fail when the repository is checked out under a symlinked path:

```
• manifest resolves when sourced from an unrelated cwd:
    expected /tmp/rel957/config/features.json, resolved /private/tmp/rel957/config/features.json
• stale CLAUDE_PLUGIN_ROOT falls back to the repo manifest:
    stale plugin root should fall back; got /private/tmp/rel957/config/features.json
```

Hit while running `make ci-local` in a release worktree created at `/tmp/rel957`. On macOS `/tmp` is a symlink to `/private/tmp`. The same commit passes 46/46 in a worktree under a non-symlinked path, so this is not a defect in `scripts/lib/features.sh`.

**Cause.** `features.sh` resolves the manifest with `cd -P … && pwd`, which resolves symlinks. The test derives its expected `PROJECT_ROOT` without `-P`, so the two disagree on any symlinked checkout. The code is right; the assertion is comparing a logical path against a physical one.

**Fix.** Resolve `PROJECT_ROOT` in the test with `cd -P` / `pwd -P` so it matches what the library computes.

**Why it is worth fixing rather than avoiding.** CI never sees it because runners check out under a real path, so it only ever bites a developer using a `/tmp` worktree — which `RELEASING.md` §0 explicitly recommends ("Use a separate worktree"). It also reads as a release-blocking failure at exactly the moment someone is cutting a release.

This is the fifth instance this cycle of a test depending on ambient host state rather than on the code under test (the others: the $HOME quota marker in #690, reviewer-flip reading the real PATH, the CLI prereq probes needing stubs, and the BSD/GNU `grep -E "\t"` divergence). Worth a sweep of the unit suite for the same class.

nyldn:
Fixed in #713. Root cause was `tests/unit/test-feature-disclosure.sh:6-7` resolving `PROJECT_ROOT` with logical `cd && pwd` while `scripts/lib/features.sh:89,98` resolves the manifest path with physical `cd -P && pwd` — they diverge on a symlinked checkout. Test-only fix: match the `-P` resolution.

Reproduced the failure by symlinking a checkout and confirmed both cases now pass; 44/46 in the suite locally (2 pre-existing unrelated failures, confirmed identical with this change reverted). Will mark resolved once merged.

---
[...]

PR #758, opened 2026-08-03
ci: run the unit suite through a symlinked path (#752)

nyldn:
Second half of #752. Pairs with #757 (test reachability); independent of it.

## Why a dedicated job

This failure mode is **invisible to every other job by construction.** Runners check out under a real path, so a test deriving `PROJECT_ROOT` with `cd && pwd` while the code under test resolves with `cd -P && pwd` agrees there — and diverges only behind a symlink.

#712 was exactly that: `test-feature-disclosure.sh` failed *solely* in a `/tmp` release worktree on macOS, which is the workflow `RELEASING.md` §0 recommends. So it bit at release time, and no amount of ordinary CI could have warned.

## Verified, not assumed

I reverted #713's `cd -P` fix and ran the same suite both ways:

```
through the symlink : 44 passed, 2 failed
via the real path   : 46 passed, 0 failed
```

The job catches the class; the existing matrix cannot. That asymmetry is the entire argument for adding a job rather than a test.

## Design notes

**The step aborts if logical and physical paths match.** Without that guard, a future runner change could make the symlink a no-op and the job would silently duplicate the ordinary unit run while still reporting green. A check that cannot fail is worse than no check.

**Ubuntu only.** The divergence is a property of symlinked paths, not of macOS, and ubuntu unit is the faster of the two (~6m vs ~8m). Gated on `heavy_tests` like the existing matrix and running in parallel with it, so it adds no wall-clock — only a runner slot.

**No untrusted input.** The `run:` block uses `$GITHUB_WORKSPACE` and `mktemp` only, with no `github.event.*` interpolation anywhere in the job.

## Scope

Five bugs this cycle came from tests reading ambient host state rather than the code under test — a `$HOME` quota marker (#690), reviewer-flip reading the real `PATH`, CLI prereq probes needing stubs, the BSD/GNU `grep -E "\t"` divergence, and #712. This closes the path subset mechanically. The others need per-case hermeticity work and are not addressed here.

## Summary by CodeRabbit

* **Tests**
  * Added Ubuntu-based test coverage for projects accessed through symlinked paths.
  * Added failure log uploads to support diagnosing test failures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -245,6 +245,57 @@ jobs:
           path: /tmp/test_*.log
           retention-days: 7
 
+  # Catches tests that conflate logical and physical paths. This failure mode is
+  # invisible to every other job by construction: runners check out under a real
+  # path, so a test deriving PROJECT_ROOT with `cd && pwd` while the code under
+  # test resolves with `cd -P && pwd` agrees there and diverges anywhere behind a
+  # symlink. #712 hit exactly that — test-feature-disclosure.sh failed only in a
+  # /tmp release worktree on macOS, which is the workflow RELEASING.md §0
+  # recommends, so it bit at release time and CI could never have warned.
+  #
+  # Ubuntu only: the divergence is a property of symlinked paths, not of macOS,
+  # and the ubuntu unit run is the faster of the two. Runs in parallel with the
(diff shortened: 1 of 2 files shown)


nyldn:

Issue #710, opened 2026-08-01
commandcode has no permission-mode env override, unlike OCTOPUS_CODEX_SANDBOX


nyldn:
Follow-up from #696.

`get_agent_command` picks the Command Code permission mode purely from the role:

```bash
local commandcode_mode="plan"
case "$role" in
    implementer|developer) commandcode_mode="yolo" ;;
esac
```

`commandcode-exec.sh` itself accepts `OCTOPUS_COMMANDCODE_PERMISSION_MODE` as a fallback, but dispatch always passes an explicit positional argument, so the env var is dead on the dispatch path — it only applies when the shim is invoked directly.

Codex has `OCTOPUS_CODEX_SANDBOX` (`workspace-write` | `danger-full-access` | `read-only`, validated with a fallback to the default on an invalid value). Command Code should get the equivalent so a user can hold implementer dispatches at `plan` or `default` without disabling the provider.

Note this is a gap, not a regression: `yolo` for implementers is already narrower than gemini and qwen, which pass `--approval-mode yolo` for every role.

Suggested shape, matching the codex precedent in `scripts/lib/dispatch.sh`:

- read `OCTOPUS_COMMANDCODE_PERMISSION_MODE` in `get_agent_command`
- validate against `plan|default|dont-ask|auto-accept|yolo` (the set `commandcode-exec.sh` already enforces)
- log an ERROR and fall back to the role-derived default on an invalid value
- forward it as the positional argument, so the shim's own validation stays the second line of defence

Tests belong in `tests/unit/test-commandcode-harness.sh` alongside the existing role-mapping case: override honoured, invalid value rejected and defaulted, and the resulting command still passing `validate_agent_command`.

nyldn:
Fixed in #718. Root cause was `scripts/lib/dispatch.sh:464-472` — `get_agent_command` derived Command Code's permission mode purely from `$role` and always passed it as a positional arg, so `commandcode-exec.sh`'s own `OCTOPUS_COMMANDCODE_PERMISSION_MODE` fallback never ran on the dispatch path. Added the same env-var override pattern used for `OCTOPUS_CODEX_SANDBOX`, with validation against the shim's allowed set and an ERROR-logged fallback to the role default on an invalid value. Targeted tests pass locally. Will mark resolved once merged.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #759, opened 2026-08-03
test: hold every documented OCTOPUS_* variable accountable (closes #749)

### LINKED ISSUE #749: No test asserts a documented OCTOPUS_* env var actually takes effect (4 were inert this cycle)
**Highest-leverage gap of the four I'm filing.** Four bugs in one release cycle were all the same shape: a documented environment variable that did nothing.

| Issue | Variable | Why it was inert |
|---|---|---|
| #715 | `OCTOPUS_GEMINI_VIA_AGY` | Three dispatch sites read the raw env var, so a recorded disclosure answer never reached them |
| #720 | `OCTOPUS_REVIEWER_FLIP` | `octo_features_choice` only forwards env values matching a declared choice, so `=1` was dropped — while a comment promised it worked |
| #710 | `OCTOPUS_COMMANDCODE_PERMISSION_MODE` | `commandcode-exec.sh` accepted it, but dispatch always passed an explicit positional argument that overrode it |
| — | `invocation: human_only` (not an env var, same class) | Custom key stripped by the build; enforcement was a hardcoded list that disagreed with it |

Three of those four share one root cause, and none was caught by a test.

## The gap

41 distinct `OCTOPUS_*` variables are documented across `CLAUDE.md`, `docs/` and `config/features.json`. 91 test files mention an `OCTOPUS_` var somewhere. **No test enumerates the documented set and asserts each one changes an observable.** So a variable can be documented, read by a shim, and overridden three layers up, with every existing test still green.

## Proposal

A generated test that:

1. Harvests every documented `OCTOPUS_*` variable from `CLAUDE.md`, `docs/**/*.md` and `config/features.json`.
2. For each, looks up an assertion in a table mapping variable → (setup, observable, expected delta).
3. Runs it and asserts the observable changes.
4. **Fails on any documented variable with no table entry**, rather than silently skipping it. That inverted default is the important part — a var added to the docs without an assertion should break the build, which is what makes this self-maintaining.

Not every variable is cheap to assert (some need a live provider). Those get an explicit `skip` entry with a stated reason, so the skip is a recorded decision rather than an omission.

## Why this over more unit tests

Each of the four bugs above had tests around it that passed. The failure was never "nobody tested this function" — it was "the declaration and the enforcement live in different files and nothing links them". A test derived *from the documentation* is the link.

Cross-reference: this is the pattern behind all four issues I'm filing today — see also the dispatch round-trip and the human-only list drift fixed in #726.

nyldn:
Closes #749.

## The problem this addresses

Four variables shipped inert in one release cycle:

| Issue | Variable | Why it did nothing |
|---|---|---|
| #715 | `OCTOPUS_GEMINI_VIA_AGY` | Three dispatch sites read the raw env var, so a recorded answer never reached them |
| #720 | `OCTOPUS_REVIEWER_FLIP` | `octo_features_choice` only forwards values matching a declared choice, so `=1` was dropped — while a comment promised it worked |
| #710 | `OCTOPUS_COMMANDCODE_PERMISSION_MODE` | The shim accepted it; dispatch always passed a positional argument that overrode it |
| #726 | `invocation: human_only` | Custom key stripped at build time; enforcement was a hardcoded list that disagreed with it |

Three share one root cause. **Every one had passing tests around it.**

The failure was never "nobody tested this function". It was that documentation and enforcement live in different files with nothing linking them, so a variable could be documented, read by a shim, and overridden three layers up with every existing test green. This suite is the link: it derives its work list **from the documentation**.

## What it does and does not do

It does **not** attempt to assert all 41 variables — that is a much larger job and most of the remainder need live providers. It asserts that **none is unaccounted for**, which is the property that rots silently. A variable added to `CLAUDE.md`, `docs/*.md` or `config/features.json` without a manifest entry fails here. That inverted default is the whole design.

Manifest breakdown:

- **27 `covered-by`** — points at a suite that exercises the variable
- **10 `skip`** — real variable, assertion impractical, each with a *required* reason. An unexplained skip is indistinguishable from an oversight six months later.
- **4 `placeholder`** — not variables. The harvest regex matches doc template fragments like `OCTOPUS_AGENT_<NAME>` and `OCTOPUS_MYPROV_ALLOWED_MODELS`; without this category the suite would be permanently red on documentation prose.

## Guards, because a manifest rots in more than one way

- a `covered-by` target must exist **and mention its variable** — coverage on paper fails, which is the exact state the four bugs above were in
- a `placeholder` must not actually be read as a variable in `scripts/` or `hooks/`
- stale entries for no-longer-documented variables are reported
- the `covered-by` count has a floor, so coverage can improve but not regress
- the harvest must yield ≥20 variables, or every assertion below it would be vacuous

## Verification

Mutation-checked both directions:

```
new knob added to CLAUDE.md, no manifest entry  -> fails, names OCTOPUS_BRAND_NEW_KNOB
covered-by repointed at a file that never
  mentions its variable                          -> fails, flags (not-mentioned)
```

## Honest limitation

`covered-by` asserts a suite *mentions* the variable, not that it asserts the variable's effect. That is weaker than #749 proposed. Strengthening it means writing real behavioural assertions for the remaining 27, which is worth doing incrementally — moving entries from `skip` to `covered-by`, and then tightening what `covered-by` requires. The floor ratchet is there to make that direction one-way.

## Summary by CodeRabbit

* **Tests**
  * Added automated validation to ensure documented configuration variables are accurately represented in the project manifest.
  * Checks now identify missing or outdated entries, invalid classifications, incomplete references, and missing skip explanations.
  * Added reporting that summarizes validation results for improved consistency and reliability.

### DIFF
--- a/tests/unit/fixtures/env-var-effects.tsv
+++ b/tests/unit/fixtures/env-var-effects.tsv
@@ -1,0 +1,57 @@
+# Env-var effect manifest (#749)
+#
+# Every OCTOPUS_* variable documented in CLAUDE.md, docs/*.md or
+# config/features.json must appear here. tests/unit/test-env-var-effects.sh
+# fails on any documented variable missing an entry — that inverted default is
+# the point: four variables shipped inert in one cycle (#715, #720, #710, and
+# OCTOPUS_COMMANDCODE_PERMISSION_MODE) and no test noticed.
+#
+# Format: <VAR><TAB><kind><TAB><detail>
+#   covered-by  <test file>  a suite that exercises this variable's effect
+#   skip        <reason>     real variable, assertion impractical — reason required
+#   placeholder <note>       not a variable; a template fragment in docs
+#
+# Prefer moving entries from skip to covered-by over time. A skip is a recorded
+# decision, not an excuse.
+
+OCTOPUS_AGENT_	placeholder	doc template fragment, not a real variable
+OCTOPUS_AGENT_LIFECYCLE_HOOK	covered-by	test-agent-lifecycle-events.sh
+OCTOPUS_AGENT_LIFECYCLE_HOOK_LOG	covered-by	test-agent-lifecycle-events.sh
+OCTOPUS_AGY_MODEL	covered-by	test-agy-provider.sh
+OCTOPUS_AUTO_TITLE	skip	sets the Claude Code session title; observable only in the host UI
+OCTOPUS_CLAUDE_SDK_MODEL	covered-by	test-claude-sdk-provider.sh
+OCTOPUS_CODEX_MODEL	covered-by	test-role-mapping-v929.sh
+OCTOPUS_COMMANDCODE_ALLOWED_MODELS	skip	allowlist for a provider with no CLI on CI runners
+OCTOPUS_COMMANDCODE_BIN	covered-by	test-commandcode-harness.sh
+OCTOPUS_COMMANDCODE_MAX_TURNS	covered-by	test-commandcode-harness.sh
+OCTOPUS_COMMANDCODE_MODEL	covered-by	test-commandcode-harness.sh
+OCTOPUS_COMPACT_BANNERS	skip	changes banner prose only; asserting it would pin cosmetic output
+OCTOPUS_EFFORT_OVERRIDE	covered-by	test-execution-mechanism.sh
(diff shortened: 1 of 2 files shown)


nyldn:

PR #762, opened 2026-08-03
refactor: centralize provider identity and capability contracts

Jhacarreiro:
## Summary

- add a Bash 3.2-compatible canonical provider registry for IDs, aliases, CLI commands, organizations, and capabilities
- separate provider identity/capabilities from ordered workflow policy defaults
- make Command Code a first-class provider across model configuration, Council, health, detection, preflight, help, dispatch-related inventory, and allowlists
- replace unsafe first-hyphen provider parsing with canonical provider resolution
- add parity and collision tests covering every registered provider

## Problem

Provider support was spread across multiple independent hard-coded lists. A provider could work in dispatch/model resolution while being rejected by another public interface.

Command Code reproduced this inconsistency directly:

```text
octo-model-config set commandcode ...
→ Unknown provider commandcode

orchestrate.sh council --providers commandcode ...
→ unknown provider commandcode
```

The same drift also affected help, preflight, detection, health aggregation, and provider aliases.

Several consumers additionally inferred provider identity by truncating at the first hyphen:

```text
claude-sdk-reviewer       → claude
openai-compatible-agent   → openai
cursor-agent              → cursor
```

That breaks compound provider IDs and can corrupt metrics, circuit-breaker state, routing decisions, and idle-provider accounting.

### Canonical identity registry

`scripts/lib/provider-registry.sh` stores only shared identity metadata:

- canonical ID
- aliases
- CLI command
- organization/vendor
- declared capabilities

It is source-safe and Bash 3.2 compatible; no associative arrays or newer parameter-expansion features are used.

### Policy remains separate

`scripts/lib/provider-policy.sh` owns ordered workflow defaults such as:

- Council default provider order
- smoke-test routing candidates

These remain configurable through:

```text
OCTOPUS_COUNCIL_DEFAULT_PROVIDERS
OCTOPUS_SMOKE_ROUTING_PROVIDERS
```

The existing upstream Council default is preserved. Command Code is supported explicitly but is not silently inserted into the default roster.

### Canonical compound IDs

Provider-router metrics, cooldown filtering, circuit-breaker inventory, and agent-team idle calculations now resolve providers through the registry instead of splitting on the first hyphen.

### Allowlist behavior

Normal provider IDs and aliases use the registry. Historical ambiguous group aliases remain explicit policy:

```text
google → gemini + agy
xai    → grok + cursor-agent
```

## Provider coverage

The registry currently covers 19 canonical providers:

```text
codex commandcode claude claude-sdk gemini agy perplexity
opencode openrouter atlascloud openai-compatible openai-tools
openai-compatible-agent cursor-agent grok qwen ollama copilot vibe
```

Contract tests verify capability parity for model configuration, Council, health, detection, dispatch, and environment handling.

## Compatibility

This change does not modify:

- configured models
- `providers.json`
- `models.json`
- `openclaw.json`
- `agents/config.yaml`
- model resolver defaults
- global provider/model defaults

## Validation

Focused coverage includes:

```text
Canonical provider registry:           8/8
Consumer contracts:                   10/10
Provider parity and collision safety:  9/9
Compound provider router behavior:     3/3
AGY provider:                          43/43
Gemini provider:                       52/52
Provider reliability:                  30/30
```

Full unit suite:

```text
201 test suites passed
0 failed
```

## Summary by CodeRabbit

* **New Features**
  * Added unified provider handling with canonical IDs, aliases, capabilities, and metadata.
  * Expanded routing and configuration for providers including agy, Qwen, and Claude SDK.
  * Added Command Code detection with API-key or authenticated CLI-session support.
  * Added configurable provider lists for Council and smoke-routing workflows.
* **Documentation**
  * Updated help output and environment documentation with supported providers and authentication details.
* **Bug Fixes**
  * Improved handling of aliased and compound provider IDs across routing, filtering, status, metrics, and dispatch.
  * Added validation and clearer errors for invalid provider policies.


nyldn:
Merging. Verified the headline claim rather than taking it on trust: `provider-routing.sh:23` now derives `OCTO_MODEL_CONFIG_PROVIDERS` from `octo_provider_ids model-config` instead of a hand-maintained list, which is the duplicated-whitelist hazard `CLAUDE.md` warns about and the root cause behind the Command Code dispatch rejection in #696. Confirmed `commandcode` is accepted at runtime by `provider-allowlist.sh`, a file that no longer names it anywhere — with a filter set, `codex` and `commandcode` pass while `gemini` and an unregistered ID are rejected. Clean under real Bash 3.2.57, no associative arrays or `mapfile`. Full local CI parity green (16 smoke, 201 unit, 7 integration) plus 30 assertions across the four new registry suites, and CI is green on both runners.

On the review feedback, disposition recorded in #768 rather than dropped:

**Valid, followed up.** The defensive `source … 2>/dev/null || true` at every consumer does let a failed registry load pass silently, leaving the derived list empty. I tested where that lands, since an empty whitelist is either harmless or a hole: with the registry removed, `codex`, `gemini` and `claude` are all rejected. It fails closed, which is why it did not block here, but it fails closed with no diagnostic — a partial install presents as "model-config rejects every provider I pass." One genuine wart found along the way: `octo_model_config_provider_valid ""` returns 0 in that degraded state, because an empty list makes the case subject `"  "` and an empty argument makes the pattern match. Only reachable once the registry is already missing.

**Valid, minor.** The unconditional `unset OCTO_ALLOWED_PROVIDERS` in two suites should save and restore the caller's state. Nothing fails on it today.

**Declined: mode 644 for `provider-policy.sh`.** Its nearest siblings are all 755 — `provider-routing.sh`, `provider-allowlist.sh`, `usage-help.sh` — and `CLAUDE.md` requires shell scripts stay `100755` with Portability Lint enforcing no mode changes. `council.sh` at 644 is the outlier, not these files.

**Declined: Command Code auth-awareness across preflight and the registry contracts**, and **extending one canonicalization contract through `agent-teams-bridge.sh` and `provider-router.sh`.** Both are reasonable directions and both are larger than the change they were filed against. The second is worth its own issue if alias drift shows up in practice.

Three files still carry independent provider lists — `intelligence.sh`, `provider-policy.sh`, `permissions-manager.sh` — so this consolidates the primary whitelists without finishing the job. Noted in the release notes as follow-up rather than implied complete.

nyldn:
@Jhacarreiro — this is still marked as a draft, so I have not merged it. Converting someone else's draft to ready is your call, not mine. Mark it ready when you are and I will merge it; everything on my side is done.

For the record, so you are not re-verifying: CI is green on both runners, full local CI parity passes (16 smoke, 201 unit, 7 integration), the four new registry suites contribute 30 assertions, and the tree is clean under real Bash 3.2.57. Findings and review disposition are in the comment above and in #768.

One consequence worth flagging: v9.59.0 is going out without this. I had written the registry consolidation into the release notes as the headline change and have removed it, because shipping notes describing a change users would not have is worse than a thinner release. It will lead the next one instead.

nyldn:
@Jhacarreiro — marking this ready and merging. Your `fix: address provider registry review gaps` commit closes the entire #768 review, so there is nothing outstanding.

Verified rather than assumed. The three items I raised:

**Silent registry-load degradation.** Fixed. `provider-routing.sh`, `provider-allowlist.sh`, `usage-help.sh` and `octo-model-config.sh` now fail loudly instead of `source … 2>/dev/null || true`. Confirmed by removing the registry from a copy of the tree: sourcing exits 1 and `octo_model_config_provider_valid` is simply undefined, rather than silently defined against an empty whitelist that rejects every provider with no diagnostic.

**`octo_model_config_provider_valid ""` accepted in the degraded state.** Moot, and correctly so — the degraded state no longer exists, because the load now fails instead of continuing. That is a better fix than guarding the empty argument, which would have left the silent-empty-list path in place.

**Test env restore.** Fixed in both suites via `${OCTO_ALLOWED_PROVIDERS+x}` capture and restore, rather than an unconditional `unset`.

86 assertions green across `test-provider-registry` (8), `-contracts` (14), `-parity` (9), `-router-canonical-ids` (3), `test-agy-provider` (43) and `test-claude-sdk-provider` (9).

One thing worth recording. I was asked to reimplement this as a separate PR while it sat as a draft. I cherry-picked your three commits onto current `main` to preserve your authorship, and the result was **byte-identical** to your branch — `git diff` between them was empty. So the reimplementation would have added nothing except a different name on the PR. Discarded it and merged yours instead, which is the correct outcome.

Three files still carry independent provider lists (`intelligence.sh`, `provider-policy.sh`, `permissions-manager.sh`), so this consolidates the primary whitelists without finishing the job. Noted as follow-up, not implied complete. This will lead the next release; it missed v9.59.0 only because it was still a draft when that went out.

Issue #766, opened 2026-08-04
Codex skips async hook and clamps both SessionEnd hooks to 3 seconds


fkdarven:
## Plugin Version

9.58.0

## Claude Code Version

N/A — reproduced directly with the Codex plugin host.

Codex CLI: 0.146.0

## What happened?

Every fresh Codex process with Octo enabled prints three compatibility warnings before normal hook execution:

```text
warning: skipping async hook in .../octo/9.58.0/hooks/hooks.json: async hooks are not supported yet
warning: clamping SessionEnd hook timeout to 3s in .../octo/9.58.0/hooks/hooks.json
warning: clamping SessionEnd hook timeout to 3s in .../octo/9.58.0/hooks/hooks.json
```

The current shared hook manifest declares:

- `telemetry-webhook.sh` with `"async": true` and a 10-second timeout.
- `session-end.sh` with a 15-second timeout.
- `workflow-verification.sh` with a 10-second timeout.

Codex currently does not support async hooks and caps each SessionEnd hook at three seconds. The async hook is therefore skipped, and both SessionEnd hooks run with a substantially smaller budget than the manifest requests.

The immediate impact is persistent startup noise and silent loss of the optional async behavior. There is also a truncation risk for SessionEnd metrics, learning/handoff persistence, cleanup, or workflow verification if either script exceeds three seconds.

## What did you expect?

Octo's Codex integration should load without host-compatibility warnings and should register only hook behavior supported by Codex.

A Codex-specific hook manifest or adapter could omit/degrade the async webhook until the host supports it and make SessionEnd work fit the supported three-second budget (or omit/merge nonessential work). Any degradation should be explicit and tested.

## Steps to reproduce

1. Install and enable `octo@nyldn-plugins` 9.58.0 in Codex CLI 0.146.0.
2. Start a fresh non-interactive process:
   ```bash
   codex exec --ephemeral --skip-git-repo-check --sandbox read-only --color never "Reply with exactly CONFIG_OK. Do not call tools."
   ```
3. Observe the unsupported-async warning and two SessionEnd timeout-clamp warnings before the normal hook events.
4. The command otherwise completes and returns `CONFIG_OK`.

## Which providers are configured?

- Codex CLI

## Operating System

Linux

## Relevant logs

```shell
OpenAI Codex v0.146.0
[...]

PR #767, opened 2026-08-04
fix: stop Codex from silently skipping the telemetry-webhook hook

nyldn:
## Description

`hooks/hooks.json` declared the `telemetry-webhook.sh` PostToolUse hook with `"async": true`. Codex CLI doesn't support async hooks, and per #766 it doesn't downgrade them to synchronous — it **skips them entirely**, printing `warning: skipping async hook ... async hooks are not supported yet`. That means the webhook hook silently never fires for Codex users who opt in via `OCTOPUS_WEBHOOK_URL`, beyond the cosmetic warning noise.

Fix: drop `async: true`. The hook keeps its existing 10s timeout and `if: env.OCTOPUS_WEBHOOK_URL` gate (opt-in only), so on hosts that do support async it now just blocks briefly instead of running in the background — no behavior change when the webhook isn't configured.

**This is a partial fix for #766.** The other two warnings in that issue (`session-end.sh` and `workflow-verification.sh` SessionEnd hooks clamped to Codex's 3s cap) are intentionally not touched here: `hooks/hooks.json` is shared across every host, and shrinking the declared timeout from 15s/10s to 3s to silence the Codex warning would also cut Claude Code's actual budget for `session-end.sh`, which does several `jq` calls, file writes, and shells out to `write-handoff.sh` and `session-manager.sh cleanup`. There's no per-host hook manifest mechanism in this repo today, so fixing that properly is a design decision (per-host manifest vs. trimming `session-end.sh` to fit 3s), not a one-line change — left a comment on #766 explaining this and asking for direction rather than guessing.

## Type of Change
- [x] Bug fix

## Checklist
- [x] `hooks/hooks.json` is valid JSON (verified with `python3 -m json.tool`)
- [ ] OpenClaw registry in sync — not applicable, no tool/skill/command change
- [ ] New skills/commands registered — not applicable
- [ ] Version bump — not applicable, no release cut in this change
- [ ] Documentation updated — not applicable
- [ ] CHANGELOG.md updated — deferred to whoever finishes the SessionEnd half of #766, so the entry covers the full fix in one line

## Testing

The change is a single-line JSON edit (removing `"async": true"` from one hook entry), so I ran the directly affected surface rather than the full suite:

```bash
python3 -c "import json; json.load(open('hooks/hooks.json'))"   # valid JSON
bash -n scripts/*.sh scripts/lib/*.sh hooks/*.sh                 # syntax check, all pass
bash tests/unit/test-auto-router-hooks.sh                        # 13/13 pass
bash tests/unit/test-docs-sync.sh                                 # 141/141 pass
bash tests/unit/test-github-work-queue-hook.sh                    # 4/4 pass
bash tests/unit/test-hook-err-traps.sh                             # 8/8 pass
bash tests/unit/test-shell-safe-hooks-v2183.sh                    # 8/8 pass
bash tests/unit/test-openclaw-compat.sh                           # 32/32 pass
```

No test in the repo currently pins the `async`/timeout values in `hooks/hooks.json` (confirmed by grep), so nothing was pinning the old behavior and nothing regresses it. #766 itself suggests adding regression coverage for this — that's still open.

## Related Issues
Refs #766 (partial — see PR description and issue comment for what's still open)

---
_Generated by [Claude Code](https://claude.ai/code/session_01SDXrwHfZYqsnR9gd2MZE8k)_

## Summary by CodeRabbit

* **Refactor**
  * Updated telemetry webhook execution settings for more consistent hook behavior.
  * Webhook conditions, command, and timeout remain unchanged.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -113,7 +113,6 @@
           {
             "type": "command",
             "command": "${CLAUDE_PLUGIN_ROOT}/hooks/telemetry-webhook.sh",
-            "async": true,
             "timeout": 10
           }
         ]


nyldn:
Verified rather than reasoned about, because this is a **PostToolUse** hook — it fires on every tool call, so making it synchronous would be serious if the script actually blocked.

It doesn't. `telemetry-webhook.sh:87` already self-backgrounds:

```bash
curl "${CURL_ARGS[@]}" >/dev/null 2>&1 &
```

Backgrounded *and* with stdout/stderr redirected, so it neither waits nor holds the hook's pipe open. Measured against a local endpoint that sleeps 5 seconds:

```
run 1: 37ms   (endpoint sleeps 5000ms)
run 2: 32ms
```

and 20ms in the common case where no webhook is configured. So `"async": true` was redundant with the script's own design, and removing it costs Claude Code nothing while unblocking Codex. Good change.

**One thing it does not close.** #766 reports three warnings and this fixes one. The other two are the SessionEnd timeout clamps:

```
warning: clamping SessionEnd hook timeout to 3s ...   (session-end.sh, declared 15)
warning: clamping SessionEnd hook timeout to 3s ...   (workflow-verification.sh, declared 10)
```

Those are a harder call than this one: Codex caps SessionEnd at 3s, so lowering the declared timeouts to match would silence the warning but could truncate work that legitimately needs longer under Claude Code — the opposite trade to this PR, where the declared value was simply redundant. Worth deciding deliberately rather than folding into this merge, so I'd keep #766 open for those two after this lands.

PR #771, opened 2026-08-04
fix: stop the macOS timing-flake and misleading integration-red on unit failure

### LINKED ISSUE #770: Release blocked by a macOS timing flake, and one unit failure reports as two red checks
The v9.59.0 release commit (`a5d38ee3`) turned `main` red, and `RELEASING.md` step 9 forbids tagging while main is red. The release commit changes only docs and version strings, so it cannot plausibly have broken a test. Two separate problems, recorded separately because only one is a flake.

## 1. `test-review-aggregation-robustness.sh` fails intermittently on macOS

Failing case: `retry wait stops when provider exits without terminal status: waiter remained alive after provider exited`.

The test starts `sleep 1` as a stand-in provider, starts `review_wait_for_result_status` as a background waiter, then polls four times at one-second intervals for the waiter to exit. The waiter should notice the dead provider on its next poll, so it passes with roughly a two-second margin inside a four-second bound. On a loaded runner that margin is not enough.

**I could not reproduce it.** Eleven local runs, three of them under CPU load (18 cores, load average ~5), all green. So the case for this being a flake rather than a regression is by elimination, not reproduction:

- `scripts/lib/review.sh`, which defines both `review_wait_for_result_status` and `review_process_is_running`, is **unchanged** since the last green run on main (`0924d0de`). The complete diff since then is docs, version metadata, `scripts/lib/council.sh`, and one council test.
- The failing test never references council, and `review.sh`'s only occurrence of the string is a `council_role="synthesizer"` literal inside an event emit. There is no path from the changed code to the failing assertion.
- It passed on `ubuntu-latest` in the same run; only `macos-latest` failed.
- The batch commit `ac23a06f` has both a **failing and a passing** run of the same suite, which is direct evidence of nondeterminism independent of this release.

The fix is not a longer sleep. Polling for a process to exit inside a fixed wall-clock budget is the wrong shape for a test that must be able to fail for the right reason; it should wait on the waiter deterministically, or assert the exit condition rather than race it.

## 2. A single unit failure cascades into a second, misleading red check

This one is not a flake and will recur on every unit failure.

`integration-heavy` declares `needs: [classify-changes, unit-required]`, so a failed `unit-required` makes GitHub **skip** it regardless of its `if:` condition. The `integration` gate then runs `if: always()` and evaluates:

```bash
if [[ "${{ needs.classify-changes.outputs.heavy_tests }}" == "false" ]]; then
  echo "Fast path: skipping integration tests for release/docs-only changes."
elif [[ "${{ needs.integration-heavy.result }}" != "success" ]]; then
  exit 1
```

With `heavy_tests=true` and `integration-heavy.result=skipped`, it exits 1. So one flaky unit test produces **two** failing required checks, and the second reads as "Integration Tests failed" when integration never ran at all. I initially diagnosed these as two independent failures because of it, which is the cost of the current wording.

The gate logic is correct to fail closed — a skipped heavy suite must not count as a pass. What is wrong is that it cannot distinguish "skipped because a dependency failed" from "skipped when it was required to run", and reports both as an integration failure. Distinguishing those two, and saying which, would have made the real cause obvious immediately.

nyldn:
## Description

Fixes both problems reported in #770.

**1. macOS timing flake in `test-review-aggregation-robustness.sh`.** The "retry wait stops when provider exits without terminal status" case raced a fixed 4×1s poll budget against `review_wait_for_result_status`'s own ~2s internal detection cycle (`scripts/lib/review.sh:391-429`), leaving only ~1-2s of margin against scheduler jitter — enough to flake on a loaded macOS runner. Replaced the poll with a blocking `wait` on the waiter's own exit (SIGCHLD-driven, no scheduling race), bounded by a generous 15s watchdog so a genuine regression (waiter never detects the dead process) still fails the test instead of hanging forever. Verified the fail path independently by pointing the same wait/watchdog pattern at a process that never exits — it correctly reported failure after the watchdog fired.

**2. Misleading double-red in `.github/workflows/test.yml`.** The `integration` gate (`.github/workflows/test.yml:371-384`) only checked `needs.integration-heavy.result`. `integration-heavy` requires `unit-required` to succeed, so a failed `unit-required` causes GitHub to skip `integration-heavy` outright. `integration` then read that skip as its own failure and reported "Integration Tests failed" — a single flaky/failing unit test produced two red required checks, with the second misattributing a skip as an integration bug. The gate now checks `unit-required.result` first and reports explicitly that it was blocked by Unit Tests before falling through to the original integration-result check. Still fails closed either way; only the diagnosis is now correct.

## Type of Change
- [x] Bug fix

## Checklist
- [x] OpenClaw registry in sync — untouched by this change
- [x] New skills/commands registered — n/a, no new skills/commands
- [x] Version bump — n/a, not a release PR
- [x] Documentation updated — n/a
- [x] CHANGELOG.md updated — n/a, internal CI/test fix, no user-facing behavior change

## Testing

```bash
bash -n tests/unit/test-review-aggregation-robustness.sh   # syntax check
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/test.yml'))"  # YAML validity

bash tests/unit/test-review-aggregation-robustness.sh      # 19/19 pass
make test-smoke                                             # 16/16 suites pass
make test-unit                                               # 203/205 pass
```

`make test-unit`'s 2 failures (`test-feature-disclosure.sh`, `test-hud-smart-mode.sh`) are pre-existing on unmodified `origin/main` — confirmed by stashing this change and re-running both directly against the release commit. Neither touches `review.sh` or the workflow file.

Also confirmed no unintended file-mode drift from local test runs (`git diff --summary` shows only the two content changes, no `mode change` lines) before committing, per this repo's Portability Lint rule.

## Related Issues
Closes #770

---
_Generated by [Claude Code](https://claude.ai/code/session_013mV3LDTpvjhBKk7A9JAXzQ)_

## Summary by CodeRabbit

- **Tests**
  - Improved automated verification reliability by ensuring integration checks run only after required prerequisite checks succeed.
  - Added clearer failure reporting to distinguish blocked integration checks from integration test failures.
  - Added timeout protection for retry-wait verification, preventing stalled checks from running indefinitely.

- **Chores**
  - Updated continuous integration workflows to provide more predictable and actionable validation results.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -371,13 +371,16 @@ jobs:
   integration:
     name: Integration Tests
     runs-on: ubuntu-latest
-    needs: [classify-changes, integration-heavy]
+    needs: [classify-changes, unit-required, integration-heavy]
     if: always()
     steps:
       - name: Verify integration tests passed
         run: |
           if [[ "${{ needs.classify-changes.outputs.heavy_tests }}" == "false" ]]; then
             echo "Fast path: skipping integration tests for release/docs-only changes."
+          elif [[ "${{ needs.unit-required.result }}" != "success" ]]; then
+            echo "Integration Tests did not run: blocked by Unit Tests (${{ needs.unit-required.result }}), not an integration failure."
+            exit 1
           elif [[ "${{ needs.integration-heavy.result }}" != "success" ]]; then
             echo "Integration result: ${{ needs.integration-heavy.result }}"
             exit 1
--- a/tests/unit/test-review-aggregation-robustness.sh
+++ b/tests/unit/test-review-aggregation-robustness.sh
@@ -169,20 +169,27 @@ sleep 1 &
 dead_provider_pid=$!
 review_wait_for_result_status "$dead_result" "$dead_provider_pid" dead-provider "$TEST_TMP_DIR" 02 01 &
 waiter_pid=$!
-waiter_finished=false
-for _ in 1 2 3 4; do
-    if ! review_process_is_running "$waiter_pid"; then
-        waiter_finished=true
-        break
-    fi
-    sleep 1
-done
+# Block on the waiter's own exit (SIGCHLD-driven) instead of racing a fixed
+# poll budget against scheduler jitter — the old 4x1s poll had only ~1-2s of
+# margin over the waiter's own ~2s internal cycle and went red intermittently
+# on loaded macOS runners. A generous watchdog still bounds the test so a
+# real regression (waiter never detects the dead process) fails instead of
+# hanging.
+(
+    sleep 15
+    kill -KILL "$waiter_pid" 2>/dev/null
+) &
+watchdog_pid=$!
+if wait "$waiter_pid" 2>/dev/null; then
+    waiter_finished=true
+else
+    waiter_finished=false
+fi
+kill "$watchdog_pid" 2>/dev/null || true
+wait "$watchdog_pid" 2>/dev/null || true
 if [[ "$waiter_finished" == "true" ]]; then
-    wait "$waiter_pid" 2>/dev/null || true
     test_pass
 else
-    kill -KILL "$waiter_pid" 2>/dev/null || true
-    wait "$waiter_pid" 2>/dev/null || true
     test_fail "waiter remained alive after provider exited"
 fi
 


nyldn:
Verified independently rather than on the strength of the description.

**The test can still fail.** Mutated `review_wait_for_result_status` to never exit; the case failed with the correct message after the watchdog fired. That was the thing worth checking, because replacing a poll with a blocking `wait` is exactly the kind of change that can turn a flaky test into a vacuous one. Unmutated, 19/0 across five consecutive runs.

**One latent sharp edge, not blocking.** `wait "$waiter_pid"` yields the waiter's *exit status*, not merely the fact that it exited, so the assertion now conflates "exited" with "exited 0". I checked: `review_wait_for_result_status` returns 0 in this scenario today, so the test is correct as written. But if it ever changes to signal a dead provider with a non-zero status — a plausible and arguably more correct behaviour — this case goes red reporting "waiter remained alive after provider exited" when the waiter in fact exited promptly. If that happens, the message is the thing that will mislead, not the logic.

**The gate fix is right and fails closed.** It still exits 1 when the heavy suite did not run, which it must, but now distinguishes "blocked by Unit Tests" from a genuine integration failure. That distinction had a real cost: I diagnosed yesterday's red main as two independent failures before finding `integration-heavy` was merely skipped for unmet `needs`.

For the record on the flake itself: I never reproduced it locally, across eleven runs including three under CPU load. What confirmed nondeterminism was the re-run of the *identical* commit going green with no code change.

nyldn:
Thanks for the mutation-testing verification — good catch on the exit-status coupling. Agreed it's real: `wait "$waiter_pid"` surfaces `review_wait_for_result_status`'s exit code, so the assertion is currently correct only because that function always returns 0 today, not because the test asserts "exited" independent of "exited 0".

Since this landed as "not blocking" and the PR merged before I could push a follow-up, I'm not reopening it or filing a new PR for this alone per this run's scope — but noting the fix for whoever picks it up next: swap the exit-status check for a watchdog sentinel file (`touch`ed only when the watchdog fires and has to `kill -KILL`), so `waiter_finished` reflects "exited before the bound" regardless of what status code the waiter returns. That decouples the assertion from `review_wait_for_result_status`'s current always-0 behavior entirely.

---

PR #778, opened 2026-08-06
fix: drop the E2E job that called the removed test-e2e target

nyldn:
#776 removed `make test-e2e` because it ran the integration suites
under a second name. The workflow's `e2e` job still invoked it, so
main went red immediately after that merge with:

    make: *** No rule to make target 'test-e2e'.  Stop.

The job's `if:` restricted it to main pushes and scheduled runs, so
every PR skipped it. PR CI was green on #776 and on the v9.60.0
release PR, and the breakage only appeared once it hit main — after
the release commit had already landed, blocking the tag.

Removing the job rather than restoring the target: it ran the
integration suites, which the `integration` job already covers, so it
was the duplication #776 set out to remove.

Also closes the blind spot that let this through. My own
test-suite-reachability.sh derives CI categories from the workflow's
`make test-*` calls and silently skipped any it could not resolve, so
a step calling a deleted target produced an empty category and no
complaint. It now fails with the exact make error CI would hit.
Mutation-checked by re-adding a step that calls test-e2e.

I also missed this in review: my dangling-reference grep for
test-e2e piped through `head -8` and I read the truncated output as
the complete set.

## Summary by CodeRabbit

* **Tests**
  * Improved validation of test targets referenced by automation workflows, including support for hyphenated target names.
  * Added checks for missing Makefile targets and insufficient workflow test coverage.
  * Improved handling of target definitions to avoid incorrectly parsing unrelated recipes.

* **Chores**
  * Added workflow run controls to cancel superseded pull-request runs while preserving main-branch runs.
  * Removed the end-to-end testing job and its setup, execution, and artifact-upload steps.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,20 @@ on:
 permissions:
   contents: read
 
+# Two pull_request runs for the same SHA have repeatedly raced here, and the
+# loser is reported as a failed required check even though the winner passed —
+# which blocked merges four times while the underlying tests were green.
+# Collapse duplicates per PR so only the newest run reports.
+#
+# cancel-in-progress is deliberately restricted to pull_request. A main-branch
+# run must never be cancelled: RELEASING.md step 9 gates the release tag on the
+# Test Suite for the exact post-squash commit, and cancelling it would either
+# block a valid release or, worse, leave main unverified.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+
 jobs:
   classify-changes:
     name: Classify Changes
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -386,48 +386,6 @@ jobs:
             exit 1
           fi
 
-  e2e:
-    name: End-to-End Tests
-    runs-on: ubuntu-latest
-    timeout-minutes: 45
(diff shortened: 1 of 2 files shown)


nyldn:
Both fixed in `61fa224a`. Verified each against the code first; neither was a live defect, but both were real fragility and the second was mine, introduced in this very PR.

**Comment-blind extraction.** There are no commented `make test-*` lines in the workflow today, so nothing was misbehaving. But the grep matched anywhere in the file, so a commented-out step would either demand a target nobody invokes or let a deleted step keep satisfying the count guard. Comments are now stripped before matching.

**Two spellings of one parse — my inconsistency.** `ci_categories()` matched `test-[a-z0-9]+` while the `workflow_make_targets()` I added matched `test-[a-z0-9-]+`. A target like `test-plugin-name` would parse as `test-plugin` in the first and correctly in the second. They agree today only because no hyphenated target is invoked yet. Both now read through a single extractor — duplicated parsing is exactly the class of bug this PR exists to fix, so having introduced one here was the wrong direction.

**Makefile recipe boundary.** The awk now stops at the next target definition, so a recipe further down cannot donate its `run-all.sh` line to a target whose own recipe has none.

Mutation-checked in both directions rather than just confirming green: a commented `make test-ghost` is correctly ignored (0 failures), and a real `- run: make test-ghost` fails with `workflow calls make target(s) that do not exist: test-ghost`. 9/9 otherwise.

nyldn:
@coderabbitai review

PR #782, opened 2026-08-07
fix(embrace): YAML runtime — await sequential agents, real gate thresholds, stdout contract

borng:
## Summary

Live `/octo:embrace` runs surfaced five defects in the YAML workflow runtime (`scripts/lib/yaml-workflow.sh`). The workflow "completed" every time, but phases were racing their own agents and running context-blind. Diagnosed by correlating result-file mtimes against the run log.

- **Sequential (`parallel: false`) agents were never awaited.** The sequential branch called `spawn_agent` (which backgrounds the provider internally) and fell through — each phase synthesized and moved on while its final agent was still running. Evidence: tangle's synthesis was written at 19:32:11 while its claude agent finished at 19:32:53, so that output never reached the ink phase. Sequential agents now go through `spawn_agent_capture_pid` and block, and phases wait for the `.done` completion markers (written after the result file) before collecting output.

- **Cross-phase context was completely lost.** Phase banners were echoed to stdout inside command-substituted functions, so the "synthesis file path" return value was actually `banner…\npath`; the `-f` check failed and every phase received empty previous-phase context. Decorative output now goes to stderr; stdout carries only the path.

- **`prompt_template` blocks were silently discarded without `yq`.** The awk fallback returned empty and agents ran a bare `role: prompt` fallback, dropping `{{probe_synthesis}}` et al. The fallback now extracts block scalars scoped to phase+provider.

- **Quality-gate thresholds parsed as empty.** The awk fallback only matched 4-space-indented fields, missing the nested `quality_gate.threshold`; the last phase also bled into the document-level `quality_gates:` block (ink reported the consensus 0.75 instead of its own 0.80). The parser now matches any indent within the phase, stops at the next top-level key, and returns non-zero when empty so caller defaults apply.

- **`bridge_mark_task_complete` was never called anywhere**, so `completed_tasks` stayed 0 and `bridge_evaluate_gate` logged "quality gate did not pass" on every phase unconditionally — and nothing enforced it. Completions are now recorded from the `.done` markers, the gate is evaluated on results-produced / agents-spawned against the parsed threshold, and a failed gate halts the workflow (matching the fail-fast behavior of the hardcoded path) instead of warning and continuing. The embrace call site checks the runtime's exit status.

Also: `test-orchestrate-cwd-routing.sh` and `test-agy-research-defaults.sh` now unset `OCTOPUS_GEMINI_VIA_AGY` / `OCTOPUS_AGY_MODEL` / `OCTOPUS_AGENT_TIMEOUT` at suite start — session-level env (e.g. exported via `~/.claude/settings.json`) flipped gemini dispatch to the agy path and failed assertions locally; CI's clean env was unaffected.

## Test plan

- New regression suite `tests/unit/test-yaml-workflow-runtime.sh` (9 cases): parser fallbacks (per-phase thresholds incl. the ink 0.80 case, template extraction and scoping), stdout contract, sequential-agent output present in synthesis, ledger completion recording, gate failure on zero results.
- `test-embrace-fail-fast.sh`, `test-orchestrate-cwd-routing.sh`, `test-agy-research-defaults.sh` green.
- Live end-to-end verification: before the fix, gates logged empty thresholds and "did not pass" on all 4 phases while syntheses raced agents; after, all 4 phases pass with correct thresholds (probe 3/3 @0.5, grasp 1/1 @0.75, tangle 3/3 @0.75, ink 3/3 @0.80), grasp's prompt embeds the actual probe synthesis, and every synthesis is written after all of its agents finish.

## Summary by CodeRabbit

* **Bug Fixes**
  * YAML-driven workflows now stop promptly when a phase fails instead of incorrectly reporting completion.
  * Quality gates verify that expected results are produced before allowing a phase to complete.
  * Improved phase-boundary handling, completion tracking, and workflow failure reporting.

* **Reliability**
  * Sequential workflow steps now wait for preceding agents to finish before synthesis.
  * Improved support for multiline prompts and configurations without optional tooling.

* **Tests**
  * Added regression coverage for parsing, execution, completion tracking, and failure handling.

### DIFF
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -3707,7 +3707,7 @@ ${obs_ctx}"
                 _failed_phase=$(jq -r '.current_phase // empty' "${HOME}/.claude-octopus/session.json" 2>/dev/null)
             fi
             _abort_embrace_phase "${_failed_phase:-unknown}" \
-                "YAML runtime halted: quality gate failed" "$yaml_result"
+                "YAML runtime failed" "$yaml_result"
             return 1
         fi
 
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -3698,7 +3698,11 @@ ${obs_ctx}"
         echo ""
 
         local yaml_result
-        yaml_result=$(run_yaml_workflow "embrace" "$prompt" "$task_group")
+        if ! yaml_result=$(run_yaml_workflow "embrace" "$prompt" "$task_group"); then
+            _abort_embrace_phase "${OCTOPUS_WORKFLOW_PHASE:-unknown}" \
+                "YAML runtime halted: quality gate failed" "$yaml_result"
+            return 1
+        fi
 
         # Mark workflow complete
         export OCTOPUS_WORKFLOW_PHASE="complete"
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -3699,7 +3699,14 @@ ${obs_ctx}"
 
         local yaml_result
         if ! yaml_result=$(run_yaml_workflow "embrace" "$prompt" "$task_group"); then
-            _abort_embrace_phase "${OCTOPUS_WORKFLOW_PHASE:-unknown}" \
+            # run_yaml_workflow runs in a command-substitution subshell, so its
+            # phase exports cannot reach this scope. It records the failed
+            # phase in session.json before returning — read it from there.
+            local _failed_phase=""
+            if command -v jq &>/dev/null && [[ -f "${HOME}/.claude-octopus/session.json" ]]; then
+                _failed_phase=$(jq -r '.current_phase // empty' "${HOME}/.claude-octopus/session.json" 2>/dev/null)
+            fi
+            _abort_embrace_phase "${_failed_phase:-unknown}" \
                 "YAML runtime halted: quality gate failed" "$yaml_result"
             return 1
         fi
--- a/scripts/lib/yaml-workflow.sh
+++ b/scripts/lib/yaml-workflow.sh
@@ -74,24 +74,35 @@ yaml_get_phase_config() {
     local phase_name="$2"
     local field="$3"
 
+    local value=""
     if command -v yq &>/dev/null; then
-        yq eval ".phases[] | select(.name == \"$phase_name\") | .$field" "$yaml_file" 2>/dev/null
+        # Fall back to the phase quality_gate block for nested fields (threshold)
+        value=$(yq eval ".phases[] | select(.name == \"$phase_name\") | .$field // .quality_gate.$field // \"\"" "$yaml_file" 2>/dev/null)
     else
-        # awk fallback for simple fields
-        awk -v phase="$phase_name" -v field="$field" '
+        # awk fallback: match the field at any indent inside the phase block.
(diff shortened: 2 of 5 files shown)


nyldn:
Good find and well evidenced — correlating result-file mtimes against the run log to prove sequential agents were never awaited is exactly the right way to catch a workflow that "completes" while its phases race. `tests/unit/test-agy-research-defaults.sh` (6/6) and `tests/unit/test-orchestrate-cwd-routing.sh` (15/15) pass, and `scripts/lib/yaml-workflow.sh` itself is clean.

**One blocker, in the new test rather than the fix.** `tests/unit/test-yaml-workflow-runtime.sh:50` uses an associative array:

```bash
declare -A expected_thresholds=([probe]="0.5" [grasp]="0.75" ...)
```

`CLAUDE.md` sets a Bash 3.2 floor because macOS ships `/bin/bash` 3.2.57, which has no associative arrays. On that shell the subscript is evaluated arithmetically, so the suite dies on its second assertion:

```
tests/unit/test-yaml-workflow-runtime.sh: line 50: probe: unbound variable
```

Reproduced on `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)`. The suite therefore provides no coverage at all for anyone running system bash on macOS — it aborts before reaching the runtime assertions this PR exists to add.

**CI cannot tell you this, which is why it is green.** GitHub's macOS runners resolve `#!/usr/bin/env bash` to Homebrew bash 5 on PATH, not `/bin/bash`, so `Unit Tests (macos-latest)` passes. And `Portability Lint` only greps for GNU-only `sed` patterns and runs ShellCheck, over `scripts/` and `hooks/` — it never checks for `declare -A` and never scans `tests/`. I am adding that check separately so this class stops depending on someone happening to run the suite on stock macOS.

**Suggested fix**, keeping the same assertion:

```bash
for pair in "probe=0.5" "grasp=0.75" "tangle=0.75" "ink=0.80"; do
    phase="${pair%%=*}"; want="${pair#*=}"
    got=$(yaml_get_phase_config "$EMBRACE_YAML" "$phase" "threshold") || got="(empty)"
    [[ "$got" == "$want" ]] || threshold_failures+=" $phase=$got(want $want)"
done
```

Everything else here looks right to me, and I would like to merge it once the suite runs on the declared floor.

nyldn:
Pushed the Bash 3.2 fix to your branch (`maintainerCanModify` was on) so this is not blocked on a round-trip.

`declare -A` aborted the suite at its **second** assertion on macOS `/bin/bash` 3.2.57 with `probe: unbound variable`, which meant none of the runtime assertions this PR exists to add ever ran on that shell. Replaced with a `key=value` list; verified 9/9 on GNU bash 3.2.57.

CI could not have told you: the macOS runner resolves `#!/usr/bin/env bash` to Homebrew bash 5, so the violation passes there and only bites on a stock macOS shell. #785 has since added a lint for exactly this class, scanning `tests/` as well as `scripts/` and `hooks/`.

The fix itself is good work and I have not touched it. Diagnosing the missing await by correlating result-file mtimes against the run log — synthesis written at 19:32:11 while the claude agent finished at 19:32:53 — is the kind of evidence that makes a race obvious instead of arguable. `test-agy-research-defaults` (6/6) and `test-orchestrate-cwd-routing` (15/15) pass unchanged.

Rebased onto current `main` too. Merging once CI clears.

PR #784, opened 2026-08-07
fix: derive provider intelligence from registry

Jhacarreiro:
## Summary

- make provider intelligence derive canonical provider identity from the provider registry
- preserve concrete agent/variant identity separately from canonical provider telemetry
- canonicalize legacy telemetry at read time without destructive migration
- derive ranking/fairness universe from providers with the dispatch capability instead of a fixed codex/gemini/claude list

## Compatibility

Existing JSONL telemetry is not rewritten. Legacy provider values such as claude-sdk-reviewer and commandcode-research are canonicalized when read.

New telemetry records preserve provider, agent, and provider_input separately so provider-level intelligence stays canonical while variant identity remains available.

Provider policy remains separate: this change does not alter provider-policy defaults or routing allowlists.

## Validation

Focused tests:

- test-provider-registry.sh: PASS
- test-provider-registry-governance.sh: PASS
- test-provider-registry-contracts.sh: PASS
- test-provider-registry-parity.sh: PASS
- test-intelligence.sh: 47 passed, 0 failed

Additional smoke coverage verified record_outcome -> score -> ranking, legacy claude-sdk-reviewer telemetry, command-code alias canonicalization, and the no-jq fallback.

Full unit suite for this worktree before commit: 211 suites passed, 0 failed.

## Summary by CodeRabbit

* **Improvements**
  * Provider performance data now consistently recognizes aliases and provider variants.
  * Historical performance records remain compatible when provider naming changes.
  * Provider rankings now include all providers available through the registry, rather than a fixed set.
  * Ranking results provide fairer comparisons across the available provider inventory.

* **Tests**
  * Added coverage for alias handling, legacy records, registry-based rankings, and unbiased provider comparisons.

### DIFF
--- a/scripts/lib/intelligence.sh
+++ b/scripts/lib/intelligence.sh
@@ -6,12 +6,16 @@
 # Sourced by orchestrate.sh. All functions are prefixed or namespaced
 # to avoid collisions with the main script.
 
-# count_words() lives in lib/utils.sh; pull it in when this file is sourced
-# standalone (tests do) rather than through orchestrate.sh's load order.
+# Shared dependencies must also load when this file is sourced standalone.
+_INTELLIGENCE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 if ! declare -f count_words >/dev/null 2>&1; then
-    if ! source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/utils.sh"; then
+    source "${_INTELLIGENCE_LIB_DIR}/utils.sh" || { return 1 2>/dev/null || exit 1; }
+fi
+if ! declare -f octo_provider_canonical >/dev/null 2>&1; then
+    source "${_INTELLIGENCE_LIB_DIR}/provider-registry.sh" || {
+        echo "intelligence: failed to load provider registry" >&2
         return 1 2>/dev/null || exit 1
-    fi
+    }
 fi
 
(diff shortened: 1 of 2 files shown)


nyldn:
Rebased onto `main` and resolved the conflict I caused.

#789 fixed `grep -c … || echo 0` repo-wide (closes #786, #787) and that touched `intelligence.sh:245`, which this PR also rewrites. I took your version of the refactor and re-applied the safe idiom on top, since your rewrite reintroduced the unsafe form at what is now line 248:

```bash
successes=$(… | grep -c '"outcome":"success"') || successes=0
```

The fallback has to sit on the assignment rather than inside the command substitution: `grep -c` prints `0` **and** exits 1 on no-match, so `|| echo 0` appends a second zero and the next arithmetic comparison dies with `syntax error in expression`. That is what broke two hooks on ordinary user input.

Verified after the rebase: `test-intelligence` 47/47, `test-grep-c-arithmetic` 6/6, `test-dispatch-round-trip` 6/6. The middle one is the guard added in #789, so it passing here confirms the resolution preserved both changes rather than silently dropping mine.

The canonicalization itself looks right — I checked that `codex-fast` collapses to `codex`, `antigravity` resolves to `agy`, `claude-sonnet` to `claude`, and an unregistered ID passes through unchanged rather than being dropped. Aggregating variants while preserving unknown rows is the correct trade for telemetry.

PR #788, opened 2026-08-07
fix: consolidate overlapping tangle subtasks

Jhacarreiro:
## Summary

- deterministically consolidate connected Tangle coding subtasks when the reformatter still leaves overlapping write scopes
- preserve disjoint coding work and reasoning tasks instead of falling back to a monolithic worker
- keep the existing fail-closed behavior if deterministic consolidation still cannot produce safe parallel scopes
- add regression coverage for transitive overlaps and the exact Memory Keeper decomposition that exposed the failure

## Why

Tangle currently retries an unsafe decomposition once through the reformatter, then aborts if any write-scope overlap remains. In the observed Memory Keeper run, the retry improved the decomposition but retained a shared file scope, so execution stopped before parallel dispatch even though the overlap could be repaired deterministically.

This change treats overlapping coding scopes as connected components and merges only the conflicting component. Independent coding tasks remain parallel.

## Validation

- `bash tests/unit/test-tangle-overlap-consolidation.sh` — 4/4 pass
- `bash tests/unit/test-tangle-memory-keeper-decomposition.sh` — 5/5 pass
- `bash tests/unit/test-tangle-write-scope-safety.sh` — 11/11 pass on current `main`
- `git diff --check origin/main...HEAD`
- local supervised Memory Keeper run progressed through reformat -> residual overlap -> deterministic consolidation -> `Step 2: Parallel execution`, where the previous run aborted

Related context: #762

## Summary by CodeRabbit

* **New Features**
  * Automatically consolidates overlapping coding tasks into safe, unified work items while preserving independent reasoning and validation tasks.
  * Rechecks task scope and validity after consolidation before execution continues.

* **Bug Fixes**
  * Prevents valid decompositions from being incorrectly aborted due to repairable overlapping work scopes.

* **Tests**
  * Added coverage for connected, disjoint, transitive, and residual overlap scenarios.
  * Added regression coverage confirming correct task dispatch and parallel write-scope safety.

### DIFF
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -2479,10 +2479,9 @@ Every [CODING] line must include a same-line Files: clause."
     if ! parallel_safety_reason=$(tangle_validate_parallel_write_scopes "$subtasks"); then
         log WARN "Reformatted decomposition still overlaps (${parallel_safety_reason}); consolidating connected coding scopes"
         subtasks=$(tangle_consolidate_overlapping_subtasks "$subtasks")
-        # Diagnostics belong on stderr with the log WARN above, not on stdout
-        # where a caller piping this workflow would collect them as output.
-        log INFO "Consolidated subtasks:"
-        log INFO "$subtasks"
+        echo -e "${CYAN}Consolidated subtasks:${NC}"
+        echo "$subtasks"
+        echo ""
         parseable_subtask_count=$(tangle_parseable_subtask_count "$subtasks")
         parseable_coding_subtask_count=$(tangle_parseable_coding_subtask_count "$subtasks")
         parallel_safety_reason=""
--- a/scripts/lib/workflows.sh
+++ b/scripts/lib/workflows.sh
@@ -2479,9 +2479,10 @@ Every [CODING] line must include a same-line Files: clause."
     if ! parallel_safety_reason=$(tangle_validate_parallel_write_scopes "$subtasks"); then
         log WARN "Reformatted decomposition still overlaps (${parallel_safety_reason}); consolidating connected coding scopes"
         subtasks=$(tangle_consolidate_overlapping_subtasks "$subtasks")
-        echo -e "${CYAN}Consolidated subtasks:${NC}"
-        echo "$subtasks"
-        echo ""
+        # Diagnostics belong on stderr with the log WARN above, not on stdout
+        # where a caller piping this workflow would collect them as output.
+        log INFO "Consolidated subtasks:"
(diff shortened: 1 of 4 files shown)


nyldn:
Rebased onto `main` and applied the first finding directly (`maintainerCanModify` was on), so this is not blocked on a round-trip.

**Finding 1 — accepted.** The consolidation block used raw `echo` beside a `log WARN` in the same conditional, so identical diagnostics went to two different streams. `log` writes to stderr (`orchestrate.sh:16-19`), which is the right stream here. Worth being precise about the impact: `_tangle_develop_in_workspace` is called directly rather than in command substitution, so nothing was being corrupted today — but a caller piping this workflow's stdout would have collected the subtask listing as output.

**Findings 2 and 3 — reasonable, not blocking, and I have not applied them.** Both ask to strengthen existing test fixtures: a more demanding transitive-overlap arrangement, and asserting dispatched prompts and counts rather than just `TANGLE_STATUS`. Sharper assertions than what is there. But the current tests do fail when the behaviour breaks, which is the bar for merging, and rewriting another author's test fixtures to a bot's preferred shape is a poor trade against the risk of changing what they actually cover. Worth doing as a follow-up by whoever next touches that suite.

Verified after the rebase: `test-tangle-overlap-consolidation` 4/4, `test-tangle-write-scope-safety` 11/11.

The consolidation logic itself reads well — collapsing transitively overlapping write scopes into one component is the right shape, given the alternative is parallel agents writing the same paths.

nyldn:
Correcting myself: I applied finding 1 (`echo` → `log INFO`) and it **broke a passing test**. Reverted.

`tests/unit/test-tangle-memory-keeper-decomposition.sh:51-54` captures the run and greps it for `Consolidated subtasks:`. Reproduced locally — 4 passed / 1 failed with the change, 5/5 without. CI agreed: Unit Tests failed on both ubuntu and macos, plus Symlinked Path and Integration.

The reason it is not equivalent: `log` is not defined in `workflows.sh`. It comes from `orchestrate.sh`, so routing through it depends on what sourced the lib, and in that test's context the line produced nothing at all. My earlier comment reasoned about which *stream* the output lands on and never checked whether the function was even in scope.

So all three findings are now declined, and I was wrong to treat this one as the safe, obvious fix. It read as a style nit and had a functional consequence. Apologies for the churn on your branch.

Your code is unchanged from what you submitted, aside from the rebase onto current `main`. `test-tangle-overlap-consolidation` 4/4, `test-tangle-write-scope-safety` 11/11, `test-tangle-memory-keeper-decomposition` 5/5.

PR #790, opened 2026-08-08
fix: governed council reliability on sandboxed macOS

### LINKED ISSUE #3: Versioning question
Hey, I'm going to try your framework today, but I'm very curious what led you to the decision to start version from number 7? 😅

klass-723:
## Problem

Running the governed council under Claude Code's sandbox on macOS, `orchestrate.sh council` launches but the seats fail, so the run abandons the runner and falls back to direct single-provider (`codex exec` / `agy`) dispatch — losing the multi-vendor council. Three independent causes:

### 1. `agy-exec.sh` — normalize `--print-timeout` to a valid Go duration
agy's `--print-timeout` is a Go duration and **requires a unit**. A bare integer fails hard:
```
invalid value "600" for flag -print-timeout: time: missing unit in duration "600"   (exit 2)
```
…which takes the whole agy seat down. `OCTOPUS_AGY_PRINT_TIMEOUT` is now normalized — a plain number is treated as seconds (`600` → `600s`), while unit-suffixed values (`600s`, `5m0s`) pass through untouched. (The built-in default `5m0s` was already fine; this hardens the env-override path.)

### 2. `scripts/mcp-memory-bridge.sh` — guard `timeout`
macOS ships neither `timeout` nor `gtimeout`, so the two bare `timeout …` calls are a latent `command not found` (127). Routed through a `_mcp_run` helper that uses `timeout`/`gtimeout` when present and runs the tool directly otherwise — mirroring the guard already in `scripts/lib/spawn.sh`.

### 3. `skill-council` — document the sandbox-safe invocation
The sandbox's unsandboxed-command allowlist is a **leading-anchored prefix match**, so a command that begins with `cd …` or an inline `VAR=… ` env-assignment never matches an `…/orchestrate.sh *` allowlist entry — the runner then executes **sandboxed**, and its `codex`/`agy` seat children fail to write temp files (`Operation not permitted`; agy `mkstemp` in `/var/folders/…/T`). The skill now requires `orchestrate.sh` to be the **leading token** and to pass a worktree via `--dir <path>` (the runner already honors `-d/--dir`) instead of `cd <path> && …`.

## Verification

- New unit test: a bare-number `OCTOPUS_AGY_PRINT_TIMEOUT` is unit-suffixed in agy's argv. **agy suite 44/44.**
- `_mcp_run` falls through to direct execution on a host without `timeout`/`gtimeout` (verified on macOS).
- `orchestrate.sh` confirmed to honor `-d/--dir`. No file-mode drift.

Note: the specific `cd … && … orchestrate.sh` wrapper that triggers cause #3 typically lives in the *caller's* governed-run harness, not this repo — fix #3 documents the required invocation shape so callers can correct it; the runner side already supports `--dir`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for sandboxed macOS council operations.
  * Numeric timeout settings are now interpreted correctly as seconds.
  * Memory-related command operations continue working when common timeout utilities are unavailable.

* **Documentation**
  * Added guidance for launching the council runner safely in sandboxed environments, including worktree selection and command formatting.

* **Tests**
  * Added coverage to verify correct handling of numeric timeout values.


nyldn:
Verified the core claim against `main` rather than the description. `agy-exec.sh:26` passes `OCTOPUS_AGY_PRINT_TIMEOUT` straight through to `--print-timeout`, so the built-in default `5m0s` is fine but any caller setting a bare number hands agy an invalid Go duration and the seat dies with exit 2. Real bug on the override path.

The normalization is correct and Bash 3.2 safe — a `case` glob rather than a regex:

```
600   -> 600s      600s  -> 600s
30    -> 30s       5m0s  -> 5m0s
abc   -> abc       (left alone; agy reports it, which is right)
```

Leaving a non-numeric value untouched is the correct call: silently rewriting `abc` would hide a caller error behind a plausible-looking default.

Merging.

PR #791, opened 2026-08-08
fix: guard Ollama and Codex CLI against unbounded OSS-model auto-pull

Adam2kg:
## Summary
- Guard the Ollama provider fallback path so it no longer silently auto-pulls arbitrarily large OSS models on first use
- Apply the same pull guard to the Codex CLI OSS-model path
- Add unit test coverage for both guards

## Test plan
- [ ] `tests/unit/test-ollama-pull-guard.sh` passes
- [ ] `tests/unit/test-codex-pull-guard.sh` passes
- [ ] Manual: trigger Ollama fallback with a large model and confirm the guard prompts/blocks instead of auto-pulling

## Summary by CodeRabbit

* **New Features**
  * Added guarded execution for local Ollama and Codex OSS models.
  * Missing models are not downloaded automatically unless explicitly enabled.
  * Configurable download-size limits help prevent unexpectedly large pulls.
  * Cloud Codex models continue to run without local-model restrictions.
  * Added clearer handling for unavailable tools, invalid model settings, and failed downloads.

* **Bug Fixes**
  * Prevented unsafe or unapproved automatic model downloads.

* **Tests**
  * Added coverage for model routing, download controls, size limits, and execution safeguards.


nyldn:
@Adam2kg — closing this as already fixed, and I want to show the evidence rather than just assert it, because you clearly diagnosed a real incident.

The guard you describe is already on `main`, merged as **#532** ("fix: guard codex and ollama providers against unbounded OSS-model auto-pull"). It cites the same failure — a provider-failure cascade falling back to local Ollama and kicking off an unbounded multi-GB pull with no human in the loop, ~42 GB in the observed case.

Every file this PR touches already exists there:

| file | on `main` |
|---|---|
| `scripts/helpers/ollama-run.sh` | present |
| `scripts/helpers/codex-run.sh` | present |
| `scripts/helpers/ollama-pull-guard.lib.sh` | present |
| `tests/unit/test-ollama-pull-guard.sh` | present, 9/9 |
| `tests/unit/test-codex-pull-guard.sh` | present, 11/11 |

The shared decision lives in `ollama-pull-guard.lib.sh` so both vectors decide identically, and dispatch routes all Ollama execution through the one shim, which is the property that makes a cascade unable to sidestep it.

**One thing worth flagging, because it is why I did not just resolve the conflicts and merge.** Rebasing this branch produced a conflict in `scripts/lib/utils.sh` where your side carries an older shim allowlist:

```bash
if [[ "$cmd_executable" == */vibe-exec.sh || "$cmd_executable" == */ollama-run.sh ]]; then
```

`main` now also lists `grok-exec.sh` and `claude-sdk-exec.sh` there, added in #769 because their absence meant **every grok and claude-sdk dispatch was rejected before the CLI ran**. Merging this branch as-is would have reverted that and silently broken two providers. That is a good argument for rebasing early on a long-lived branch, and it is on me that #769 landed while this was open.

If you have a case the current guard misses — a fallback path that reaches `ollama run` without going through the shim, or a Codex OSS model the cap does not catch — please reopen with that specific path and I will treat it as a live gap. The diagnosis here was sound; it just arrived after the same fix.

PR #805, opened 2026-08-08
fix: stop grok/openrouter/vibe/atlascloud inheriting codex's default model

### LINKED ISSUE #797: Four providers resolve to another vendor's model (grok dispatches gpt-5.6-sol to the xAI CLI)
Four providers dispatch another vendor's model. Found during a model-currency audit, not reported by a user — it produces no error, just wrong output from the wrong model.

## Reproduction

With no user config (`HOME` empty, i.e. the shipped state):

```bash
for l in provider-registry utils model-resolver provider-routing dispatch; do
  source scripts/lib/$l.sh
done
get_agent_model grok probe researcher
```

| provider | resolves to | should be |
|---|---|---|
| `grok` | `gpt-5.6-sol` | an xAI model |
| `openrouter` (bare) | `gpt-5.6-sol` | a namespaced OpenRouter ID |
| `atlascloud` | `gpt-5.6-sol` | an AtlasCloud model |
| `vibe` | `gpt-5.6-sol` | vibe's own config value |

## Root cause

`scripts/lib/model-resolver.sh:420-447` has arms for codex, gemini, commandcode, agy, claude-sdk, claude, perplexity, the openrouter *sub-types*, openai-compatible, ollama, copilot, qwen, cursor-agent and opencode — then falls through:

```bash
*) resolved_model="$(codex_default_model)"   # :446
```

`codex_default_model` is `gpt-5.6-sol` (`model-resolver.sh:66-68`). Any provider without an explicit arm inherits OpenAI's frontier model.

## Why grok is the worst case

The full path: `dispatch.sh:443-455` resolves the model, sees a value that is not `"default"`, and emits `env OCTOPUS_GROK_MODEL=gpt-5.6-sol .../grok-exec.sh`. Then `grok-exec.sh:14,17-19` appends `--model gpt-5.6-sol`, handing an OpenAI model ID to the xAI CLI. The comment at `grok-exec.sh:4` claiming "default: grok's own default" is only true if dispatch is never used.

Bare `openrouter` is similar: OpenRouter requires namespaced IDs (`openai/gpt-…`), so an unnamespaced `gpt-5.6-sol` is not a valid model there either.

## Why CI does not catch it

`tests/unit/test-grok-provider.sh:75-77` writes `{"providers":{"grok":{"default":"grok-4-fast"}}}` into a temp config before asserting. Nothing exercises resolution in the shipped no-config state, which is the state every new install is in.

## Suggested fix shape

Add explicit arms, and make the guard derivable rather than hand-maintained: `provider-registry.sh` already carries an `organization` field per provider (verified: `grok`→`xai`, `codex`→`openai`). A test can assert that no provider resolves to a model whose vendor prefix disagrees with its registered organization, so a future provider added without an arm fails immediately instead of silently inheriting OpenAI's default.

Related: #696, #697, #705, #769 were the same class of drift — a list that did not include a provider — but each of those produced a loud dispatch rejection. This one is silent.

nyldn:
## Description
`model-resolver.sh`'s Tier-7 hardcoded-default case (the last-resort fallback used when no env var, session override, or `providers.json` entry supplies a model) had no arm for `grok`, bare `openrouter`, `vibe`, or `atlascloud`. All four fell through to the catch-all `*)` arm, which calls `codex_default_model()` (`gpt-5.6-sol`) — an OpenAI model ID handed to non-OpenAI CLIs with no error, just wrong output.

Root cause: `scripts/lib/model-resolver.sh:420-447`.

Repro (shipped no-config state — a fresh install with no `~/.claude-octopus/config/providers.json`):
```bash
for l in provider-registry utils model-resolver provider-routing dispatch; do source scripts/lib/$l.sh; done
get_agent_model grok probe researcher   # was: gpt-5.6-sol, an OpenAI ID handed to the xAI CLI
```

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Fix
Added explicit arms to the Tier-7 case in `scripts/lib/model-resolver.sh`:
- `grok*` / `vibe*` → `"default"` — both CLIs already treat the `"default"` sentinel as "use the CLI's own configured default" (`dispatch.sh` and `grok-exec.sh` only pass `--model` when the resolved value isn't `"default"`; `vibe-exec.sh` never wires a `--model` flag at all, relying on `~/.vibe/config.toml`).
- bare `openrouter)` → `"anthropic/claude-sonnet-4"` — OpenRouter requires a namespaced `vendor/model` ID; an unnamespaced OpenAI string isn't valid there either. This mirrors the default arm of `get_openrouter_model()` in `providers.sh`.
- `atlascloud*` → `""` — Atlas Cloud has no safe universal default (it's a routed OpenAI-compatible endpoint whose catalog depends on the user's account). `atlascloud-agent` dispatch (`dispatch.sh:372-380`) already requires an explicit `ATLASCLOUD_MODEL`/`OCTOPUS_ATLASCLOUD_MODEL`/config pin and errors without one; this makes `resolve_octopus_model` fail closed the same way instead of silently returning a bogus model. Callers already handle a `resolve_octopus_model` failure gracefully (`2>/dev/null || echo unresolved`/`|| true`), verified by grep across all call sites.

## Checklist
- [x] Shell scripts pass `bash -n` syntax check (`bash -n scripts/*.sh scripts/lib/*.sh scripts/helpers/*.sh`)
- [x] New/changed file permissions unchanged (`git diff origin/main...HEAD --summary` has no mode changes)
- [ ] OpenClaw registry in sync: `scripts/build-openclaw.sh --check` (not applicable — no OpenClaw-facing surface touched)
- [ ] New skills/commands registered in `.claude-plugin/plugin.json` (not applicable)
- [ ] Version bump (left for a release PR per repo convention)
- [ ] Documentation updated (not applicable — internal fallback logic only)
- [ ] CHANGELOG.md updated (left for a release PR, per repo convention of batching fix entries)

## Testing
Ran the reproduction from #797 directly (before/after) confirming grok/openrouter/vibe now resolve to vendor-correct values and atlascloud fails closed instead of returning `gpt-5.6-sol`.

Added `tests/unit/test-shipped-state-model-resolution.sh`, which resolves all four providers against a **fresh `HOME` with no `providers.json`** — the exact gap `test-grok-provider.sh` didn't cover (it always seeds a config before resolving, so it never exercised the no-config path every new install actually starts from).

```bash
bash tests/unit/test-shipped-state-model-resolution.sh   # 5/5 pass
bash tests/unit/test-grok-provider.sh                     # 7/7 pass (no regression)
bash tests/unit/test-permission-provider-disclosure.sh    # 9/9 pass
bash tests/unit/test-provider-registry-contracts.sh       # 14/14 pass
bash tests/unit/test-dispatch-round-trip.sh               # 6/6 pass
bash tests/unit/test-resolve-model.sh                      # 17/17 pass
bash tests/unit/test-role-mapping-v929.sh                  # 14/14 pass
bash tests/unit/test-opencode-model-defaults.sh             # 12/12 pass
bash tests/unit/test-usage-report.sh                        # 7/7 pass
bash tests/unit/test-gemini-provider.sh                     # 52/52 pass
bash tests/unit/test-shared-marketplace-sync.sh              # 13/13 pass
bash tests/unit/test-readme-release-sync.sh                   # 8/8 pass
```

Also ran the full `make test-unit` suite (all 242 `tests/unit/*.sh` files): 240/242 passed. The 2 failures are in suites unrelated to model resolution or these four providers — every suite that references `model-resolver.sh`, `resolved_model`, grok, openrouter, vibe, or atlascloud (checked via grep across `tests/`) passes cleanly, individually verified above.

## Related Issues
Closes #797

---
_Generated by [Claude Code](https://claude.ai/code/session_018XyFbJUXKx6JBJM5SgmvuG)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider-specific default model selection for OpenRouter, Grok, and Vibe.
  * OpenRouter now selects a namespaced Claude model when no model is specified.
  * AtlasCloud now requires an explicit model selection instead of using an implicit fallback.
  * Prevented non-Codex providers from inheriting Codex’s default model.

* **Tests**
  * Added regression coverage for model selection in a fresh, unconfigured environment.

### DIFF
--- a/scripts/lib/model-resolver.sh
+++ b/scripts/lib/model-resolver.sh
@@ -435,6 +435,7 @@ resolve_octopus_model() {
             openrouter-glm*)  resolved_model="z-ai/glm-5" ;;
             openrouter-kimi*) resolved_model="moonshotai/kimi-k2.5" ;;
             openrouter-deepseek*) resolved_model="deepseek/deepseek-r1-0528" ;;
+            openrouter)      resolved_model="anthropic/claude-sonnet-4" ;; # bare OpenRouter needs a namespaced ID, not an OpenAI model string (#797)
             openai-compatible|openai-tools|openai-compatible-agent*) resolved_model="${OPENAI_COMPAT_MODEL:-gpt-5.4}" ;;
             ollama*)         resolved_model="llama3.3" ;;
             copilot*)        resolved_model="claude-sonnet-4.5" ;; # Copilot default; actual model selected by copilot CLI
@@ -443,6 +444,9 @@ resolve_octopus_model() {
             opencode-research*) resolved_model="opencode/glm-5.1" ;;
             opencode-fast*)  resolved_model="opencode/deepseek-v4-flash-free" ;;
             opencode*)       resolved_model="opencode/deepseek-v4-flash-free" ;;
+            grok*)           resolved_model="default" ;; # xAI's own default; dispatch.sh/grok-exec.sh omit --model for "default" (#797)
+            vibe*)           resolved_model="default" ;; # Mistral Vibe's own default from ~/.vibe/config.toml; never wired to --model (#797)
+            atlascloud*)     resolved_model="" ;; # No safe universal default; atlascloud-agent dispatch already requires an explicit model pin (#797)
             *)              resolved_model="$(codex_default_model)" ;; # Safest universal fallback
         esac
         [[ -n "$_trace" ]] && echo "[model-trace] Tier 7 (hardcoded fallback): $resolved_model ← SELECTED" >&2
--- a/tests/unit/test-shipped-state-model-resolution.sh
+++ b/tests/unit/test-shipped-state-model-resolution.sh
@@ -101,7 +101,7 @@ test_atlascloud_no_config_fails_closed() {
     local rc out
     rc=0
     out="$(_resolve_in_empty_home atlascloud atlascloud 2>/dev/null)" || rc=$?
-    if [[ "$rc" -ne 0 && "$out" != "gpt-5.6-sol" ]]; then
+    if [[ "$rc" -ne 0 && "$out" != "$(codex_default_model)" ]]; then
         test_pass
     else
         test_fail "expected atlascloud resolution to fail closed with no config, got rc=$rc out='$out'"
--- a/tests/unit/test-shipped-state-model-resolution.sh
+++ b/tests/unit/test-shipped-state-model-resolution.sh
@@ -25,16 +25,32 @@ log() { :; }
 _resolve_in_empty_home() {
     local provider="$1" agent_type="$2"
     local empty_home="$TEST_TMP_DIR/empty-home-$$-$RANDOM"
+    local provider_model_env
     mkdir -p "$empty_home"
-    local old_home="$HOME"
-    HOME="$empty_home"
-    local out=""
-    out="$(resolve_octopus_model "$provider" "$agent_type" "" "" 2>/dev/null)" || true
-    HOME="$old_home"
-    printf '%s' "$out"
+    provider_model_env="OCTOPUS_$(printf '%s' "$provider" | tr '[:lower:]' '[:upper:]' | tr '-' '_')_MODEL"
+    (
+        HOME="$empty_home"
+        unset "$provider_model_env"
+        resolve_octopus_model "$provider" "$agent_type" "" ""
+    )
 }
 
-source "$PROJECT_ROOT/scripts/lib/model-resolver.sh" 2>/dev/null || true
+source "$PROJECT_ROOT/scripts/lib/provider-registry.sh"
+source "$PROJECT_ROOT/scripts/lib/utils.sh"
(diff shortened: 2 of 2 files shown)


nyldn:
Addressed all three review findings in 35f0aca2:

- empty-HOME helpers now isolate HOME and unset provider-specific model variables without mutating the caller
- resolver sourcing now propagates failures
- OpenRouter asserts the exact shipped fallback
- added a registry-derived guard so every non-OpenAI provider is checked against the Codex default

Verification: targeted Bash 3.2 suite passes 6/6; `make sync-check` passed; full `make ci-local` reached 241/242 unit suites with the pre-existing intermittent `test-v8.2.0-agent-fields.sh` failure, which passes 30/30 in isolation.

nyldn:
Addressed the follow-up review in 226883d5: the AtlasCloud assertion now compares against `codex_default_model`, so it tracks the source-of-truth fallback. The isolated Bash 3.2 regression suite remains green at 6/6.

PR #807, opened 2026-08-08
fix: migrate stale gpt-5.x model pins, not just pre-gpt-5 ones

### LINKED ISSUE #798: Config migration misses the gpt-5 line, so existing installs silently stay on gpt-5.4
`migrate_provider_config` rewrites stale model pins in a user's `~/.claude-octopus/config/providers.json`, but its rewrite map is a full generation behind. Real installs therefore keep running the model they were first configured with, while the docs promise the current one.

## Evidence from a live machine

```
$ jq '.providers.codex' ~/.claude-octopus/config/providers.json
default        gpt-5.4
fallback       gpt-5.4
spark          gpt-5.4
mini           gpt-5.4
reasoning      o3
large_context  gpt-5.4
```

`CLAUDE.md` documents GPT-5.6 Sol as the frontier default and `model-resolver.sh:66-68` agrees. This machine silently runs `gpt-5.4` on every codex dispatch, and `o3` for reasoning.

## Root cause

`scripts/lib/provider-routing.sh:435-447` rewrites only:

- `gpt-4o*`
- `o1-*`
- `gemini-1.5-*`
- `gemini-2.0-*`

There is no case for the `gpt-5.x` line, so `gpt-5.4` is not recognised as stale and is left alone.

Second limitation in the same function: it only inspects `.providers.codex.*` and `.providers.gemini.*` (`:424-432`). No other provider's pins are ever migrated, so a stale `perplexity`, `opencode` or `agy` pin persists indefinitely.

## Why this matters more than a stale default

A wrong *default* affects new installs. A missing *migration* affects every existing install, permanently, and is invisible — the user sees no error, just an older model's output. Anyone who configured Octopus before the GPT-5.6 refresh (`972d9597`, 2026-07-27) is still on 5.4 today.

## Suggested fix

Extend the rewrite map to the gpt-5 line and widen the provider scope beyond codex/gemini. Consider driving the map from the model catalog rather than a hand-maintained case list, so a model marked legacy in `models.sh` is migrated automatically — the same derivation argument as #797.

nyldn:
## Description
`migrate_provider_config`'s stale-model rewrite map (`scripts/lib/provider-routing.sh:437-447`) only recognized `gpt-4o`/`gpt-4-turbo`/`o1-*`/`chatgpt-*` as stale for codex — there was no case for the gpt-5.x line itself. An install pinned to `gpt-5.4` (or any other pre-GPT-5.6 point release: `gpt-5.5`, `gpt-5.4-pro`, `gpt-5.3-codex`, `gpt-5.2-codex`, `gpt-5.1-codex-max`) was never recognized as stale and stayed on it permanently, silently — even though `CLAUDE.md` and `model-resolver.sh`'s `codex_default_model()` both document `gpt-5.6-sol` as current since the GPT-5.6 refresh (`972d9597`, 2026-07-27). Unlike a wrong *default* (which only affects new installs), a missed *migration* affects every existing install forever.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Fix
Added a case arm matching the `gpt-5.5`/`5.4`/`5.3-codex`/`5.2-codex`/`5.1-codex-max` prefixes, rewriting them to `gpt-5.6-sol` via the same `jq --arg` path already used for the existing entries. The new globs don't overlap `gpt-5.6-*` itself, so already-current configs are left untouched (covered by a test case).

### Scope note
The issue also flags a second, separate limitation: `stale_paths` only inspects `.providers.codex.{default,fallback}` and the gemini equivalents, so `codex.spark`/`mini`/`reasoning`/`large_context` and every non-codex/gemini provider's pins are never migrated at all. That's a materially larger change (it needs a staleness definition per provider/slot) and is left as follow-up rather than guessed at here. This fix is scoped to the issue's primary, clearly-evidenced symptom — the live-machine repro showing `gpt-5.4` surviving migration on every codex slot the function *does* check. The gap is documented explicitly with a regression test asserting `codex.spark` is **not** touched by this change, so a future widening of `stale_paths` updates that assertion instead of silently leaving it stale itself.

## Checklist
- [x] Shell scripts pass `bash -n scripts/*.sh scripts/lib/*.sh scripts/helpers/*.sh`
- [x] New/changed file permissions unchanged (`git diff origin/main...HEAD --summary` has no mode changes; new test file created at the standard `755`)
- [ ] OpenClaw registry in sync (not applicable — no OpenClaw-facing surface touched)
- [ ] New skills/commands registered in `plugin.json` (not applicable)
- [ ] Version bump (left for a release PR per repo convention)
- [ ] CHANGELOG.md updated (left for a release PR, per repo convention of batching fix entries)

## Testing
Added `tests/unit/test-migrate-stale-gpt5-line.sh`, reproducing the issue's exact live-machine `providers.json` (every codex slot pinned to `gpt-5.4`) against a real `migrate_provider_config()` run. Verified the test fails on unmodified `main` (2/6 cases fail: `codex.default`/`codex.fallback` stay at `gpt-5.4`) and passes with this fix (6/6).

```
bash tests/unit/test-migrate-stale-gpt5-line.sh   # 6/6 pass (was 4/6 on main)
```

Also ran the 18 other unit tests that source `provider-routing.sh` or call `migrate_provider_config` — all pass, no regressions:
```
test-agent-command-validation, test-agy-provider, test-agy-research-defaults,
test-claude-sdk-provider, test-codex-pull-guard, test-codex-sandbox-mode,
test-commandcode-harness, test-credential-isolation, test-dispatch-round-trip,
test-gemini-via-agy-consent, test-grok-provider, test-openai-compatible-agent,
test-provider-history-toggle, test-provider-registry-contracts,
test-provider-registry-parity, test-provider-registry, test-quota-fast-fail,
test-role-mapping-v929
```

Did not run the full `make test-unit`/`make test-integration` suite in this sandbox (a full run exceeded the available command timeout here); the targeted set above covers every suite that touches the changed function, consistent with "if the change is tiny and integration is irrelevant, run only the directly affected tests."

## Related Issues
Closes #798

---
_Generated by [Claude Code](https://claude.ai/code/session_01LSqUF4TbRc24GztjZueTqm)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated migration handling for outdated GPT-5.x and Codex model names.
  * Preserved current Gemini, GPT-5.6, and supported tier-specific model configurations.
  * Improved handling of unsupported model paths without changing valid provider settings.

* **Tests**
  * Added regression coverage for stale model defaults, fallback behavior, preserved configurations, and unsupported model paths.

### DIFF
--- a/scripts/lib/provider-routing.sh
+++ b/scripts/lib/provider-routing.sh
@@ -445,8 +445,9 @@ EOF
                 replacement="gemini-3-pro-image" ;;  # shutdown 2026-06-25 (codex review)
             gpt-4o*|gpt-4-turbo*|gpt-4-*|o1-*|chatgpt-*)
                 replacement="gpt-5.6-sol" ;;
-            gpt-5.5*|gpt-5.4*|gpt-5.3-codex*|gpt-5.2-codex*|gpt-5.1-codex-max)
-                replacement="gpt-5.6-sol" ;;  # gpt-5.x line predates the GPT-5.6 refresh (#798); gpt-5.6-* itself doesn't match these globs
+            gpt-5.5|gpt-5.5-pro|gpt-5.4|gpt-5.4-pro|gpt-5.3-codex|gpt-5.2-codex|gpt-5.1-codex-max)
+                replacement="gpt-5.6-sol" ;;  # gpt-5.x line predates GPT-5.6 (#798). Exact names, not wildcards —
+                                               # gpt-5.4-mini/gpt-5.3-codex-spark are distinct current tiers, not stale
         esac
 
         if [[ -n "$replacement" ]]; then
--- a/scripts/lib/provider-routing.sh
+++ b/scripts/lib/provider-routing.sh
@@ -445,6 +445,8 @@ EOF
                 replacement="gemini-3-pro-image" ;;  # shutdown 2026-06-25 (codex review)
             gpt-4o*|gpt-4-turbo*|gpt-4-*|o1-*|chatgpt-*)
                 replacement="gpt-5.6-sol" ;;
+            gpt-5.5*|gpt-5.4*|gpt-5.3-codex*|gpt-5.2-codex*|gpt-5.1-codex-max)
+                replacement="gpt-5.6-sol" ;;  # gpt-5.x line predates the GPT-5.6 refresh (#798); gpt-5.6-* itself doesn't match these globs
         esac
 
         if [[ -n "$replacement" ]]; then
--- a/tests/unit/test-migrate-stale-gpt5-line.sh
+++ b/tests/unit/test-migrate-stale-gpt5-line.sh
@@ -89,8 +89,41 @@ assert_stale_model_migrates() {
     [[ "$val" == "gpt-5.6-sol" ]] && test_pass || test_fail "expected gpt-5.6-sol, got: $val"
 }
 
-for stale_model in gpt-5.5 gpt-5.4-pro gpt-5.3-codex gpt-5.2-codex gpt-5.1-codex-max; do
+for stale_model in gpt-5.5 gpt-5.5-pro gpt-5.4-pro gpt-5.3-codex gpt-5.2-codex gpt-5.1-codex-max; do
     assert_stale_model_migrates "$stale_model"
+done
+
+# Regression for a review finding on this fix (#807): the first pass used
+# gpt-5.4*/gpt-5.3-codex* wildcards, which also matched gpt-5.4-mini (budget
+# tier) and gpt-5.3-codex-spark (spark tier) — still-current, differently
+# tiered models a user may have deliberately pinned, not stale defaults. The
+# fix narrowed the case arm to exact names; these must NOT migrate.
+assert_current_tier_variant_is_untouched() {
+    local current_model="$1"
+    local fixture_home="$TEST_TMP_DIR/home-untouched-$current_model"
+    local fixture_file="$fixture_home/.claude-octopus/config/providers.json"
+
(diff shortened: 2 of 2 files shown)


nyldn:
Addressed the review finding in acf0852b by adding table-driven migration coverage for every stale GPT-5 pattern handled by the implementation: gpt-5.5, gpt-5.4-pro, gpt-5.3-codex, gpt-5.2-codex, and gpt-5.1-codex-max.

Verification: targeted Bash 3.2 suite passes 11/11; `make sync-check` passed; full `make ci-local` reached 241/242 unit suites with the pre-existing intermittent `test-v8.2.0-agent-fields.sh` failure, which passes 30/30 in isolation.

nyldn:
Addressed the "Exclude active GPT-5.x model variants" finding in 242b3f88 — good catch, that was a real bug: `gpt-5.4*`/`gpt-5.3-codex*` also matched `gpt-5.4-mini` (budget tier) and `gpt-5.3-codex-spark` (spark tier), which would have silently forced those onto premium `gpt-5.6-sol` if a user had them pinned. Narrowed to exact model names per your suggestion; added regression tests that fail against the wildcard version and pass against the fix.

Skipping the `provider-lockout.sh` silent-failure finding (lines 612-616): that code is outside this PR's diff — it predates this change (introduced by #803, a separate provider-lockout extraction) and isn't touched by the gpt-5.x migration fix here. Changing its error-handling behavior is a legitimate concern but belongs in its own PR against #803's code, not bundled into an unrelated bugfix.

---

PR #824, opened 2026-08-09
fix: make skill-template test hermetic, stop mutating the checkout

### LINKED ISSUE #822: Unit template test rewrites the active checkout during make ci-local
## Summary

`tests/unit/test-skill-templates.sh` runs `scripts/gen-skill-docs.sh` directly against `PROJECT_ROOT`. The generator overwrites tracked `.claude/skills/*/SKILL.md` files from stale templates, so a successful `make ci-local` leaves the checkout dirty and can silently replace current skill behavior with older generated content.

## Reproduction

From a clean checkout:

```bash
git status --short
make ci-local
git status --short
```

Observed after the otherwise-green gate:

```text
 M .claude/skills/flow-define/SKILL.md
 M .claude/skills/flow-deliver/SKILL.md
 M .claude/skills/flow-develop/SKILL.md
 M .claude/skills/flow-discover/SKILL.md
```

The test contains both the direct production-tree generation call and a stale-marker append:

```bash
"$GEN_SCRIPT" > /dev/null 2>&1
echo "# stale marker" >> "$(resolve_claude_skill_path "flow-discover")"
"$GEN_SCRIPT" > /dev/null 2>&1
```

This is especially risky because the same file comments say the flow skills are directly authored since v9.10.2. In an isolated CI worktree, the generator still resolved and rewrote the active task checkout, demonstrating that the test is not hermetic.

## Proposed fix

Run generator mutation checks against a temporary copied fixture/root, add a regression assertion that the tracked source hashes do not change, and retain read-only structural checks against the real checkout.

nyldn:
## Description

`tests/unit/test-skill-templates.sh` ran `scripts/gen-skill-docs.sh` directly against the tracked `.claude/skills/*/SKILL.md` files to exercise `--dry-run` staleness detection, then "restored" them by regenerating from templates. Since those files have been directly authored (not template-derived) since v9.10.2, the regenerate step silently overwrote hand-authored content with template output, leaving the checkout dirty after an otherwise-green `make ci-local`:

```
 M .claude/skills/flow-define/SKILL.md
 M .claude/skills/flow-deliver/SKILL.md
 M .claude/skills/flow-develop/SKILL.md
 M .claude/skills/flow-discover/SKILL.md
```

I reproduced this on `main` (3d3f972f) before making any change: the test suite passes 42/42, but `git status --short` shows the four files above modified afterward.

## Root cause

`scripts/gen-skill-docs.sh:12` hardcodes its plugin root from its own script location, so any invocation — including from a test — always reads and writes the real checkout. The test has no way to point the generator at a throwaway copy.

## Fix

- `scripts/gen-skill-docs.sh:12`: `PLUGIN_ROOT` now honors an optional `GEN_SKILL_DOCS_ROOT` override, defaulting to the existing behavior when unset (no change to normal usage — CI and manual `scripts/gen-skill-docs.sh` calls are unaffected).
- `tests/unit/test-skill-templates.sh`: the mutating checks (fresh-file dry-run, stale-file dry-run) now copy the skill/template/block/command inputs into a `mktemp -d` fixture and run the generator against that fixture via `GEN_SKILL_DOCS_ROOT`, instead of against `PROJECT_ROOT`.
- Added a regression assertion that hashes the real tracked `flow-*.md` skill files before and after the fixture runs and fails if they changed — so a future reintroduction of this bug fails loudly instead of silently.

## Type of Change
- [x] Bug fix

## Checklist
- [x] Shell scripts pass `bash -n` syntax check (`bash -n scripts/*.sh scripts/lib/*.sh scripts/helpers/*.sh` — clean)
- [ ] OpenClaw registry in sync — not applicable, no OpenClaw-facing surface touched
- [ ] New skills/commands registered — not applicable
- [ ] Version bump — not applicable, no release in this PR
- [ ] Documentation updated — not applicable, no user-facing behavior changed (generator default behavior is unchanged when the new env var is unset)
- [ ] CHANGELOG.md updated — deferring to whoever cuts the next release

## Testing

```bash
bash tests/unit/test-skill-templates.sh   # 43/43 (was 42/42 on main, +1 new regression check)
bash tests/unit/test-openclaw-compat.sh   # 32/32
bash -n scripts/*.sh scripts/lib/*.sh scripts/helpers/*.sh   # clean
```

Confirmed on `main` before this change that the test passes but leaves the checkout dirty (see repro above), and confirmed after this change that the checkout stays clean across full and stale/dry-run cycles (checked `git status --short` before/after).

Did not run the full `make test-unit` / `make test-integration` matrix in this session — it exceeded the run's time budget (same constraint noted in the immediately preceding PR #819 today). This change only touches one test file and one script's root-resolution, and `tests/unit/test-skill-templates.sh` is the only file anywhere in the repo that references `gen-skill-docs.sh` (`grep -rl gen-skill-docs tests/ scripts/ Makefile` confirms), so the targeted run above covers everything this change can affect.

## Related Issues
Closes #822

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01M57sX5uxAfnTkpYdbZD9J7)_

## Summary by CodeRabbit

- **New Features**
  - Added support for generating skill documentation from an alternate project root while retaining the existing default behavior.
  - This enables documentation generation in isolated or custom project setups.

- **Tests**
  - Improved generator tests to use isolated temporary fixtures.
  - Added validation that generation does not modify tracked project files.
  - Expanded coverage for stale-file handling and fixture cleanup.

### DIFF
--- a/scripts/gen-skill-docs.sh
+++ b/scripts/gen-skill-docs.sh
@@ -5,11 +5,15 @@
 # Templates use {{PLACEHOLDER}} syntax. Shared blocks from skills/blocks/
 # are resolved first, then metadata placeholders (COMMAND_COUNT, etc.).
 # --dry-run exits non-zero if any generated file differs from committed file.
+#
+# GEN_SKILL_DOCS_ROOT overrides the plugin root the generator reads from and
+# writes into — used by tests to run the generator against a throwaway
+# fixture instead of the tracked checkout.
 
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-PLUGIN_ROOT="$(dirname "$SCRIPT_DIR")"
+PLUGIN_ROOT="${GEN_SKILL_DOCS_ROOT:-$(dirname "$SCRIPT_DIR")}"
 SKILLS_DIR="$PLUGIN_ROOT/.claude/skills"
 BLOCKS_DIR="$PLUGIN_ROOT/skills/blocks"
 DRY_RUN=false
--- a/tests/unit/test-skill-templates.sh
+++ b/tests/unit/test-skill-templates.sh
@@ -109,8 +109,10 @@ done
 # checkout. Copy the inputs gen-skill-docs.sh reads/writes into a throwaway
 # root and point it there via GEN_SKILL_DOCS_ROOT. ────────────────────────────
 
-FIXTURE_ROOT=$(mktemp -d)
-trap 'rm -rf "$FIXTURE_ROOT"' EXIT
+# Nest under TEST_TMP_DIR (already created by test_suite() and cleaned up by
+# the framework's own EXIT trap) instead of a standalone mktemp+trap, which
+# would clobber that trap.
+FIXTURE_ROOT="$TEST_TMP_DIR/gen-skill-docs-fixture"
 
 mkdir -p "$FIXTURE_ROOT/.claude/skills" "$FIXTURE_ROOT/skills/blocks" \
     "$FIXTURE_ROOT/commands" "$FIXTURE_ROOT/agents/personas" "$FIXTURE_ROOT/hooks"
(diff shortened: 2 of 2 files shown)


nyldn:
Addressed all 3 in ae8a514c — verified each against the actual repo before applying:

1. **Env-assignment quoting** (line 124): confirmed this repo has a hard rule against bare-prefix env assignments (secret-scanner false-positive), documented in `CLAUDE.md`. Fixed to `env "GEN_SKILL_DOCS_ROOT=${FIXTURE_ROOT}" "$GEN_SCRIPT" "$@"`.
2. **`TEST_TMP_DIR` convention** (line 113): confirmed `tests/helpers/test-framework.sh` already defines `TEST_TMP_DIR` and registers its own `trap cleanup_test_environment EXIT`. My standalone `mktemp -d` + `trap ... EXIT` was silently overwriting that framework trap rather than composing with it — a real bug, not just style. Fixture now lives under `$TEST_TMP_DIR`, cleaned up by the existing trap.
3. **Stale-output assertion** (line 205): agreed an unrelated generator error could have satisfied the old exit-code-only check. Now also asserts `stale_output` contains `STALE: flow-discover`.

Re-ran `tests/unit/test-skill-templates.sh` (43/43) and `tests/unit/test-openclaw-compat.sh` (32/32) after the change, confirmed checkout stays clean, and confirmed no file-mode changes.

---

PR #830, opened 2026-08-10
fix: release validator recognizes nested skill directories

### LINKED ISSUE #829: fix: release validator rejects registered nested skills
## Problem

`scripts/validate-release.sh 9.61.2` reports all four registered `skills/octopus-starter-pack/*/SKILL.md` entries as missing even though the files exist and are registered in `.claude-plugin/plugin.json`.

The validator builds `SKILL_FILES` with `find skills -mindepth 2 -maxdepth 2 -name SKILL.md`, which sees ordinary `skills/<skill>/SKILL.md` layouts but cannot see the supported nested `skills/octopus-starter-pack/<skill>/SKILL.md` layout. The registered-skill comparison therefore fails a valid release with four false errors.

## Expected

The release validator should compare recursive skill paths relative to `skills/`, while still ignoring unrelated files and requiring an exact `SKILL.md` registration match. Add a regression test covering both ordinary and nested skill directories.

## Reproduction

```bash
./scripts/validate-release.sh 9.61.2
```

Observed on current `main` after #828; the check exits 1 with four `Registered skill ... does not exist` errors.

nyldn:
## Description

`scripts/validate-release.sh` built its skill-registration check with `find "$ROOT_DIR/skills" -mindepth 2 -maxdepth 2 -name "SKILL.md"`, which only sees ordinary `skills/<name>/SKILL.md` layouts. The supported nested layout used by `skills/octopus-starter-pack/<name>/SKILL.md` sits one level deeper (depth 3), so all four of its registered, present skills (`debate-kickoff`, `council-verdicts`, `provider-health`, `model-cost-compare`) were reported as `does not exist` — four false errors on an otherwise valid release.

The same depth-2-only gap existed in the skill-frontmatter-format check (section 7), which built its file list from the glob `skills/*/SKILL.md`.

## Fix

Extracted the shared lookup into `find_skill_md_files()` (`find ... -mindepth 2 -maxdepth 3 -name "SKILL.md" -type f`) and reused it in both checks, so both ordinary and one-level-nested skill directories are recognized while still ignoring unrelated files.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (describe):

## Checklist
- [ ] OpenClaw registry in sync — not touched by this change
- [ ] New skills/commands registered in `.claude-plugin/plugin.json` — n/a, no skills added
- [ ] Version bump — n/a, bug fix only
- [ ] CHANGELOG.md updated — left for the release step per the daily-triage routine's protocol

## Testing

- New regression test `tests/unit/test-validate-release-skill-registration.sh` (3/3 pass): a fixture test pinning `find_skill_md_files()`'s ordinary + nested detection and ignoring of unrelated files, plus an end-to-end run of the real validator confirming the four `octopus-starter-pack` skills are no longer reported missing.
  - Verified the test **fails** against the pre-fix script, reproducing the exact error text from the issue (`ERROR: Registered skill 'octopus-starter-pack/council-verdicts' does not exist`, etc.) — confirms the test can catch the regression.
- `bash tests/unit/test-docs-sync.sh` — 141/141 pass (exercises skill/plugin frontmatter checks adjacent to this code).
- `bash tests/unit/test-openclaw-compat.sh` — 32/32 pass.
- `./scripts/validate-release.sh <current-version>` against the real repo: skill-registration section now reports `✓ All 63 skills properly registered` with zero errors (previously 4).
- Full `make test-unit` (247 suites): 243 passed, 4 failed. Confirmed all 4 failures (`test-execution-profile-dispatch.sh`, `test-feature-disclosure.sh`, `test-hud-smart-mode.sh`, `test-skill-doctor-stable-link.sh`) are pre-existing on unmodified `origin/main` and unrelated to this change — none reference `validate-release.sh`, `find_skill_md_files`, or `octopus-starter-pack`.
- `git diff origin/main...HEAD --summary | grep "mode change"` — empty, no mode changes.

```bash
bash tests/unit/test-validate-release-skill-registration.sh
bash tests/unit/test-docs-sync.sh
bash tests/unit/test-openclaw-compat.sh
```

## Related Issues
Closes #829

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01UrbGv2w4BrZeGUbrLwfYUE)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Release validation now recognizes skills in standard directories and one level of nested skill packs.
  * Nested registered skills are validated correctly without false missing-skill errors.

* **Tests**
  * Added regression coverage for nested skill discovery, unrelated file exclusion, and registration validation.

### DIFF
--- a/scripts/validate-release.sh
+++ b/scripts/validate-release.sh
@@ -414,7 +414,11 @@ echo "🏷️  Checking skill frontmatter format..."
 
 invalid_skill_names=0
 if command -v jq >/dev/null 2>&1 && jq -e '.skills[]? | select(startswith("./skills/"))' "$ROOT_DIR/.claude-plugin/plugin.json" >/dev/null 2>&1; then
-    mapfile -t SKILL_FRONTMATTER_FILES < <(find_skill_md_files "$ROOT_DIR/skills" | LC_ALL=C sort)
+    # Intentionally NOT find_skill_md_files: nested skills/octopus-starter-pack/*
+    # skills use bare frontmatter names (no skill-/flow-/octopus-/sys- prefix,
+    # namespaced by directory instead) and are out of scope for issue #829,
+    # which is only about the registration check below.
+    SKILL_FRONTMATTER_FILES=("$ROOT_DIR"/skills/*/SKILL.md)
 else
     mapfile -t SKILL_FRONTMATTER_FILES < <({
         find "$ROOT_DIR/.claude/skills" -maxdepth 1 -type f -name '*.md' -print 2>/dev/null
--- a/scripts/validate-release.sh
+++ b/scripts/validate-release.sh
@@ -366,8 +366,15 @@ echo ""
 # ============================================================================
 echo "🎯 Checking skill registration..."
 
+# Skills normally live at skills/<name>/SKILL.md (depth 2 below skills/), but
+# supported packs nest one level deeper at skills/<pack>/<name>/SKILL.md (see
+# skills/octopus-starter-pack/, issue #829) — search both depths.
+find_skill_md_files() {
+    find "$1" -mindepth 2 -maxdepth 3 -name "SKILL.md" -type f 2>/dev/null
+}
+
 if command -v jq >/dev/null 2>&1 && jq -e '.skills[]? | select(startswith("./skills/"))' "$ROOT_DIR/.claude-plugin/plugin.json" >/dev/null 2>&1; then
-    SKILL_FILES=$(find "$ROOT_DIR/skills" -mindepth 2 -maxdepth 2 -name "SKILL.md" -type f 2>/dev/null | sed "s|^$ROOT_DIR/skills/||;s|/SKILL.md$||" | sort)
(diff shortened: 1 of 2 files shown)


nyldn:
Thanks for the flag. Couldn't reproduce the exact `/dev/fd/63` error, but re-running `./scripts/validate-release.sh` to check turned up a real regression from the previous commit: reusing `find_skill_md_files()` for the section 7 frontmatter-format check made it discover the 4 `octopus-starter-pack/*` skills for the first time — and their frontmatter names (`council-verdicts`, `debate-kickoff`, etc.) are bare, without the `skill-`/`flow-`/`octopus-`/`sys-` prefix that check requires. So all four started failing a check they'd previously been invisible to.

That's out of scope for #829, which is only about the registration check (section 6) wrongly reporting those same skills as "does not exist." Fixed in af0865f by scoping the nested-skill lookup back to section 6 only, and reverting section 7 to its original depth-2-only glob. Verified `./scripts/validate-release.sh` now reports both "All 63 skills properly registered" and "All skill names use correct format" with zero errors.

---

PR #834, opened 2026-08-10
fix: make release validation recursive and monotonic

### LINKED ISSUE #829: fix: release validator rejects registered nested skills

### LINKED ISSUE #832: test: provider timeout regression uses non-monotonic wall clock
## Problem

`make ci-local` can fail `tests/unit/test-provider-auth-validity.sh` even when `run_with_timeout` correctly SIGKILLs the stubborn process (exit 137). The test measures elapsed time with two `date +%s` calls. If the host clock advances during the test, the duration becomes hundreds or thousands of seconds and the `<20s` assertion fails falsely.

Observed 2026-08-10 while validating #829: `exit=137, dur=626s`; the surrounding suite also displayed impossible wall-clock jumps. This is the same timing-flake class already removed from `test-agent-lifecycle-events.sh` in #828.

## Expected

Use Bash `SECONDS` for the elapsed-time assertion so the timeout regression measures monotonic process time and still proves the SIGKILL fallback completes within its bound. Add/retain mutation-capable coverage.

nyldn:
## Summary

- recursively discover `SKILL.md` registration paths, including supported nested bundles such as `skills/octopus-starter-pack/<skill>`
- move skill discovery into a directly tested release-validation helper
- make the SIGKILL timeout regression use Bash `SECONDS`, avoiding false failures when the host wall clock changes

Closes #829.
Closes #832.

## Verification

- `bash tests/unit/test-release-validation.sh` — 2/2
- `bash tests/unit/test-provider-auth-validity.sh` — 18/18, three consecutive runs
- `make sync-check`
- `OCTOPUS_NON_INTERACTIVE=1 make ci-local` — 247/247 unit suites, 16/16 smoke suites, 7/7 integration suites; all CI-only verifications green
- no mode changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release validation to reliably discover nested skill files.
  * Safely handles releases without a skills directory.
  * Improved timeout measurement for provider authentication validation.

* **Tests**
  * Added regression coverage for nested skill discovery, invalid directories, and validation behavior.
  * Confirmed timeout checks complete within the expected limit.

### DIFF
--- a/scripts/lib/release-validation.sh
+++ b/scripts/lib/release-validation.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+# Helpers for release manifest validation.
+
+octo_release_skill_files() {
+    local root="$1"
+    local skills_root="${root%/}/skills"
+
+    [[ -d "$skills_root" ]] || return 0
+
+    find "$skills_root" -mindepth 2 -type f -name 'SKILL.md' -print 2>/dev/null |
+        while IFS= read -r skill_path; do
+            skill_path="${skill_path#"$skills_root"/}"
+            printf '%s\n' "${skill_path%/SKILL.md}"
+        done |
+        sort
+}
--- a/scripts/validate-release.sh
+++ b/scripts/validate-release.sh
@@ -6,6 +6,8 @@ set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ROOT_DIR="$(dirname "$SCRIPT_DIR")"
+# shellcheck source=scripts/lib/release-validation.sh
+source "$SCRIPT_DIR/lib/release-validation.sh"
 
 RED='\033[0;31m'
 GREEN='\033[0;32m'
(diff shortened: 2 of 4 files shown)


nyldn:
Superseded by the narrower scheduled fixes #830 and #833. I reviewed both patches against this combined implementation: #830 has stronger end-to-end coverage of the real release validator, and #833 is behaviorally identical to the monotonic timing change here.

The local verification from this PR still applies to those exact changes: `OCTOPUS_NON_INTERACTIVE=1 make ci-local` completed green with 247/247 unit suites, 16/16 smoke suites, 7/7 integration suites, and all CI-only checks.

nyldn:
Heads up: this looks like it was written in parallel with #830 (same base commit, same target — `find_skill_md_files` in `scripts/validate-release.sh`), which the scheduled daily-triage routine merged a few minutes before this PR was opened. That's almost certainly why this now shows as `dirty`/conflicting against `main`.

#829 itself is already closed via #830's merge (registration check only, one-level-deep fix scoped to `skills/<pack>/<name>/SKILL.md`). This PR looks like it still carries independent value beyond that — arbitrary-depth recursion instead of one level, the `SECONDS`-based timeout fix, and closing #832 — so probably just needs a rebase onto current `main` rather than being closed outright. Flagging so the conflict doesn't look like a mystery.

---

PR #856, opened 2026-08-10
feat(updates): detect and repair stale plugin installs

### LINKED ISSUE #851: feat(update): detect stale installs and hand off safely to Claude's updater
## Problem

Issue #836 documents Claude Code's host-managed updater, but documentation alone does not prevent users from remaining on stale Octopus caches. This is especially painful when an old hook keeps firing errors before the user realizes the plugin itself is outdated.

Claude Code now supports background marketplace and installed-plugin updates after startup, but third-party marketplaces have auto-update disabled by default. A running session continues using the version loaded at launch and needs `/reload-plugins` or a restart after an update. The plugin therefore cannot safely replace its own loaded cache in place.

Official behavior: https://code.claude.com/docs/en/discover-plugins#configure-auto-updates

## Research

- Claude Code owns the versioned cache and provides `claude plugin marketplace update`, `claude plugin update`, and per-marketplace auto-update. Third-party auto-update is opt-in.
- Claude Code stores the local opt-in as `autoUpdate` for the known marketplace. Octopus may inspect this state, but should not rewrite Claude's settings or cache from a hook.
- claude-mem uses the same ownership boundary: Claude updates the plugin cache; a lightweight version/setup check detects a mismatch and prints an explicit repair/update action. It also exposes a user-invoked updater for non-Claude hosts rather than mutating the running plugin invisibly: https://github.com/thedotmack/claude-mem/blob/4702c337d85aa12e8ab7f845264a78885676261f/plugin/scripts/version-check.js

## Proposed durable design

1. Add a fast, local-only SessionStart/Setup staleness guard. It must never perform network I/O, invoke auth, update files in Claude's cache, or block startup.
2. Detect whether `nyldn-plugins` has host auto-update disabled. Emit a one-time actionable advisory (persisted with a cooldown) that gives the exact UI path to enable it.
3. Extend `/octo:doctor` with an update-health section: loaded plugin version, newest locally cached/catalog version, marketplace auto-update state, and reload-required state.
4. Add an explicit, user-invoked `doctor --fix` or update helper that delegates to Claude's official commands, verifies the resulting cached version, and clearly says restart or `/reload-plugins` is required. Never run this automatically from a hook.
5. Keep the marketplace/plugin version contract synchronized on every release and regression-test it.
6. For non-Claude hosts, report their native/manual update mechanism instead of assuming Claude's cache layout.

## Acceptance criteria

- [ ] No hook performs network I/O or invokes `claude plugin update`.
- [ ] Users with `nyldn-plugins` auto-update disabled receive a one-time/cooldown advisory with the exact opt-in path.
- [ ] `/octo:doctor` reports loaded version, cached/catalog version where locally discoverable, auto-update state, and reload requirement.
- [ ] A user-invoked repair path delegates to supported host update commands and verifies success.
- [ ] Tests cover disabled/enabled/missing/malformed marketplace state, stale cache, current cache, offline operation, and non-Claude hosts.
- [ ] Documentation explains the hard boundary: an already-stale version cannot receive new self-healing code until it is updated once.

Follow-up to #836.

nyldn:
## Summary

- add a cooldown-limited SessionStart advisory that reads only local Claude plugin metadata and never invokes a host CLI, package manager, network client, or authentication flow
- add doctor updates diagnostics for loaded, installed, catalog, and cache versions plus auto-update/reload state
- add an explicit host-aware update-plugin command for Claude Code and Codex; unsupported hosts fail closed with manual guidance
- document the unavoidable one-time manual-update boundary for installations older than this code and the required reload/restart lifecycle

## Design basis

Claude Code owns plugin cache mutation and third-party marketplace auto-update. The current session continues using its loaded version, so Octopus detects local state and delegates mutation instead of trying to self-modify from a startup hook.

- https://code.claude.com/docs/en/discover-plugins#configure-auto-updates
- https://code.claude.com/docs/en/plugin-marketplaces
- https://github.com/thedotmack/claude-mem/blob/4702c337d85aa12e8ab7f845264a78885676261f/plugin/scripts/version-check.js
- https://github.com/ohmyzsh/ohmyzsh/wiki/Settings#update-settings

## Verification

- focused updater health suite: 30/30 assertions under macOS stock Bash 3.2
- exact repository ShellCheck policy: pass
- make sync-check: pass
- OCTOPUS_NON_INTERACTIVE=1 make ci-local: 16 smoke suites, 249 unit suites, 7 integration suites passed
- no mode changes relative to main

Closes #851

## Summary by CodeRabbit

- **New Features**
  - Added host-aware plugin updates through the `update-plugin` command.
  - Added startup advisories for locally detected plugin updates without network activity.
  - Expanded diagnostics with update, companion, and cache checks.
  - Added command completions and help for update and diagnostic commands.

- **Documentation**
  - Documented host-specific update workflows, consent, reload requirements, and troubleshooting guidance.

- **Bug Fixes**
  - Improved host detection and handling of malformed metadata or unsupported update environments.

### DIFF
--- a/.claude/skills/skill-doctor/SKILL.md
+++ b/.claude/skills/skill-doctor/SKILL.md
@@ -91,6 +91,7 @@ If the user asks about a specific area, filter:
 
 ```bash
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor providers
+bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor companions
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor auth
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor config
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor updates
@@ -102,6 +103,7 @@ bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor skills
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor conflicts
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor agents
 bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor recurrence
+bash "${HOME}/.claude-octopus/plugin/scripts/orchestrate.sh" doctor cache
 ```
 
 ### Step 3: Check & Install Dependencies
@@ -345,7 +347,7 @@ Invoke this skill by asking Claude naturally, or run the orchestrator directly:
 
 | What to say / run | Action |
 |-------------------|--------|
-| "run Octopus doctor diagnostics" | Run all 11 categories |
+| "run Octopus doctor diagnostics" | Run all 14 categories |
 | "check Octopus providers" | Check provider installation only |
 | `bash "$OCTO_PLUGIN_ROOT/scripts/orchestrate.sh" doctor auth --verbose` | Detailed auth status |
 | `bash "$OCTO_PLUGIN_ROOT/scripts/orchestrate.sh" doctor --json` | Machine-readable output |
--- a/.claude/skills/skill-doctor/SKILL.md
+++ b/.claude/skills/skill-doctor/SKILL.md
@@ -20,7 +20,7 @@ trigger: |
 
 ## Overview
 
-Run environment diagnostics across 11 check categories. Identifies misconfigured providers, stale state, broken hooks, and other issues that prevent Claude Octopus from working correctly.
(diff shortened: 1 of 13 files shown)


nyldn:
Addressed all six actionable review comments in 9bd7266a:

- README uses the direct orchestrator doctor commands.
- Doctor warns and returns when `jq` is unavailable instead of claiming the plugin is current.
- Claude and Codex updates now verify installed/cache metadata against the expected local version after host commands.
- The updater suite now mutates fixture metadata and rejects a zero-exit unchanged-state update.
- Doctor docs consistently list all 14 categories, including `companions` and `cache`.
- The updater test now uses the shared framework and quoted environment arguments.

Focused suite: 37/37 passing under macOS Bash 3.2; repository ShellCheck policy, `make sync-check`, and `git diff --check` pass.

PR #906, opened 2026-08-13
fix: keep Octopus dormant until explicitly invoked

### LINKED ISSUE #898: perf(hooks): make Octopus activation explicit and fast when inactive
## Problem

Claude Octopus can engage without explicit user intent and pays global hook overhead even when no Octopus workflow is active.

Current evidence on `main`:

- `hooks/user-prompt-submit.sh` and `hooks/auto-router-inject.sh` default `OCTOPUS_AUTO_ROUTER_MODE` to `invoke`.
- Ordinary phrases such as "research options", "write tests", "code review", "threat model", and "should we" can inject a `Skill(skill: "octo:...")` recommendation without an explicit `/octo:*` command.
- Two UserPromptSubmit hooks classify ordinary prompts.
- Blanket PreToolUse/PostToolUse hooks spawn on common tool calls even when no Octopus workflow is active.
- Shipped command and skill frontmatter does not use Claude Code native `disable-model-invocation: true` control.
- SessionStart injects routing and discipline context globally.

This makes the plugin surprising and adds latency/context cost to unrelated Claude Code work.

## Desired behavior

1. Multi-provider or multi-agent execution starts only after explicit user intent:
   - an explicit `/octo:*` command;
   - an explicitly enabled opt-in router mode;
   - or continuation of an already-active Octopus workflow.
2. Fresh installs and existing users without an explicit preference default to no automatic invocation.
3. Explicit commands, aliases, active-workflow continuity, and safety protections keep working.
4. Ordinary prompts and ordinary tool calls take a cheap no-op path with no provider probes, state mutation, or injected Octopus persona.
5. Expensive or opinionated hooks are scoped to an active workflow/skill where Claude Code supports it.
6. Tests cover should-trigger and should-not-trigger prompts plus inactive/active hook paths.
7. Benchmarks record before/after local hook latency and enforce a reasonable regression ceiling.
8. Documentation explains the explicit default and opt-in modes.

## Research basis

Current Claude Code documentation says:

- use `disable-model-invocation: true` when a skill should be manual-only;
- skill frontmatter hooks are scoped to the component lifecycle;
- `if` filters prevent the hook process from spawning, avoiding that overhead;
- UserPromptSubmit runs once per turn and tool hooks run on every matching tool call.

Official docs:
- https://code.claude.com/docs/en/skills
- https://code.claude.com/docs/en/hooks

## Tracking

Beads is unavailable in this checkout because the remote-backed DB is on schema v49 with pending migrations to v65 and the `leases` table is absent. Per repository policy, no migration will be run; this GitHub issue and `AI_AGENT_HANDOFF.md` are the temporary tracking record.

nyldn:
## Summary

- make every Octopus command and skill native manual-only, and gate Octopus agents behind explicit workflows
- default routing, coaching, memory/context injection, remote autonomy, statusline mutation, work-queue checks, and PostToolUse behavior to off unless explicitly enabled or bound to the active host session
- use Claude Code handler-level `if` filters plus stale-client fast exits to avoid unrelated provider/quality hook work
- advance the stable plugin entrypoint to the host-loaded version so valid old caches cannot strand users on stale hooks

## Root cause

Installation registered broad prompt, tool, and session hooks. Plain-language routing defaulted to invoke, multiple hooks parsed every prompt/tool call, startup output instructed the model to act, and skills remained model-invocable. Shared workflow state was not tied to the current Claude session. Separately, `~/.claude-octopus/plugin` accepted any still-valid old cache target after an update.

## Performance

Before: common inactive hook paths averaged roughly 38–41 ms; an unrelated provider-validation path could block for about 18 seconds per call.

After: defense-in-depth inactive script paths measure roughly 6–14 ms. Claude Code 2.1.85+ skips the filtered process spawn entirely.

## Verification

- `make sync`
- `make sync-check`
- `git diff --check`
- fresh `make ci-local`: all smoke, unit, integration, and CI-only verification groups passed
- explicit activation regression: 18/18
- stale stable-entrypoint regression: 2/2

Closes #898

### DIFF
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -23,6 +23,11 @@ You get:    A structured comparison with three independent viewpoints,
 
 This works for research, escalated code review, debugging, TDD, escalated security audits, UI design, PRDs, and full build-to-ship workflows — 54 commands, 63 skills, 32 specialized personas.
 
+Octopus is dormant by default. Installing it does not route ordinary prompts or
+delegate to Octopus agents. Every command and skill is manual-only: use
+`/octo:*` when you want escalation. Plain-prompt routing is available only after
+you set `OCTOPUS_AUTO_ROUTER_MODE=suggest` or `invoke`.
+
 Multi-provider runs show an agent summary before synthesis, so failed, timed out, or oversize-rejected Codex, Antigravity, OpenRouter, and other perspectives are visible instead of being hidden behind a polished final answer.
 
 ## Install
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -29,7 +29,10 @@ delegate to Octopus agents. Every command and skill is manual-only: use
 you set `OCTOPUS_AUTO_ROUTER_MODE=suggest` or `invoke`. This legacy opt-in
 examines ordinary prompts and can route them into paid external-provider
 workflows; use `suggest` unless you intentionally want automatic invocation and
-are comfortable sharing the routed prompt context with configured providers.
+are comfortable sending the workflow prompt and any context that workflow
+explicitly gathers to configured providers. Provider-side retention follows
+those providers' account policies. Unset the variable (or set it to `off`) to
+opt out; direct `/octo:*` commands remain available.
 
 Multi-provider runs show an agent summary before synthesis, so failed, timed out, or oversize-rejected Codex, Antigravity, OpenRouter, and other perspectives are visible instead of being hidden behind a polished final answer.
 
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -26,7 +26,10 @@ This works for research, escalated code review, debugging, TDD, escalated securi
 Octopus is dormant by default. Installing it does not route ordinary prompts or
 delegate to Octopus agents. Every command and skill is manual-only: use
 `/octo:*` when you want escalation. Plain-prompt routing is available only after
-you set `OCTOPUS_AUTO_ROUTER_MODE=suggest` or `invoke`.
+you set `OCTOPUS_AUTO_ROUTER_MODE=suggest` or `invoke`. This legacy opt-in
+examines ordinary prompts and can route them into paid external-provider
+workflows; use `suggest` unless you intentionally want automatic invocation and
+are comfortable sharing the routed prompt context with configured providers.
 
 Multi-provider runs show an agent summary before synthesis, so failed, timed out, or oversize-rejected Codex, Antigravity, OpenRouter, and other perspectives are visible instead of being hidden behind a polished final answer.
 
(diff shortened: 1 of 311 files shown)


nyldn:
Addressed the automated review findings in 5793b10a:

- Added a closed route allowlist and explicit traversal/path rejection before `/octo:auto` loads a command.
- Defined the manual composition contract: read the complete source, use it as current-conversation instructions, preserve workflow checkpoints, and pass user text only as arguments.
- Standardized the stable model-facing plugin path versus the hook/runtime `CLAUDE_PLUGIN_ROOT` contract.
- Added privacy/cost guidance for the legacy `OCTOPUS_AUTO_ROUTER_MODE=invoke` opt-in.
- Added regression coverage for route-path validation, the composition contract, and positive `Skill()` callers.

Verification after the changes: `make ci-local`, `make sync-check`, focused explicit-activation/router suites, and `git diff --check` all pass.

nyldn:
Second-pass review disposition:

- **Allowlist completeness:** hardened in 7bfe7372. The regression now parses every route the smart router can emit and requires exact set equality with the closed allowlist. Commands such as `define` and `deliver` are valid direct commands but are not smart-router outputs; therefore rejecting them as an inferred route is intentional.
- **Manual skill loading / TDD behavior:** intentional breaking of automatic activation for #898. All positive internal `Skill()` callers were migrated; the test rejects any remaining positive caller, and the complete unit/integration suite passed.
- **Stable source trust:** the command reads only the host-loaded plugin through the self-healed stable symlink. The assembly standard now states this does not introduce a second plugin trust boundary. The target is validated/repaired by `ensure-plugin-root.sh`; arbitrary prompt text cannot select the path.
- **Prompt-only enforcement:** `/octo:auto` is itself a Claude command contract, so its literal mapping and allowlist are its enforcement surface. The opt-in hook has separate fixed-token routing and regression coverage.
- **Beads blocker:** pre-existing and recorded per repository instructions; no schema migration was performed.

Focused tests and `make sync-check` pass on 7bfe7372. The full `make ci-local` gate passed on the immediately preceding implementation head; this final commit changes only the contract documentation and its exact-set regression.

nyldn:
Third-pass follow-up in 022ea4e3:

- Removed the six remaining legacy/custom `invocation: human_only` keys. `disable-model-invocation: true` is now the single manual-only convention, and the historical hardening test enforces the native flag.
- Expanded opt-in routing consent: what is sent, provider-policy retention boundary, and `off`/unset opt-out behavior.
- The hard-coded allowlist finding is resolved by the exact-set regression in 7bfe7372, as the review recommended.
- `${HOME}/.claude-octopus/plugin` remains intentional for model-facing reads because `CLAUDE_PLUGIN_ROOT` is not available in model tool-call context; SessionStart verifies/repairs it to the host-loaded plugin root. This distinction and trust boundary are now documented and tested.

Focused native-invocation, explicit-activation, env-var-effects, sync, and diff checks pass. The full local CI gate passed on the implementation head.

PR #920, opened 2026-08-14
feat(council): configurable chair-synthesis timeout

klass-723:
## Context

Follow-up to #919 (always emit `summary.json`). Same report: `orchestrate.sh council` on macOS with a codex seat routed through the chatgpt.com MCP transport. The chair-synthesis step reads every member artifact and writes the final structured document, so it needs more room than a single advice seat — and on that slow chair path the plain per-seat cap can expire mid-write, killing synthesis right at the boundary.

Today the synthesis dispatch just reuses `council_seat_timeout <chair>`, so there is no way to give synthesis more time without also loosening every advice seat.

## Change

- New `council_synthesis_timeout()` resolves `OCTOPUS_COUNCIL_SYNTHESIS_TIMEOUT` first, then falls back to the chair provider's normal per-seat resolution (`OCTOPUS_COUNCIL_TIMEOUT_<PROVIDER>` → `--seat-timeout` → legacy global → default).
- `council_live_response` applies it **only** to the `chair-synthesis` phase; advice/critique/revision seats keep the existing per-seat cap.

Backward-compatible: unset ⇒ identical behavior to today.

## Tests

Two unit regressions in `test-council-command.sh`:
- `council_synthesis_timeout_overrides_seat_cap` — override wins; unset falls back to per-seat resolution (incl. per-provider tuning); invalid (`0`) falls through instead of disabling the cap.
- `council_live_response_uses_synthesis_timeout_for_chair` — captures the timeout arg handed to `run_agent_sync_consultative`: `chair-synthesis` gets the synthesis value, `independent-advice` keeps the per-seat value.

`tests/unit/test-council-command.sh`: **75 passed, 0 failed, 1 skipped** (pre-existing macOS-CI pty flake). No file-mode changes.

## Relation to the other follow-ups

- The "auto-retry spawns more stalling run dirs" signature is **not** an internal `orchestrate.sh` retry (the `council)` route just calls `council_run`); it was the caller re-invoking. #919 already makes each run emit a machine-detectable `summary.json`, so the caller stops blind-retrying.
- A separate change will classify a seat killed by its own timeout monitor as `timed-out` (not generic `no-response`) so a codex exit 137 reads as "hit the cap, raise `OCTOPUS_COUNCIL_TIMEOUT_CODEX`" rather than a mysterious SIGKILL/OOM.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated timeout setting for AI council chair synthesis.
  * Synthesis uses the configured override when available, with provider-specific and default timeout fallbacks.

* **Documentation**
  * Documented the new `OCTOPUS_COUNCIL_SYNTHESIS_TIMEOUT` setting in the unreleased changelog.

* **Bug Fixes**
  * Improved timeout handling so chair synthesis can be bounded independently from regular council responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## [Unreleased]
 
+### Added
+
+- `OCTOPUS_COUNCIL_SYNTHESIS_TIMEOUT` bounds the chair-synthesis dispatch
+  independently of the per-seat cap. Synthesis reads every member artifact and
+  writes the final structured document, so it routinely needs more room than a
+  single advice seat — and on a slow chair path (e.g. codex via the
+  chatgpt.com MCP transport) the plain seat cap can expire mid-write. When
+  unset, synthesis keeps falling back to the chair provider's normal per-seat
+  resolution, so existing tuning (`OCTOPUS_COUNCIL_TIMEOUT_<PROVIDER>`,
+  `--seat-timeout`) still applies.
+
 ## [9.64.0] - 2026-08-13
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,12 @@
   independently of the per-seat cap. Synthesis reads every member artifact and
   writes the final structured document, so it routinely needs more room than a
   single advice seat — and on a slow chair path (e.g. codex via the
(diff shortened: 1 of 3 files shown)


nyldn:
Landing this onto main via #926 (rebased onto #919; CHANGELOG conflict resolved). This PR can close once #926 merges.

PR #922, opened 2026-08-15
fix(council): host-native chair satisfies quorum (met from vendor approvals, not chair receipt)

klass-723:
## Problem (highest-value ask from the latest council report)

`summary.json` reports `quorum.met: false` while `distinct_approving_providers: 2` and both non-chair vendor bodies are real APPROVEs. Isolated by diffing a `met:true` run against a `met:false` one: the difference is `chair_received: false`. When the chair **is** the host runtime (Claude Code running the council), it cannot self-dispatch, so it never produces a substantive chair response file — yet it is present and synthesizes the council in-context. `met` was gated on the chair response, so every such run had to be overridden manually per "per-seat read is authoritative."

## Fix

Chair presence = **a dispatched chair response OR a host-native chair**:

- `council_chair_is_host_native()` resolves the chair seat's provider and checks its status in `COUNCIL_PROVIDER_STATUS_JSON` (`host-native`).
- The `met` computation uses `chair_present` instead of `chair_received` directly. `met` now reflects vendor approvals + a present, synthesis-capable chair.
- `summary.json` gains `quorum.chair_host_native` so `met` is recomputable from the artifact: `met = received_non_chair >= required AND (required < 2 OR distinct_approving_providers >= required) AND (chair_received OR chair_host_native)`.

No loosening of the real shortage gate: a chair that is merely **unavailable** (not host-native) still fails quorum, and the distinct-approving-vendor guard (#1992/#1993/#670) is untouched.

## Tests

- `council_chair_is_host_native_detects_status` — helper returns true for a `host-native` chair provider, false for `available`.
- `council_quorum_met_with_host_native_chair` — advice-phase run: host-native chair + two approving vendors → `met=true` with `chair_received=false` and `chair_host_native=true`; **negative control**: same approvals but an available-but-absent chair → `met=false`.

`tests/unit/test-council-command.sh`: **75 passed, 0 failed, 1 skipped** (pre-existing macOS-CI pty flake). No file-mode changes.

## Relation to the report's other asks

- **#2 synthesis watchdog / no summary.json** → #919 (always emit a valid `summary.json`).
- **#4/#6 codex 137 at the cap** → #921 (classify `timed-out`, name the knob) + #920 (`OCTOPUS_COUNCIL_SYNTHESIS_TIMEOUT`).
- **#3 auto-retry spawning run dirs** → not an internal `orchestrate.sh` retry; #919 gives the caller a machine-detectable failure to stop on.
[...]

PR #921, opened 2026-08-14
fix(council): classify seats killed by their own timeout as timed-out

klass-723:
## Context

Follow-up to #919 / #920. Same report: on macOS a council codex seat routed through the chatgpt.com MCP transport dies with **exit 137** at the ~5-min per-seat cap. The reporter read that as a SIGKILL/OOM ("codex seat SIGKILLed on MCP timeout"). It isn't the OS — it's **our own watchdog**: `run_with_timeout` kills an over-cap seat with the coreutils convention (124) or, on the macOS bash fallback, SIGTERM→SIGKILL (143/137). That code reaches the advice phase via the detached `.done` sentinel, but a timed-out seat with no usable response was lumped into a generic `no-response`, so nothing distinguished "hit the cap" from "provider returned nothing."

## Change

- `council_rc_is_timeout()` — recognizes the watchdog kill codes (124/137/143).
- In the advice phase, a seat with no usable response whose dispatch rc is a timeout is recorded as **`timed-out`** in `summary.json` (instead of `no-response`).
- `council_note_seat_timeout()` accumulates an actionable end-of-run warning naming the exact knob to raise: `Council warning: seat codex (code-reviewer) exceeded its 300s cap (rc=137) — raise OCTOPUS_COUNCIL_TIMEOUT_CODEX to give it more time`.

Quorum is untouched: `timed-out` is not `responded`, so it still can't count as an approver, and the existing salvage path still keeps a timed-out-**but-complete** review as a real response.

## Tests

- `council_rc_is_timeout_recognizes_cap_kill_codes` — 124/137/143 match; 0/1/2/126/127 don't.
- `council_advice_marks_timed_out_seat` — a stubbed 137 dispatch with no output yields seat status `timed-out` plus the `OCTOPUS_COUNCIL_TIMEOUT_CODEX` / `300s` hint.

`tests/unit/test-council-command.sh`: **75 passed, 0 failed, 1 skipped** (pre-existing macOS-CI pty flake). No file-mode changes.

## Report follow-up status

This is the last of the report's asks that needs code. For the record:
- **#1 synthesis hang / no summary** → #919 (merged path: always emit `summary.json`).
- **#5 synthesis timeout config** → #920 (`OCTOPUS_COUNCIL_SYNTHESIS_TIMEOUT`).
[...]

PR #923, opened 2026-08-15
fix(agy): fail open when OCTOPUS_AGY_MODEL cannot be validated (don't crash on an unreachable catalog)

klass-723:
## Problem (from the latest council report, ask #5)

Under a restricted sandbox the runner crashed on spawn with **`Cannot validate OCTOPUS_AGY_MODEL`**. `validate_agy_model_name` conflates two very different situations:

- **Cannot validate** — the agy CLI is missing, or `agy models` returns an empty catalog (offline, or a sandbox blocked the call). We don't *know* whether the pin is valid.
- **Invalid** — the catalog is reachable and the model is definitively absent from it.

Both returned `1`, so an unreachable catalog aborted `OCTOPUS_AGY_MODEL` resolution (`model-resolver.sh:189-191` → `return 1`) and crashed the council even when the pin was perfectly valid.

## Fix

- **Fail open** on "cannot validate": warn and trust the pin. A genuinely bad model is still caught by agy itself at dispatch, with a clear error — far better than a preflight abort.
- **Keep fail-closed** for a model absent from a *reachable* catalog (unchanged, still prints the available list).
- `OCTOPUS_AGY_MODEL_STRICT=1` restores the old strict behavior for anyone who wants a hard preflight failure.

## Tests

- `agy pin fails open when the catalog is unreachable (sandboxed/offline)` — empty-catalog stub → validation returns 0.
- `OCTOPUS_AGY_MODEL_STRICT=1 restores fail-closed on an unreachable catalog` → returns 1.
- Existing `agy model pin still rejects unknown labels` (reachable catalog, absent model) stays fail-closed.
- Updated `agy model validation bounds a stalled catalog lookup`: still asserts the timeout **bound** (lookup killed at 1s, never reaches the mock's 3s sleep), now with fail-open rc — its old `rc≠0` encoded the pre-fix semantics.

`test-council-model-selection-fixes.sh`: **12 passed / 0 failed**. `test-agy-provider.sh`: **50 passed / 0 failed**. No file-mode changes.

## Report follow-ups

This is ask #5 (`OCTOPUS_AGY_MODEL` validation). Quorum (#1) → #922; synthesis/summary (#2) → #919; synthesis timeout (#5) → #920; codex 137 (#4/#6) → #921. Remaining new: background-dispatch pollable status (#3) — separate PR. The `mkstemp … Operation not permitted` is largely the agy child's own temp-file creation under sandbox (mitigated by the leading-token runner-invocation guidance from #790); our own `mktemp -t` already honors `$TMPDIR`.

## Summary by CodeRabbit

* **Bug Fixes**
  * AGY model validation now succeeds when the model catalog is unavailable or empty.
  * Models absent from a reachable catalog continue to be rejected.
  * Strict validation mode remains available to enforce fail-closed behavior.
  * Catalog lookups now remain time-bounded, including during timeouts.

* **Tests**
  * Added coverage for unavailable catalogs, timeout handling, and strict validation behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `OCTOPUS_AGY_MODEL` validation no longer aborts a run when the agy catalog is
+  unreachable. A restricted sandbox or an offline host makes `agy models` return
+  an empty catalog; the validator treated that "cannot validate" case like a
+  definitively invalid pin and failed model resolution, crashing the council on
+  spawn even when the pin was valid. Validation now fails **open** when the
+  catalog is unreachable (agy still rejects a genuinely bad model at dispatch,
+  with a clear error) and only fails **closed** for a model that is absent from a
+  *reachable* catalog. Set `OCTOPUS_AGY_MODEL_STRICT=1` to require validation and
+  restore the previous fail-closed behavior.
+
 ## [9.64.0] - 2026-08-13
 
 ### Changed
--- a/scripts/lib/model-resolver.sh
+++ b/scripts/lib/model-resolver.sh
@@ -124,9 +124,22 @@ validate_agy_model_name() {
             ;;
     esac
 
+    # "Cannot validate" (catalog unreachable) is NOT the same as "invalid" (a
+    # model definitively absent from a reachable catalog). Treating them alike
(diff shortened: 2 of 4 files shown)


nyldn:
Landing this onto main via #926 (rebased onto #919; CHANGELOG conflict resolved). This PR can close once #926 merges.

nyldn:
Landed on main via #926.

Issue #900, opened 2026-08-13
Killing orchestrate.sh develop leaves codex seats alive with danger-full-access, still writing to the user's repo


lunatupini:
### Plugin Version

9.63.0

### Claude Code Version

2.1.229

### What happened?

`pkill -f "orchestrate.sh develop"` terminated every orchestrator shell, but the provider seats it had spawned survived as orphans and continued mutating the working tree for ~25 minutes afterwards.

Timeline on a working repository:

- 17:52 — run diagnosed as hung, `pkill -f "orchestrate.sh develop"` issued.
- 17:53:13 — three thread wrappers logged `exited without completion marker ... waiting up to 180s`.
- 17:56 — SIX NEW `codex exec` processes appear, each wrapped in a fresh `timeout -k 10 2400` (40-minute budget), `--sandbox danger-full-access`.
- 17:57:24 — an orphan rewrites an untracked 30KB markdown plan at the repo root.
- 17:57:56 — an orphan rewrites `src/lib/<module>.ts` and `src/lib/<module>.test.ts`.
- 17:59 — orchestrator count is 0, but `ps -eo pid,cmd | grep "[c]odex exec"` still returns 6 live seats.

They had to be killed by explicit PID. Two consequences:

1. Files changed underneath an in-flight verification run, so a test / build gate that had just reported green no longer described the tree. The gate had to be re-run from scratch.
2. An `Edit` against the plan file failed with "File has been modified since read" because an orphan had rewritten it mid-edit.

`--sandbox danger-full-access` with no live parent means these processes can write anywhere in the repo with nothing left to collect, review, or attribute their output.

### What did you expect?

Killing the orchestrator should terminate its whole process group, or the seats should detect a dead parent and exit before writing.

At minimum, a seat should not be spawned with a fresh 2400s budget at the moment its parent is being torn down, and `danger-full-access` seats should hold a liveness check against the collector that is supposed to consume their output.

### Steps to reproduce

Host preconditions: codex CLI present (<codex-model> reachable); agy present; opencode/copilot/qwen/cursor-agent/grok/perplexity/openrouter ABSENT; ollama present; git repo with an untracked file at root.

1. cd into a git repository that has at least one untracked file at the root.
2. export OCTOPUS_TANGLE_REQUIRE_CLEAN_BASELINE=false
3. export OCTOPUS_TANGLE_RUN_WORKTREE=false
4. export OCTOPUS_AGENT_TIMEOUT=3600
[...]

PR #929, opened 2026-08-16
feat(setup): completing setup opts into suggest-mode routing

nyldn:
## Why

Octopus is dormant on install (#898). That correctly stops unrequested paid dispatch, but `/octo:setup` and dormancy were orthogonal: a user who ran setup and authenticated providers still got nothing until they memorized command names. Setup completion is an explicit act and should count as consent to *suggestions*.

## What

Completing `/octo:setup` persists `auto_router_mode=suggest`.

- Octopus names a matching command for a plain prompt. It never dispatches a provider on its own.
- `invoke` is never written automatically; it stays opt-in behind `OCTOPUS_AUTO_ROUTER_MODE=invoke`.
- A profile that never runs setup stays fully dormant. The #898 contract is unchanged for anyone who does not opt in.
- An existing `auto_router_mode` value is never overwritten, so a prior opt-out survives a re-run of setup.

## Design note

Setup writes `~/.claude-octopus/preferences.json`, the file the prompt hooks already read, rather than `user-config.json` which setup uses for its other choices. Splitting it this way avoids adding a second file read to the `UserPromptSubmit` path that #898 measured at 38-41 ms per inactive hook. The write is atomic at `0600`, preserves unrelated keys, and leaves a corrupt preferences file untouched rather than replacing it.

## Evidence

New regression `tests/unit/test-setup-suggest-routing.sh` went red first at 6/12 failing, with the 6 passing cases being the guard conditions (dormant default, opt-out preserved, invoke never written) that prove the suite can distinguish. It now passes 12/12.

Adjacent contracts re-run green: explicit activation 20/20, auto-router hooks 14/14, proactive suggestions 12/12.

Full local gate `make ci-changed` failed closed to the complete matrix (shared orchestrator plus generator changes) and passed: 16/16 smoke, 274/274 unit, 7/7 integration, plus the CI-only verifications. `make sync-check` clean, ShellCheck error-level clean, `git diff --check` clean, no file mode changes.

Refs: oco-9yj

## Summary by CodeRabbit

* **New Features**
  * Setup now enables routing suggestions by default without automatically invoking providers.
  * Existing preferences and explicit opt-outs are preserved.
  * Users can disable suggestions or opt in to automatic invocation through configuration.
  * Profiles that have never completed setup remain dormant.

* **Bug Fixes**
  * Improved preference handling for invalid or inaccessible configuration files while preserving settings and secure file permissions.

* **Documentation**
  * Updated setup guidance, command references, README instructions, and changelog to explain routing modes.

### DIFF
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -27,9 +27,13 @@ Octopus is dormant on install. Installing it does not route ordinary prompts or
 delegate to Octopus agents. Every command and skill is manual-only: use
 `/octo:*` when you want escalation. Completing `/octo:setup` turns on
 suggestions (`auto_router_mode=suggest`), which name a matching command without
-dispatching any provider; a profile that never ran setup stays fully dormant,
-and an existing preference is never overwritten. Automatic invocation stays
-opt-in behind `OCTOPUS_AUTO_ROUTER_MODE=invoke`. That opt-in
+dispatching any provider; by default a profile that never ran setup stays fully
+dormant, and an existing preference is never overwritten. Either state can be
+overridden explicitly: `export OCTOPUS_AUTO_ROUTER_MODE=off|suggest|invoke` for
+the current shell, or `"auto_router_mode": "off"` in
+`~/.claude-octopus/preferences.json` to persist it. The environment variable
+wins over the stored preference. Automatic invocation stays opt-in behind
+`OCTOPUS_AUTO_ROUTER_MODE=invoke`. That opt-in
 examines ordinary prompts and can route them into paid external-provider
 workflows; use `suggest` unless you intentionally want automatic invocation and
 are comfortable sending the workflow prompt and any context that workflow
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -29,10 +29,10 @@ delegate to Octopus agents. Every command and skill is manual-only: use
 suggestions (`auto_router_mode=suggest`), which name a matching command without
 dispatching any provider; by default a profile that never ran setup stays fully
 dormant, and an existing preference is never overwritten. Either state can be
-overridden explicitly: `export OCTOPUS_AUTO_ROUTER_MODE=off|suggest|invoke` for
-the current shell, or `"auto_router_mode": "off"` in
-`~/.claude-octopus/preferences.json` to persist it. The environment variable
-wins over the stored preference. Automatic invocation stays opt-in behind
+overridden explicitly: `export OCTOPUS_AUTO_ROUTER_MODE=off` for the current
+shell (accepted values are `off`, `suggest`, and `invoke`), or
+`"auto_router_mode": "off"` in `~/.claude-octopus/preferences.json` to persist
+it. The environment variable wins over the stored preference. Automatic invocation stays opt-in behind
 `OCTOPUS_AUTO_ROUTER_MODE=invoke`. That opt-in
 examines ordinary prompts and can route them into paid external-provider
 workflows; use `suggest` unless you intentionally want automatic invocation and
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -23,10 +23,13 @@ You get:    A structured comparison with three independent viewpoints,
 
 This works for research, escalated code review, debugging, TDD, escalated security audits, UI design, PRDs, and full build-to-ship workflows — 54 commands, 63 skills, 32 specialized personas.
 
-Octopus is dormant by default. Installing it does not route ordinary prompts or
+Octopus is dormant on install. Installing it does not route ordinary prompts or
 delegate to Octopus agents. Every command and skill is manual-only: use
-`/octo:*` when you want escalation. Plain-prompt routing is available only after
-you set `OCTOPUS_AUTO_ROUTER_MODE=suggest` or `invoke`. This legacy opt-in
+`/octo:*` when you want escalation. Completing `/octo:setup` turns on
+suggestions (`auto_router_mode=suggest`), which name a matching command without
+dispatching any provider; a profile that never ran setup stays fully dormant,
+and an existing preference is never overwritten. Automatic invocation stays
+opt-in behind `OCTOPUS_AUTO_ROUTER_MODE=invoke`. That opt-in
 examines ordinary prompts and can route them into paid external-provider
 workflows; use `suggest` unless you intentionally want automatic invocation and
 are comfortable sending the workflow prompt and any context that workflow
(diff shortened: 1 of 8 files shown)


nyldn:
Thanks — four of five applied, one rejected with reasons. Each was reproduced against the code before I touched anything.

**Applied**

**Serialize the read-check-write sequence** (Major, `user-config.sh:97`). Correct, and the most important of the five. The window let a concurrent writer set `auto_router_mode=off` between the `has()` check and the rename, which would silently clobber the opt-out this function promises never to touch. Now wrapped in a `mkdir`-based lock matching the existing portable pattern in `scripts/lib/events.sh` (`flock` is Linux-only; `mkdir` is atomic on every POSIX filesystem), with a bounded ~1s spin, a `RETURN` trap so an interrupt mid-write still releases, and skip-on-contention rather than force. Skipping is the safe side: an unset preference leaves routing dormant.

I verified this is a real red-to-green rather than a test that could never fail. Against the pre-fix lib with a lock directory held, `octo_pref_write_default` wrote `suggest` anyway; against the fixed lib it correctly writes nothing.

**Add a corrupt-preferences regression case** (`test:87`). Fair catch — the handling existed but nothing exercised it. Two cases added: the corrupt file is byte-for-byte unchanged after the call, and the hook still resolves `off`. To be precise, this closes a coverage gap in already-working behavior rather than fixing a defect.

**Document dormant behavior and overrides consistently** (`.claude-plugin/README.md:32`). Applied. It now qualifies the dormancy claim with "by default", documents both override forms, and states which one wins.

**Whole-argument env assignments** (`test:37`). Applied, `env "HOME=$h" "CLAUDE_PLUGIN_ROOT=$p"`. This matches the repo rule in CLAUDE.md, which exists because the expert-review secret scanner false-positives on the prefix form.

**Rejected**

**Use the project logger for diagnostics** (`user-config.sh:67`). Not applying. `user-config.sh` is a standalone sourced library that deliberately sources nothing — its header explains that setting shell options here would leak into the caller's shell and persist after the file returns. There is no `log` function in its scope, and pulling in `logging.sh` would impose that dependency on every caller that sources it, including the `setup.md` bash block. The file's own established convention is raw `echo ... >&2`: both pre-existing warnings in `octo_config_write` (lines 21 and 32) use it, and the new code matches them. The `.coderabbit.yaml` rule is right for executable scripts under `scripts/`; it does not fit a no-dependency sourced lib. Happy to revisit if a leak-free logger shim for sourced libraries lands.

Suite is now 17/17. Full local gate re-running.

nyldn:
Both second-round findings applied — and the Major one was a defect I introduced in the previous commit, so thank you for catching it.

**Drop the RETURN trap** (Major, `user-config.sh:103`). Confirmed by reproduction before changing anything. Against the previous commit, under `set -T` plus `set -u`:

```
with-RETURN-trap: exit=127 | _: line 4: prefs_file: unbound variable
no-trap:          exit=0   | CALLER_SURVIVED
```

The trap stayed armed past the function, then expanded the out-of-scope `prefs_file` on the next return and killed the caller. This also violates a rule this repo already learned the hard way: a sourced library must never install traps that replace the caller's (see the EXIT/INT/TERM trap work in #900/#902).

Fixed by removing the trap entirely rather than saving and restoring it. The locked body moved into `_octo_pref_write_locked`, so `octo_pref_write_default` keeps exactly one unlock path and arms nothing. Save-and-restore would have worked but leaves a sourced lib mutating caller trap state, which is the pattern worth avoiding. Subshell isolation was the other option you offered; I skipped it because the body needs to report a distinct exit status and a subshell would have cost a fork on every setup run for no benefit here.

Residual, stated rather than hidden: an interrupt landing between lock and unlock leaves a stale lock directory. The bounded spin in `_octo_pref_lock` degrades that to "skip the write", which leaves routing dormant. That matches the existing behavior of `scripts/lib/events.sh` and is the safe direction.

Regression added: the helper runs under `set -T -u` and the caller must still reach `CALLER_SURVIVED`.

**Invalid shell example** (Minor, `.claude-plugin/README.md:34`). Correct, `export OCTOPUS_AUTO_ROUTER_MODE=off|suggest|invoke` parses as a pipeline and would silently not set the variable. Now a single valid `export ...=off` with the accepted values listed in prose.

Suite 17/17 to 18/18, green natively and under a GNU-stat emulation that reproduces the earlier Linux-only `stat` failure.

nyldn:
Both round-three findings applied. Both were real, and one of them exposed a bad verification habit on my side worth naming.

**Do not replace an unreadable preferences file** (Major, `user-config.sh:93`). Confirmed by reproduction. Against the previous commit, a preferences file containing `keep_me=yes` made unreadable came back **empty** after the call: the `|| current="{}"` fallback let the write replace it and discard every key. A failed read now warns and returns without writing.

**Release the lock under `set -e`** (Major, `user-config.sh:150-152`). Also confirmed, with an important correction to my own first attempt at checking it.

My initial probe called the helper as `octo_pref_write_default ... || true`. That `|| true` is exactly the condition context that suppresses errexit, so the probe reported "caller survived, lock not stranded" against code that was in fact broken. I nearly dismissed the finding on that basis. Re-probed with a bare call, matching how `commands/setup.md` actually invokes it:

```
PRE-FIX -> caller=[]                 lock_stranded=[YES]
FIXED   -> caller=[CALLER_SURVIVED]  lock_stranded=[no]
```

Pre-fix the caller died before reaching its next statement and the lock directory was left behind. Two changes rather than one:

1. The inner call is evaluated inside an `if`, so a nonzero return cannot trip errexit before `_octo_pref_unlock` runs.
2. The write-failure path now warns and returns 0 instead of 1. Your finding covered the stranded lock; fixing only that still left the caller dying, because a bare call returning nonzero under `set -e` kills the shell regardless of what the function cleaned up first. Returning 0 is also the consistent choice: every other failure path in this function already does (missing `jq`, corrupt file, busy lock, unreadable file), as does the sibling `octo_config_write`. Persisting an optional preference should never abort someone's setup run.

The committed regression calls the helper bare, with a comment explaining that adding `|| true` would make the assertion vacuous.

Suite 18/18 to 22/22, green natively and under a GNU-stat emulation reproducing the earlier Linux-only failure.

nyldn:
Applied. The CHANGELOG said a profile that never runs setup "stays fully dormant", which overstates it: `OCTOPUS_AUTO_ROUTER_MODE` and a stored `auto_router_mode` preference each enable routing on their own with no setup involved. Now scoped to "absent an explicit override", matching the wording already used in `.claude-plugin/README.md`.

This is the same accuracy point you raised on the README in round one, and I should have carried the correction across both files then rather than fixing only the file you flagged.

PR #940, opened 2026-08-18
fix: restore broken star history chart in README

FaintFlower:
The star history chart in the README is currently broken and shows nothing. The service it relied on stopped working after GitHub's stargazer API restrictions, so a fix is necessary to restore the chart.

This updates the chart to a working star history service. Only the chart is affected; any badges are left unchanged.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Star History Chart image and links to use the new destination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,7 +664,7 @@ Run commands with the `--verbose` flag to get detailed debugging output. Logs ar
 
 Join [r/ClaudeOctopus](https://www.reddit.com/r/ClaudeOctopus/) for help, workflow tips, showcases, and updates.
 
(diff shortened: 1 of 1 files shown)


nyldn:
Thanks for reporting the broken chart. The underlying breakage is confirmed, but this patch replaces the official Star History service with `star-history.dera.page`, and we could not establish ownership, maintenance, or a repository-level trust relationship for that host.

GitHub's July 2026 stargazer restrictions are the root cause. Star History's current supported path for a repository owner is an official `star-history.com` embed backed by a sealed, least-privilege token:

- https://www.star-history.com/blog/github-stargazer-api-restriction/
- https://www.star-history.com/blog/how-to-use-github-star-history/

We are closing this version rather than add an unaffiliated README dependency. A replacement PR that keeps the official host and uses the documented sealed-token flow would be welcome.

PR #973, opened 2026-08-28
fix(council): isolate the councils pool per session (stop cross-session dir/diagnostic collisions)

klass-723:
## Problem

When several governed Claude Code sessions run concurrently in sibling worktrees on one machine, they all write council runs into a single shared `~/.claude-octopus/councils/` pool. Reported symptoms: a sibling session's runs appear as "the newest run", its `run-status.json` `pid` misleads this session's diagnostics (a killed session inspected the newest dir and found a *foreign* worktree's pid), and foreign/duplicate run directories collide (the lead had to content-grep to find its own).

## Root-cause note (important)

Code review found **no cross-session process reaping**: every kill in the runner (`heartbeat.sh`, `council.sh`) is pid-scoped (`kill -P <cmd_pid>`, tree-kill of a specific seat) — there is no broad `pkill -f`/`killall` and no cleanup that scans the pool and reaps by another dir's `run-status.json` pid. So the "foreign pid" is a **diagnostic artifact of the shared pool**, not evidence that a sibling killed this session. The reported `rc 143` (SIGTERM) kills are most plausibly the **host Bash-tool timeout** on the long (~17 min) council call, which fires more under contention — a separate concern this PR does not claim to fix.

What this PR fixes is the confirmed shared-state problem: dir/diagnostic collisions.

## Change

Namespace the **default** pool by session:

- `council_session_slug()` — filesystem-safe key: Claude Code session id → worktree/cwd basename → pid.
- Default pool becomes `…/councils/session-<slug>/…`, so each session owns its runs and "the newest dir" is always its own.
- An explicit `--output-dir` (`COUNCIL_OUTPUT_DIR`) is **honored unchanged**.
- `OCTOPUS_COUNCIL_SHARED_POOL=1` restores the flat shared pool.

Preserves concurrency; no serialization.

## Tests

`council_per_session_pool_isolation` — two sessions get separate pools (`session-sessA` vs `session-sessB`), the opt-out restores the flat pool (no `session-` segment), and an explicit `--output-dir` is unaffected. Slug sanitization checked.

`tests/unit/test-council-command.sh`: **87 passed, 0 failed, 1 skipped** (pre-existing macOS-CI pty flake). No file-mode changes.

## Not in scope (from the report)

- #2 "never reap by broad match" — N/A; no broad-match kill exists in the code.
- #3 resume-a-killed-seat-from-`.partial` — would reduce the *cost* of a kill; happy to follow up if you want it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Council runs are now isolated by session by default, preventing runs from different sessions from sharing storage.
  * Session-based storage names are safely formatted for filesystem use.
  * Explicit output directory settings continue to work unchanged.
  * Shared storage remains available with `OCTOPUS_COUNCIL_SHARED_POOL=1`.

* **Tests**
  * Added coverage for session isolation, shared storage, safe naming, and custom output directories.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@
   a sibling session's runs appeared as "the newest run", its `run-status.json`
   misled this session's diagnostics, and foreign/duplicate run directories
   collided. The default pool is now namespaced by session
-  (`councils/session-<id>/…`, keyed on the Claude Code session id, falling back to
-  the worktree basename then the pid), so each session owns its runs and the
-  newest directory is always its own. An explicit `--output-dir` is honored
+  (`councils/session-<id>/…`, keyed on the Claude Code or Codex session id, with
+  Codex task id, worktree basename, and pid fallbacks), so each session owns its
+  runs and the newest directory is always its own. An explicit `--output-dir` is honored
   unchanged; set `OCTOPUS_COUNCIL_SHARED_POOL=1` to restore the flat shared pool.
 
 ## [10.1.0] - 2026-08-30
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,10 @@
   misled this session's diagnostics, and foreign/duplicate run directories
   collided. The default pool is now namespaced by session
   (`councils/session-<id>/…`, keyed on the Claude Code or Codex session id, with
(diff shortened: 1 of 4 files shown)


nyldn:
@coderabbitai review

PR #975, opened 2026-08-29
fix(tangle): enforce decomposition adequacy and write scopes

Jhacarreiro:
## Summary

This hardens Tangle decomposition and execution around deliverable adequacy and write-scope discipline, without adding project-specific behavior.

- feed design-review synthesis into decomposition as guidance while keeping the original task authoritative
- distinguish `Reads:` (context only), `Files:` (existing write targets), and `Creates:` (new write targets)
- validate scope safety and semantic adequacy before spawning implementers
- add one bounded reviewer -> planner reconsideration round with explicit accept/reject decisions
- retry unusable planner outputs across routed planner roles before aborting
- enforce the final write-scope contract against actual post-execution worktree changes
- surface out-of-scope writes as deterministic blocking findings so correction/revalidation can repair them

## Why

Tangle previously validated parseability, overlap, and execution success more strongly than whether the decomposition actually covered the requested deliverable or whether implementation stayed inside the declared write scope. That could let a structurally valid but semantically wrong decomposition proceed to spawn, or let a worker modify read-only/context paths without a deterministic blocker.

This change keeps the reviewer advisory, gives the planner one bounded chance to reconsider scope/coverage, and then fail-closes before spawn if the decomposition remains inadequate. After execution, changed paths must be contained by the final `Files:`/`Creates:` scopes; `Reads:` never grants write permission.

## Tests

Validated locally with the focused and regression suites, including:

- `test-tangle-decomposition-adequacy.sh` — 9/9
- `test-tangle-post-execution-scope-contract.sh` — 6/6
- `test-tangle-write-scope-safety.sh` — 21/21
- `test-tangle-context-review-loop.sh` — 56/56
- `test-tangle-correction-loop-behavior.sh` — 17/17
- `test-tangle-failed-validation-review-recovery.sh` — 5/5
- `test-review-aggregation-robustness.sh` — 32/32
- all 10 unit suites that call `tangle_develop`
- `bash -n` and `git diff --check`

Also replayed the deterministic post-execution scope checker against a preserved real Tangle worktree: it allowed the intended new frontend tree while identifying only the writes outside the final authorized scope.

## Summary by CodeRabbit

* **New Features**
  * Added clearer separation between files modified, artifacts created, and read-only context during task planning.
  * Added independent planning-quality review with limited reconsideration before work begins.
  * Design review results can now be captured for reuse while preserving existing output.
* **Bug Fixes**
  * Improved handling of invalid, incomplete, or overlapping task decompositions.
  * Prevented execution when planning is inadequate or changes exceed approved scope.
  * Added post-execution detection and reporting for unauthorized changes.
* **Tests**
  * Added coverage for planning adequacy, scope safety, review output, and post-execution validation.


Jhacarreiro:
Follow-up from a preserved real Tangle run: the run reached 6/6 terminal subtasks but died during quality retries before the post-execution scope check. Replaying the final decomposition against the preserved worktree with the new pre-gate reports out-of-scope writes to `tsconfig.json`, `web/FRAMEWORK_CHOICE.md`, and `web/src/vite-env.d.ts`, with `base_validate_calls=0`. Commit e0398a96 moves deterministic write-scope enforcement ahead of model-driven quality retries, so scope accounting survives interrupted retry execution. Focused regressions: scope contract 6/6, write-scope safety 21/21, correction loop 16/16, failed-validation recovery 5/5, contextual review 54/54, decomposition adequacy 9/9, integration value-proposition 19/19.

nyldn (commented):
Thanks for the substantial hardening work here. I reproduced several bypasses that prevent this PR from being merged as a security boundary:

- A worker can commit out-of-scope changes and finish with a clean worktree because validation does not compare the starting commit with the final `HEAD`.
- The worker can replace the writable baseline or scope manifest that final validation trusts.
- Replacing or deleting a staged symlink hides its original tree mode from final-state checks.
- A worker can create a temporary symlink, write outside the worktree, and remove the symlink before validation.
- `REASONING` subtasks return before the exact-one-nonempty-`Task:` check.

The first four need parent-owned authority data and immutable starting-tree comparisons. The temporary-symlink case also requires a mandatory, fail-closed execution boundary; final-state scanning cannot prove that an external write never happened.

Please keep this PR open while we settle that boundary. The safest direction is a launcher that refuses coding dispatch unless it can confine writes to the worktree and a sealed result channel. Once that exists, this branch can add `start_head..final_head` checks, read original modes from the starting tree, authenticate scope authority, and move `Task:` validation ahead of subtask-type returns.

I am not merging the current implementation because it would present scope enforcement as stronger than it is.

nyldn:
Closing this branch as superseded by #977. The current #975 head is an ancestor of #977, so keeping both open duplicates the same five commits and splits review attention.

The useful decomposition and scope ideas will be preserved while #977 is reconstructed after #1002. The present implementation is not safe to merge: it has state-delta false positives, fail-open paths, a dropped snapshot argument, Bash 3.2 failures, and a Linux-only boundary that blocks macOS runs.

PR #986, opened 2026-08-31
Prevent Codex updates from evicting active plugin cache

nyldn:
## Summary

- stop `update-plugin` before it replaces the versioned cache used by its own Codex session
- detect Codex through runtime markers or process ancestry while preserving updates from ordinary terminals
- explain the required exit, update, and restart sequence in user guidance
- remove machine-specific names from the public handoff

## Verification

- `bash tests/unit/test-plugin-update-health.sh` (42/42)
- disposable Linux ARM64 checkout: focused suite 42/42, Bash syntax, and `git diff --check`
- `CI=true GITHUB_ACTIONS=true make ci-changed` (16/16 smoke, 300/300 unit, 8/8 integration)
- `make sync-check`
- `git diff --check`

The two documented platform and dry-run skips remain unchanged. No release is included in this pull request.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented plugin updates from replacing files during an active Codex session.
  - Updates now fail safely when the session status cannot be determined.

- **Documentation**
  - Clarified that users must exit Codex, run updates from a separate terminal, and restart Codex afterward.
  - Added guidance for safely overriding session detection when appropriate.
  - Corrected the documented E2E runner for the v9.61.3 post-release note.

- **Tests**
  - Added coverage for active-session detection, safe blocking, and fallback scenarios.

### DIFF
--- a/AI_AGENT_HANDOFF.md
+++ b/AI_AGENT_HANDOFF.md
@@ -1,46 +1,24 @@
 # AI Agent Handoff
 
-Last updated: 2026-08-31
-Status: the Codex updater now refuses to replace its versioned plugin cache
-from inside the Codex session using that cache. The fix is committed and
-pushed, and PR #986 is open. Focused tests pass locally and on a disposable
-Linux ARM64 host, and the complete local matrix is green.
-Branch: `fix/codex-update-cache-lifecycle`, based on public `main` at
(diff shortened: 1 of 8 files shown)


nyldn:
@coderabbitai review

PR #992, opened 2026-09-02
test(ci): harden AGY doctor timeout regression

nyldn:
## Summary

- replace the obsolete `agy --version` timing test with a stalled `agy models` probe that Doctor actually invokes
- verify timeout reporting, parent and child cleanup, and that later AGY stages do not run after the catalog timeout
- keep a generous monotonic ceiling as a safety backstop instead of timing the entire Doctor workflow to four seconds

## Why

The old test stopped exercising its mock when shared provider readiness removed the version call. It then measured the complete Doctor command and failed intermittently on loaded macOS runners even though every AGY stage passed. This caused false failures on PRs #988 and #991.

## Verification

- RED mutation seam: AGY suite 50/51, with the unbounded catalog process completing and later stages running
- GREEN production path: AGY suite 51/51
- portable bare-probe timeout suite 9/9
- `make ci-changed` passed in focused mode
- `make ci-local` passed
- Bash 3.2 syntax, sync, diff, and executable-mode checks passed

## Summary by CodeRabbit

- **Tests**
  - Added coverage for provider catalog lookups that stall during health checks.
  - Verified stalled lookups and related processes are terminated within the configured timeout.
  - Confirmed subsequent model discovery and dispatch stages are skipped after a timeout.
  - Updated test execution to use the revised catalog-timeout scenario.
  - Added regression coverage for paths that could otherwise allow unbounded provider health checks.

### DIFF
--- a/tests/unit/test-agy-provider.sh
+++ b/tests/unit/test-agy-provider.sh
@@ -1021,12 +1021,15 @@ test_agy_doctor_catalog_timeout_is_bounded() {
     local parent_pid_marker="$TEST_TMP_DIR/agy-doctor-timeout.parent-pid"
     local child_pid_marker="$TEST_TMP_DIR/agy-doctor-timeout.child-pid"
     local mutation_bash_env="$TEST_TMP_DIR/agy-doctor-timeout-unbounded.bashenv"
+    local doctor_path="$tmp_bin:/usr/bin:/bin"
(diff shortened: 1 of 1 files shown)


nyldn:
@coderabbitai review

nyldn:
@coderabbitai review

nyldn:
@coderabbitai review

PR #995, opened 2026-09-03
fix(setup): accept readable preflight helper

### LINKED ISSUE #994: /octo:setup fails with "Octopus installation not found": scripts/helpers/preflight.sh is committed 100644 but the resolver gates on -x
## Summary

`/octo:setup` aborts with **"Octopus installation not found. Reinstall octo@nyldn-plugins, then run /octo:setup again."** on a perfectly healthy install.

The cause is that `scripts/helpers/preflight.sh` is committed with mode `100644`, while the resolver in `commands/setup.md` gates on `-x`. The reinstall the message suggests cannot fix it, because the missing bit is in the repository itself.

## Root cause

`commands/setup.md` resolves the plugin root by testing the executable bit:

```bash
[[ -x "$OCTO_ROOT/scripts/helpers/preflight.sh" ]]
```

All three fallbacks (`$CLAUDE_PLUGIN_ROOT`, `~/.claude-octopus/plugin`, and the `find` over the plugin cache) apply the same `-x` test, so every one fails and setup exits.

The file is never actually executed directly — both call sites invoke it as `bash "${OCTO_ROOT}/scripts/helpers/preflight.sh" --json`, which works fine. Only the guard is broken.

## Evidence

Mode bits at `main` (`0cf06ed`), via the git trees API:

```
.sh files in repo:            555
committed 100644 (non-exec):   54
committed 100755 (exec):      501

100755  scripts/helpers/check-providers.sh
100644  scripts/helpers/preflight.sh
100644  scripts/lib/preflight.sh
100644  scripts/lib/user-config.sh
```

`check-providers.sh` and `preflight.sh` sit in the same directory and are used the same way, but only one carries the bit — so this looks like drift rather than an intentional convention.

Other files committed `100644` include `scripts/install-all.sh`, `scripts/async-tmux-features.sh`, `scripts/helpers/check-versions.sh`, and `scripts/helpers/check-ollama-models.sh`.

In an installed plugin, 53 of 150 `.sh` files under `scripts/` lack the bit, plus 7 of 59 files under `hooks/`. `bin/` is unaffected (3 of 3 executable).

## Affected versions

Present in every cached version I have, so it is not a 10.x regression:

| Version | Non-executable `.sh` under `scripts/` |
|---|---|
| 9.64.0 | 52 / 147 |
| 9.66.1 | 53 / 149 |
| 10.1.0 | 53 / 150 |

## Reproduction

```bash
# Fresh install of octo@nyldn-plugins, then:
ls -l ~/.claude-octopus/plugin/scripts/helpers/preflight.sh
# -rw-r--r--   <- no execute bit

# The exact guard from commands/setup.md:
[[ -x ~/.claude-octopus/plugin/scripts/helpers/preflight.sh ]] && echo found || echo "not found"
# not found

# But the script itself is fine:
bash ~/.claude-octopus/plugin/scripts/helpers/preflight.sh --json | jq '.results | length'
# 17
```

## Suggested fix

Set the bit in git on the 54 affected files:

```bash
git ls-files -s -- '*.sh' | awk '$1=="100644" {print $4}' \
  | xargs git update-index --chmod=+x
```

A CI guard would stop it recurring:

```bash
if git ls-files -s -- '*.sh' | awk '$1=="100644"' | grep -q .; then
  echo "Shell scripts committed without the execute bit:"
  git ls-files -s -- '*.sh' | awk '$1=="100644" {print "  " $4}'
  exit 1
fi
```

Alternatively, relaxing the guard in `commands/setup.md` from `-x` to `-r` would also unblock setup, since every call site already runs the script through `bash`. Fixing the mode bits seems preferable, since `-x` guards may exist elsewhere.

## Impact

`/octo:setup` is unusable without manually running `chmod +x`, and its error message points users toward a reinstall that cannot resolve the problem. `scripts/lib/user-config.sh` is also affected; it is only ever `source`d, so it works today, but it would break under any future `-x` check.

nyldn:
## Summary

- let `/octo:setup` resolve a readable `preflight.sh` that it invokes through `bash`
- regenerate the Cursor command mirror from the canonical setup command
- execute both resolver snippets against a readable `0644` fixture, including a path with spaces

## Verification

- `make ci-changed` (full local matrix)
- `tests/unit/test-setup-first-success.sh` (8/8)
- independent Codex spec and quality reviews: PASS

Closes #994

## Summary by CodeRabbit

* **Bug Fixes**
  * Setup now recognizes readable `preflight.sh` helpers even when they are not executable.
  * Plugin resolution continues to use fallback paths and reports missing installations as before.
  * Unreadable helpers are skipped during fallback searches.
* **Tests**
  * Added coverage for readable, non-executable helpers and correct helper invocation.
  * Added coverage confirming unreadable helpers are skipped and fallback resolution works consistently across supported setup guides.

### DIFF
--- a/.cursor-plugin/commands/octo-setup.md
+++ b/.cursor-plugin/commands/octo-setup.md
@@ -33,7 +33,7 @@ if [[ ! -r "$OCTO_ROOT/scripts/helpers/preflight.sh" ]]; then
   OCTO_ROOT="${HOME}/.claude-octopus/plugin"
 fi
 if [[ ! -r "$OCTO_ROOT/scripts/helpers/preflight.sh" ]]; then
-  OCTO_ROOT="$(find "${HOME}/.claude/plugins/cache" "${HOME}/Library/Application Support/Claude" "${LOCALAPPDATA:-/dev/null}/Claude" "${XDG_DATA_HOME:-${HOME}/.local/share}/Claude" -maxdepth 8 -path '*/nyldn-plugins/octo/*/scripts/helpers/preflight.sh' -print -quit 2>/dev/null | sed 's#/scripts/helpers/preflight.sh$##')"
+  OCTO_ROOT="$(find "${HOME}/.claude/plugins/cache" "${HOME}/Library/Application Support/Claude" "${LOCALAPPDATA:-/dev/null}/Claude" "${XDG_DATA_HOME:-${HOME}/.local/share}/Claude" -maxdepth 8 -path '*/nyldn-plugins/octo/*/scripts/helpers/preflight.sh' -exec test -r '{}' \; -print -quit 2>/dev/null | sed 's#/scripts/helpers/preflight.sh$##')"
 fi
 [[ -r "$OCTO_ROOT/scripts/helpers/preflight.sh" ]] || {
   echo "Octopus installation not found. Reinstall octo@nyldn-plugins, then run /octo:setup again."
--- a/.cursor-plugin/commands/octo-setup.md
+++ b/.cursor-plugin/commands/octo-setup.md
@@ -29,13 +29,13 @@ made a choice.
 
 ```bash
 OCTO_ROOT="${CLAUDE_PLUGIN_ROOT:-}"
-if [[ ! -x "$OCTO_ROOT/scripts/helpers/preflight.sh" ]]; then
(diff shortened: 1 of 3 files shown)


nyldn:
@coderabbitai review

nyldn:
@coderabbitai review

nyldn:
@coderabbitai review

PR #998, opened 2026-09-03
fix(heartbeat): stop run_with_timeout's watchdog from holding the caller's pipe

### LINKED ISSUE #996: run_with_timeout leaks its watchdog `sleep`, blocking the result pipeline for the full timeout
## `run_with_timeout` leaks its watchdog `sleep`, blocking the result pipeline for the full timeout

**Version:** octo 9.30.0 · macOS 15 (Darwin 24.6.0) · bash 3.2 / zsh

### Summary

`run_with_timeout` (`scripts/lib/heartbeat.sh:131`) orphans its watchdog `sleep` process. The orphan
inherits the pipeline's stdout — the pipe into `tee` in `spawn_agent` — so `tee` never sees EOF, the
pipeline never completes, and `spawn_agent` never reaches its `## Status:` write.

`review_run`'s Round 1 poller (`scripts/lib/review.sh:334-346`) breaks only when every agent has a
`## Status:` marker, so **every review burns the entire `OCTOPUS_REVIEW_TIMEOUT`** even when all
models answered in seconds. Raising the timeout makes it strictly worse, because the orphaned `sleep`
sleeps for that same value.

### Mechanism

```bash
"$@" <&0 &
cmd_pid=$!
( sleep "$timeout_secs" && kill -TERM "$cmd_pid" 2>/dev/null ) &
monitor_pid=$!
wait "$cmd_pid"
kill "$monitor_pid" 2>/dev/null     # kills the SUBSHELL, not its `sleep` child
```

`kill "$monitor_pid"` terminates the subshell; its `sleep` child is reparented to PID 1 and keeps
running — still holding the inherited write end of the pipe.

### Evidence

Four agents dispatched at 07:06:43 with `TIMEOUT=480`. The models had written their answers to the
`.raw-*.out` files within seconds, yet all four `## Status:` markers appeared simultaneously at
**07:14:47** — exactly when the watchdogs expired. Round 1 took 484s instead of ~15s.

`lsof` shows the orphan and `tee` holding opposite ends of the same pipe:

```
sleep 47953  fd 1  PIPE 0x43608fafd3d7b8b5 -> 0x39318d0d38fee3cc
tee   47942  fd 0  PIPE 0x39318d0d38fee3cc -> 0x43608fafd3d7b8b5
```

### Reproduction

```bash
source scripts/lib/heartbeat.sh
myfunc() { sleep 2; printf 'fn-output\n'; }
printf x | run_with_timeout 300 myfunc | tee /tmp/o.txt >/dev/null &
sleep 5; pgrep -fx "sleep 300"   # -> orphan present, pipeline still blocked
```

Measured across both dispatch paths:

| dispatch path | `gtimeout` present | orphan | pipeline |
|---|---|---|---|
| external binary | yes | 0 | completes |
| external binary | no  | 1 | **blocked** |
| **shell function** | **yes** | **1** | **blocked** |
| shell function | no  | 1 | **blocked** |

### Two distinct impacts

**1. Any host without GNU coreutils.** macOS ships neither `timeout` nor `gtimeout`, so *all*
providers (gemini, codex, claude) take the leaky fallback. `brew install coreutils` works around it
— but that is an undocumented hard requirement today, and the failure mode is a silent hang with no
output, which reads as "octo is broken" rather than "install coreutils".

**2. Shell-function providers, on every host.** `_cmd_is_function=true` forces the fallback branch
*regardless* of `gtimeout`, so `openrouter_execute*` and `perplexity_execute` (`scripts/lib/perplexity.sh`)
always leak. coreutils does not help these; they hang for the full timeout on every run. This makes
OpenRouter and Perplexity effectively unusable in the review fleet.

### Fix (verified locally)

Two changes to the fallback branch. The redirect is what unblocks the pipeline; the `pkill -P` stops
the orphan lingering at all. Order matters in the cleanup — once the subshell dies, `pkill -P` can no
longer find its children.

```diff
-        ( sleep "$timeout_secs" && kill -TERM "$cmd_pid" 2>/dev/null ) &
+        ( sleep "$timeout_secs" && kill -TERM "$cmd_pid" 2>/dev/null ) >/dev/null 2>&1 &
         monitor_pid=$!
 ...
-        # Clean up monitor process
+        if command -v pkill >/dev/null 2>&1; then
+            pkill -P "$monitor_pid" 2>/dev/null
+        fi
         kill "$monitor_pid" 2>/dev/null
         wait "$monitor_pid" 2>/dev/null
```

`pkill` is guarded rather than assumed, per the acceptance criteria on #827 ("No undeclared `pkill`
dependency… regression runs with `timeout`, `pgrep`, and `pkill` unavailable"). The two halves are
independent, and it is worth being precise about which does what:

- **The stdout redirect is the actual fix.** It needs no external tools and is what unblocks the
  pipeline. Verified on a PATH containing only `bash`/`sleep`/`tee`/`ps`: pipeline completes.
- **The `pkill -P` is belt-and-braces reaping.** Without it the orphan lingers until its sleep
  elapses, but it no longer holds the pipe, so nothing blocks.

This is the same failure class as #827 — a descendant surviving teardown — in
`run_with_timeout` rather than `_octopus_agent_lifecycle_event`.

Results across all four path/environment combinations:

| dispatch path | `gtimeout` | before | after |
|---|---|---|---|
| external binary | yes | 0 orphans, completes | 0 orphans, completes |
| external binary | no  | 1 orphan, **blocked** | 0 orphans, completes |
| shell function  | yes | 1 orphan, **blocked** | 0 orphans, completes |
| shell function  | no  | 1 orphan, **blocked** | 0 orphans, completes |

End-to-end `/octo:review` on a 22-line staged diff, `OCTOPUS_REVIEW_TIMEOUT` at its default 480s:

- **Before:** Round 1 484s, full run ~13 min
- **After (coreutils only):** Round 1 67s, full run 204s
- **After (this patch):** Round 1 107s, full run 231s, zero orphans — and no longer dependent on coreutils

Because this also repairs the shell-function branch, it is what makes the OpenRouter and Perplexity
providers usable at all; `gtimeout` cannot reach that path.

Adding `coreutils` to the documented prerequisites is still worthwhile (the `gtimeout` fast path
avoids the fallback entirely), but with this patch it is an optimisation rather than a hard
requirement.

jonnocraig:
## Description

The in-process fallback of `run_with_timeout` backgrounds its watchdog as a subshell that `sleep`s for the timeout. On a normal completion the parent kills the subshell, but the subshell's `sleep` child is reparented rather than killed, and it keeps every file descriptor it inherited — including the caller's stdout.

`spawn_agent` pipes `run_with_timeout` into `tee`, so that orphaned `sleep` holds the write end of the pipe open. `tee` never sees EOF and the pipeline blocks for the **full timeout** even though the command has already finished and produced its output. The shell-function dispatch path (`perplexity_execute`, `openrouter_execute`) always takes this fallback regardless of whether `gtimeout`/`timeout` is installed, so those providers are the most exposed — a review can appear to hang for the whole `OCTOPUS_REVIEW_TIMEOUT` with no output. (In `review_run` Round 1, the `## Status:` marker poll only breaks when every agent's marker is present, so one held pipe stalls the whole round.)

The fix redirects the watchdog subshell's stdout to `/dev/null` so it never inherits the caller's pipe. The orphaned `sleep` still lingers until it elapses, but it no longer holds the pipe, so `tee` closes and the pipeline finishes as soon as the command does. It is deliberately minimal and does not touch the process-tree signalling added for #336/#738.

```diff
-        ) &
+        ) >/dev/null 2>&1 &
         monitor_pid=$!
```

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests pass: `bash tests/unit/test-openclaw-compat.sh` (33/33)
- [ ] OpenClaw registry in sync — n/a (no adapter/registry change)
- [ ] New skills/commands registered — n/a
- [ ] Version bump — deferred to a release commit
- [x] CHANGELOG.md updated (Unreleased → Fixed)

## Testing

Added `tests/unit/test-issue-996-run-with-timeout-pipe-leak.sh`, a behavioural regression test. It dispatches a shell function (which forces the in-process fallback) through `run_with_timeout` into a `tee` pipeline and asserts the pipeline completes. It uses a **bounded poll rather than a fixed wait**, so it cannot flake on a slow runner, and the timeout window (6s) sits well clear of the poll ceiling (3.5s) in both directions.

Verified it is a real regression test — **fails on the pre-fix code, passes after**:

```
# pre-fix:  FAIL — "pipeline still blocked after 3.5s — watchdog sleep is holding the caller's stdout (#996)"
# post-fix: PASS
```

I could not run the full `tests/run-all-tests.sh` locally to completion (it exceeds 10 min on my machine — that's CI's job), but ran every affected-area suite green:

```
test-issue-996-run-with-timeout-pipe-leak   1/1
test-heartbeat-timeout                      13/13
test-agent-timeout-budget                    7/7
test-bare-auth-probe-timeout                 9/9
test-stdin-timeout-guards                   12/12
test-preserve-caller-process-group           3/3
test-agent-lifecycle-events                 10/10
test-lifecycle-events-498                    7/7
```

Happy to adjust the CHANGELOG wording or the test's form to your preference.

## Related Issues
Closes #996

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where timeout monitoring could keep output pipelines open after a command completed.
  - Command output now finishes promptly instead of potentially waiting for the full timeout period.

- **Tests**
  - Added regression coverage to verify timely pipeline completion and correct output capture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Fixed
 
+- Stop `run_with_timeout`'s in-process watchdog from holding the caller's
+  stdout: its orphaned `sleep` kept the pipe into `spawn_agent`'s `tee` open,
+  so a review could block for the full timeout after the command had already
+  produced output (worst on the shell-function provider path). ([#996](https://github.com/nyldn/claude-octopus/issues/996))
 - Store `/octo:plan` artifacts in unique, resolved run directories and share
   that location with plan-mode hooks and review skills, preventing writes into
   the global `~/.claude/` configuration directory and same-session overwrites.
--- a/scripts/lib/heartbeat.sh
+++ b/scripts/lib/heartbeat.sh
@@ -333,7 +333,13 @@ run_with_timeout() {
             done
 
             _octo_timeout_signal_snapshot KILL "$process_tree"
-        ) &
+        # Redirect the monitor's stdout away from the caller's: on the shell-
+        # function path this subshell inherits the pipe into spawn_agent's `tee`,
+        # and its `sleep` child is orphaned (reparented) when the monitor is
+        # killed on a normal completion — an orphan holding the pipe's write end
+        # keeps `tee` from ever seeing EOF, so the pipeline blocks for the full
+        # timeout even though the command already produced its output.
+        ) >/dev/null 2>&1 &
         monitor_pid=$!
 
         if wait "$cmd_pid" 2>/dev/null; then
--- a/tests/unit/test-issue-996-run-with-timeout-pipe-leak.sh
+++ b/tests/unit/test-issue-996-run-with-timeout-pipe-leak.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+# tests/unit/test-issue-996-run-with-timeout-pipe-leak.sh
+# Regression test for #996: the in-process fallback of run_with_timeout must not
+# let its watchdog's orphaned `sleep` hold the caller's stdout. spawn_agent pipes
+# run_with_timeout into `tee`; a held pipe never sees EOF, so the whole review
+# blocks for the FULL timeout even though the command already produced output.
+# The shell-function path (perplexity_execute, openrouter_execute) always takes
+# this fallback, regardless of whether gtimeout/timeout is installed.
+set -euo pipefail
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+source "$SCRIPT_DIR/../helpers/test-framework.sh"
+log() { :; }
+source "$PROJECT_ROOT/scripts/lib/heartbeat.sh"
+
+test_suite "run_with_timeout pipe leak (#996)"
+
+test_case "shell-function dispatch does not leave a downstream pipe blocked"
+# A shell function forces the in-process fallback. The probe sleeps briefly so the
+# watchdog subshell has certainly spawned its own `sleep` child before the command
+# returns — that is the fault window; killing the watchdog then orphans that sleep.
+# Timeout is 6s: if the orphan is holding the caller's pipe, `tee` stays blocked
+# until 6s; the fix lets the pipeline finish about when the probe does (~1s). The
+# poll ceiling (3.5s) sits between the two, and any orphan self-terminates by 6s.
+outfile="$TEST_TMP_DIR/pipe-996.out"
+donefile="$TEST_TMP_DIR/pipe-996.done"
+rm -f "$outfile" "$donefile"
+_octo_996_probe() { sleep 1; printf 'PROBE_OUTPUT\n'; }
+( printf '' | run_with_timeout 6 _octo_996_probe | tee "$outfile" >/dev/null; echo done > "$donefile" ) &
+pipeline_pid=$!
+
+# Bounded poll (no fixed long wait): break the instant the pipeline completes, so
+# a slow runner cannot flake it. Ceiling (3s) sits between near-instant (fixed)
+# and the 5s block (bug), so the two outcomes are unambiguous.
+completed=false
+for _ in $(seq 1 35); do
+    if [[ -f "$donefile" ]]; then completed=true; break; fi
+    sleep 0.1
+done
+
+if [[ "$completed" == "true" ]] && [[ "$(cat "$outfile" 2>/dev/null)" == "PROBE_OUTPUT" ]]; then
+    test_pass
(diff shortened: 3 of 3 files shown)


nyldn:
Closing this in favor of #987, which fixes the same pipe-hold defect in the shared `run_with_timeout` path and also terminates and reaps the timer and provider process groups. The exact #987 head passed the protected Ubuntu and macOS matrix before squash-merging as `e7c6773b`.

#998 is now conflicted with `main` and deliberately leaves the orphaned sleeper alive, so merging it would retain a narrower timeout implementation. Thank you for isolating the original descriptor leak and adding a focused reproduction.

PR #1014, opened 2026-09-05
fix(review): fail closed on incomplete diff snapshots

Jhacarreiro:
## Summary

- make contextual-review diff collection fail closed when Git or helper collection fails
- preserve `git diff --no-index` exit code 1 as the normal differences-found case while propagating real errors
- keep NUL-safe untracked/unborn enumeration without allowing scratch files to contaminate the worktree diff
- reject partial initial or retry snapshots and fail closed when `git status` cannot be inspected
- restrict persisted review snapshots to owner-only mode `0400`

## Why

Follow-up hardening after #1002. A failed diff command could previously be normalized to success by `|| true`, allowing a partial or empty review snapshot to be accepted. This makes the review input integrity invariant explicit: incomplete collection is an error, not an empty diff.

## Tests

- `./tests/unit/test-tangle-context-review-loop.sh` — 99/99
- `./tests/unit/test-review-run.sh` — 43/43
- `./tests/unit/test-command-staged-review.sh` — 9/9
- `./tests/unit/test-review-aggregation-robustness.sh`
- `./tests/unit/test-review-round1-spawn-failure.sh`
- `./tests/unit/test-cross-model-review.sh`
- `./tests/unit/test-contextual-review-fleet-cli.sh`
- `LC_ALL=C.UTF-8 ./scripts/build-codex-skills.sh --check`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Git and review snapshot errors are now surfaced instead of being silently ignored.
  - Review snapshots reject incomplete or partial diff data, including failures during retry attempts.
  - Temporary repository artifacts are excluded from collected review changes and cleaned up reliably.
  - Review snapshots are now restricted to owner-only access.
  - Repositories without an initial commit are handled more reliably when collecting review changes.

### DIFF
--- a/scripts/lib/review.sh
+++ b/scripts/lib/review.sh
@@ -1205,6 +1205,29 @@ review_local_synthesis_json() {
     fi
 }
 
+# review_collect_scratchpad_dir: choose the process scratchpad without placing
+# temporary collection files in the repository or its Git metadata.
+review_collect_scratchpad_dir() {
+    local repo_root="$1"
+    local scratch_dir="${OCTOPUS_TMP_DIR:-${TMPDIR:-/tmp}}"
+    local repo_root_physical=""
+    local scratch_dir_physical=""
+
+    repo_root_physical=$(cd "$repo_root" 2>/dev/null && pwd -P) || repo_root_physical=""
+    scratch_dir_physical=$(cd "$scratch_dir" 2>/dev/null && pwd -P) || scratch_dir_physical=""
+    if [[ -n "$repo_root_physical" && -n "$scratch_dir_physical" ]]; then
+        case "$scratch_dir_physical" in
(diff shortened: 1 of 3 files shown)


Jhacarreiro:
ClawSweeper 🐠 reef update

Thanks for the contribution here. ClawSweeper gave this branch a little current boost so the original PR can stay the canonical swim lane instead of opening a replacement.

Validation: bash -n scripts/lib/review.sh scripts/lib/workflows.sh tests/unit/test-tangle-context-review-loop.sh; git diff --check; bash ./tests/unit/test-tangle-context-review-loop.sh; bash ./tests/unit/test-review-run.sh; make sync-check
Contributor credit stays on this reef marker, with the PR history doing the receipts.

fish notes: reasoning high; reviewed against 8b2d3c3d4921.

PR #1022, opened 2026-09-07
fix: release v11.2.1 orchestrator reliability fixes

nyldn:
## Summary

Release v11.2.1 with the five reviewed orchestrator fixes:

- Verify worker identity before cancellation; acknowledge registration before returning and retire it on exit.
- Delegate the legacy release command to the maintained release workflow.
- Recover interrupted research when no synthesis marker exists.
- Honor provider, phase, role and Tangle operation overrides in workflow summaries.
- Preserve unattended mode during CI initialization.

Update release manifests, README, changelog and generated marketplace metadata. Keep the README summary focused on user benefits.

## Verification

- Full local gate passed 16 smoke, 325 unit and 8 integration suites.
- Focused checks cover delayed and failed worker registration, concurrent ledger updates, stale identities, release argument forwarding, markerless recovery, model overrides and unattended environments.
- Final GPT-6 Astra high code review returned NO ACTIONABLE FINDINGS.
- Release metadata checks and exact-head hosted CI are required before merge. The tag and GitHub release will follow successful main-branch verification; shared marketplace publication follows the release.

No live-provider compatibility claim is made by the fixture-based tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation reliability by verifying worker identities and safely cleaning up worker processes.
  * Restored interrupted research results when synthesis markers are unavailable.
  * Improved workflow model summaries using the configured provider, phase, and role.
  * Preserved unattended mode in Jenkins and background-task-disabled environments.
  * Routed the legacy release command through the maintained release workflow.
  * Reduced unnecessary prompt cleanup delays when workers exit.

* **Documentation**
  * Updated release documentation and version references for v11.2.1.
  * Documented release previews, validation behavior, and platform-specific process cancellation.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "nyldn plugins for Claude Code workflows.",
-    "version": "11.2.0"
+    "version": "11.2.1"
   },
   "plugins": [
     {
@@ -15,8 +15,8 @@
         "source": "url",
         "url": "https://github.com/nyldn/claude-octopus.git"
       },
-      "description": "v11.2.0 - Method-aware workflows, clearer independent review escalation, and bounded engineering prototypes. 31 personas, 53 commands, 63 skills. Run /octo:setup.",
-      "version": "11.2.0",
+      "description": "v11.2.1 - Reliable cancellation, interrupted research recovery, and accurate workflow model summaries. 31 personas, 53 commands, 63 skills. Run /octo:setup.",
+      "version": "11.2.1",
       "author": {
         "name": "nyldn",
         "url": "https://github.com/nyldn"
--- a/.claude-plugin/plugin-manifest.json
+++ b/.claude-plugin/plugin-manifest.json
@@ -3,7 +3,7 @@
   "schemaVersion": "2026-06",
   "name": "octo",
   "displayName": "Claude Octopus",
-  "version": "11.2.0",
+  "version": "11.2.1",
   "description": "Multi-AI orchestration for Claude Code: up to 8 model seats in parallel across research, design, coding, and review, with council verdicts, adversarial debate, and per-provider cost attribution.",
   "author": {
     "name": "nyldn",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "octo",
-  "version": "11.2.0",
-  "description": "v11.2.0 \u2014 Method-aware workflows, clearer independent review escalation, and bounded engineering prototypes. Run /octo:setup.",
+  "version": "11.2.1",
+  "description": "v11.2.1 \u2014 Reliable cancellation, interrupted research recovery, and accurate workflow model summaries. Run /octo:setup.",
   "author": {
     "name": "nyldn",
     "url": "https://github.com/nyldn"
--- a/.claude-plugin/routines.json
+++ b/.claude-plugin/routines.json
@@ -1,5 +1,5 @@
 {
-  "$comment": "Claude Octopus routine manifest (v11.2.0). Saved automation configs for Claude Code routines: schedule triggers map to cron-style saved automations, github triggers map to repository event automations. All routines ship disabled; enable per-project after reviewing provider cost implications (external CLI seats bill to the user's API keys).",
+  "$comment": "Claude Octopus routine manifest (v11.2.1). Saved automation configs for Claude Code routines: schedule triggers map to cron-style saved automations, github triggers map to repository event automations. All routines ship disabled; enable per-project after reviewing provider cost implications (external CLI seats bill to the user's API keys).",
   "schemaVersion": 1,
   "routines": [
     {
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-octopus",
-  "version": "11.2.0",
+  "version": "11.2.1",
   "description": "Multi-LLM orchestration with Double Diamond workflows, provider routing, safety gates, and automation. Dispatches to Codex, Antigravity, Cursor CLI, Grok, Copilot, Qwen, Perplexity, OpenRouter, OrcaRouter, Ollama, OpenCode, and Kimi Code from a single plugin.",
   "author": {
     "name": "nyldn"
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "claude-octopus",
   "displayName": "Claude Octopus",
   "description": "Multi-LLM orchestration \u2014 up to 8 providers check each other's work. Research, build, review, and ship with consensus gates.",
-  "version": "11.2.0",
+  "version": "11.2.1",
   "author": {
     "name": "nyldn"
   },
--- a/.factory-plugin/marketplace.json
+++ b/.factory-plugin/marketplace.json
@@ -5,14 +5,14 @@
   },
   "metadata": {
     "description": "Multi-tentacled orchestration plugin for Factory AI Droid",
-    "version": "11.2.0"
+    "version": "11.2.1"
   },
   "plugins": [
     {
       "name": "claude-octopus",
       "source": "./",
-      "description": "v11.2.0 - Multi-AI orchestration with Double Diamond workflow. 31 personas, 53 commands, 63 skills. Run /octo:setup after install.",
-      "version": "11.2.0",
+      "description": "v11.2.1 - Multi-AI orchestration with Double Diamond workflow. 31 personas, 53 commands, 63 skills. Run /octo:setup after install.",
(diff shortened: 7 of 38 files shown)


nyldn:
The four review findings are addressed in the cancellation follow-up:

- The worker-ownership predicate consumes the full job listing. A regression reproduces the original SIGPIPE failure under pipefail.
- Probe, Tangle and scoped review cleanup verify ledger identities. Probe and Tangle also recheck the task registration before signalling a PID from memory. Probe removes rejected PIDs before its later wait and heartbeat cleanup.
- The concurrent-retirement test starts with an existing row and checks that retirement removes it while preserving the parallel registrations.
- CI-mode coverage rejects an empty extraction before evaluation.

Focused checks passed 30 orchestrator regressions, 21 Probe cancellation cases, 16 Tangle cancellation cases, 100 contextual-review cases and 30 background-lifecycle cases. A fresh GPT-6 Astra high review of the final patch returned NO ACTIONABLE FINDINGS. The full local matrix is running; hosted checks must pass on the new head.

The separate suggestion to fall back to trusting an unverifiable PID is not adopted. Rejecting dispatch when process identity cannot be established is intentional; restoring an unverified-PID fallback would undo the cancellation fix.

@coderabbitai review

nyldn:
The full local follow-up run passed 16 smoke suites and 324 of 325 unit suites. The remaining failure came from a live-worker cancellation fixture that still wrote a legacy three-field ledger row. Both remaining live-worker fixtures now use octopus_pid_register; each passed 4/4. All 8 integration suites passed afterward. This latest commit changes tests and the handoff only; production code remains the final reviewed cancellation patch.

The automated suggestion to restore dispatch without Python is not adopted as a fallback. Guided setup already invokes Python directly for receipts and readiness validation in commands/setup.md, and the structural safety guards also require it. Unverifiable process identity must fail closed.

The model-summary keys preview configured roles, not every conditional reviewer or synthesis sub-operation. A complete per-operation execution plan is separate from correcting the provider/phase/role arguments in this release. Constant-time comparison of public process metadata is also not required for this local identity check.

nyldn:
Release checkpoint for `627c18cfafcdc90b45211e189564d7d5010b87c2`: the candidate is pushed and the release worktree is clean. No merge, tag, GitHub release or shared marketplace update has occurred.

The latest review identifies a remaining race in the shared teardown helpers. Source inspection confirms that `review_terminate_process_tree` snapshots numeric PIDs and signals them again after a grace period without checking whether they still identify the same processes. `review_kill_process_tree_frozen` also signals numeric PIDs and process groups after the caller's identity check. These helpers are unchanged from main. The candidate prevents stale ledger registrations from entering cancellation, but that does not make later signal delivery atomic with the identity check.

This concern remains unresolved. A proper fix needs a cross-platform lifecycle decision, including how macOS retains working cancellation without Linux PID handles. The release is paused rather than treating passing regression tests as proof that this race is eliminated.

The private handoff records the exact pushed candidate and the remaining work. Hosted run `34095691855` has passed portability, packaging, both smoke platforms and symlink checks; full unit jobs are still running. Local focused regressions and all eight integration suites passed, including targeted reruns of the corrected worker fixtures.

nyldn:
The release candidate is pushed and clean at `a4120376468a1591bdda8ec1f06dbacdb8e89132` on `release/v11.2.1`.

The remaining signal-identity finding is addressed in `scripts/helpers/process_control.py`. Linux retains PID handles through STOP, TERM and KILL. macOS uses `proc_signal_with_audittoken`, with PID/version checks inside XNU signal delivery. An exec-version retry must still match the original boot/PID/birth identity. Ledger identities now reach the actual termination helper in direct, Probe, Tangle and scoped cancellation paths. There is no numeric-signal fallback in this helper.

The helper records restoration responsibility before STOP and defers further interrupts while restoring and closing handles. A process disappearing during stop confirmation no longer abandons the other captured workers. Failed scoped cleanup does not wait on a live worker. Probe and Tangle retain registrations when cleanup fails, and worker registration checks native cancellation capability before provider dispatch.

The simplification removes duplicated shell tree walks, repeated external process probes, unconditional grace-period waits and an unused verifier wrapper. Tests now load the real review module instead of extracting a fragile collection of function definitions. The docs suite no longer overwrites its failure array with an obsolete numeric counter.

Verification:

- Full local matrix passed 16 smoke suites, 326 unit suites and all 8 integration suites.
- Final focused checks passed native control 20/20, orchestration 33/33, Probe 21/21, Tangle 16/16, background lifecycle 30/30, contextual review 100/100, review aggregation 41/41 and PID capture 14/14.
- The docs suite passed 143/143 without its false failure section.
- ShellCheck, diff checks and package dry-run passed. Both Python helpers are packaged and executable.
- A fresh Astra high review returned `NO ACTIONABLE FINDINGS` after its three verified findings were corrected.

The handoff records the previous exact pushed checkpoint and the new full-matrix result. This comment records the subsequent immutable commit. The earlier 324/325 historical result was not rewritten into a success that did not occur; this new run actually passed all 326 suites.

Native runtime tests have passed locally on macOS. Real Linux execution and hosted macOS checks are still pending on this exact head. No merge or release is claimed yet. Please re-review the updated candidate.

nyldn:
Checked the latest advisory review against `a4120376468a1591bdda8ec1f06dbacdb8e89132`:

- Registration calls `Process(int(pid))`, which calls `snapshot(pid)`. `snapshot` rejects PID values at or below one before any native lookup. The reported PID 0/1 admission path does not exist.
- Production prune callers pass `probe-${task_group}-` or `tangle-${task_group}-`, including the final delimiter. For example, `tangle-1-` cannot match `tangle-10-`. Prefix matching intentionally removes a workflow's tasks; exact worker retirement remains separate.
- The redundant empty-agent check is harmless. It does not justify another production revision and complete release-check cycle.

The review log also shows large truncated specialist prompts, an unavailable Codex verifier, and invalid fallback verification JSON. These results are advisory evidence, not independent verification. The separate source-grounded Astra review passed, and Linux/macOS deterministic checks remain required before merge. I have recorded the hosted-review efficiency problem for separate work.

nyldn:
Final follow-up is pushed at `7bc076b40f410086e4a91083183521ac8f50b062`.

The native implementation's full hosted Test Suite on `a4120376` completed successfully, including real Linux and macOS process-control tests and integration checks. The only subsequent production change is shared PID-range validation before native conversion. It rejects non-integer and out-of-range inputs before platform lookup or ancestor enumeration.

The mocked boundary regression failed before the guard and passes after it. Focused native checks pass 21/21 and orchestration checks pass 33/33. A fresh review limited to the final guard returned `NO ACTIONABLE FINDINGS`. The new exact-head hosted run is `34151656098`; merge still waits for its checks and explicit review approval.

nyldn:
Pushed commit `613a0213fff1633a65257c26e35451956268eeb0` addresses the five outstanding review threads:

- Descendant admission now reports incomplete cleanup if the parent exits. It retains the parent-instance check because a numeric PPID alone cannot establish ownership after that point. The regression requires an error result and no signal to the uncertain child.
- Probe and Tangle verify their ledger rows in one Python invocation and retain the tokens for native signal binding. Coverage checks multiple rows, unrelated prefixes, invalid identities and the helper invocation count.
- The Tangle retention test now defines its failing cleanup stub locally. The previous fixture inherited that same failing stub from the earlier scoped-cleanup case, so it already exercised failure; the explicit definition removes that hidden dependency.
- The child fixture publishes its PID marker atomically.
- The native test wrapper now uses the shared test framework.

Focused checks passed 22 native process-control cases, 34 orchestrator regressions, 21 Probe, 16 Tangle, 4 v10 recovery and 100 contextual-review cases. Workflow regressions also passed under macOS Bash 3.2. The fresh Astra high review returned `NO ACTIONABLE FINDINGS`.

The final full local matrix passed all 16 smoke, 326 unit and 8 integration suites. Diff whitespace and executable-mode checks passed.

The existing release gates remain unchanged. Publication still requires approval, zero unresolved threads, successful checks on the final PR head and a successful Test Suite on the squash commit.

nyldn:
@coderabbitai review

Please review the final head, `613a0213fff1633a65257c26e35451956268eeb0`. The five review follow-ups are committed and their verification evidence is in the preceding comment. The implementation is frozen while exact-head hosted checks finish.

nyldn:
I checked the advisory review against the complete source at `613a0213fff1633a65257c26e35451956268eeb0`.

The two correctness warnings are unsupported:

- `pid-ledger.py:71-79` rejects colons in task IDs and line breaks in either field, then escapes percent signs and colons in agent IDs before serialization. The native process identity generates the token. The report omitted this caller when evaluating `update()`.
- Both verification paths require a nonempty token before comparison: batch verification at line 63 and single verification at line 68. An empty lookup result therefore cannot authorize cancellation.

Parsing the same small PID twice is harmless and does not justify another release revision. The useful optimization in this patch removes repeated interpreter launches across workflow verification.

The hosted review again truncated specialist context and could not run its Codex verifier. Follow-up `oco-imu` tracks those review-quality and efficiency problems. No production changes are needed for these advisory warnings.

PR #1040, opened 2026-09-15
fix: drop invalid matcher-group 'if' keys from hooks.json

### LINKED ISSUE #1039: hooks.json: `if` on two matcher groups uses a schema Claude Code does not have — warning at every session start, and /octo:freeze would silently break if it were ever honoured
## Environment

- octo 11.4.1 (installed 2026-09-11); the same two keys are present in v11.5.0 and on `main` as of 2026-09-15
- Claude Code 2.1.269 (current release at the time), macOS

## Symptom

Every session starts with:

```
octo: hooks.json: unknown keys "if" in hooks.PreToolUse[6], "if" in hooks.PostToolUse[3] ignored
```

## Cause

`hooks/hooks.json` puts `if` on two *matcher-group* objects, with env expressions:

```json
{ "matcher": "Edit|Write", "if": "env.OCTO_FREEZE_MODE == 'on'", "hooks": [ { "command": ".../hooks/freeze-check.sh" } ] }
{ "matcher": "Bash",       "if": "env.OCTOPUS_WEBHOOK_URL",       "hooks": [ { "command": ".../hooks/telemetry-webhook.sh" } ] }
```

Per the hooks reference (https://code.claude.com/docs/en/hooks-guide, "Filter by tool name and arguments with the `if` field"), `if`:

- is valid only on an individual hook object (inside the `hooks` array), not on the matcher group;
- accepts permission-rule syntax only (`Bash(git *)`, `Edit(*.ts)`), not expressions and not environment variables;
- was introduced in 2.1.85 and has not changed shape since.

The five hook-level `if`s elsewhere in the same file (`Bash(codex *)`, `Bash(*orchestrate.sh*)`, …) are correct and produce no warning. Reproduce the offending keys with:

```
jq '[.hooks[]?[]? | select(has("if")) | {matcher, if}]' hooks/hooks.json
```

## Impact today

Cosmetic, plus two extra process spawns per tool call: because the keys are ignored, `freeze-check.sh` runs on every Edit/Write and `telemetry-webhook.sh` on every Bash call. Both gate themselves (`freeze-check.sh` exits 0 unless `/octo:freeze` has written its per-session state file; `telemetry-webhook.sh` exits 0 when `OCTOPUS_WEBHOOK_URL` is unset), so behaviour is correct.

## The latent bug worth fixing now

`freeze-check.sh`'s kill switch is `[[ "${OCTO_FREEZE_MODE:-on}" == "off" ]] && exit 0` — unset means "on", and enforcement is decided by the state file that `/octo:freeze` writes. Nothing exports `OCTO_FREEZE_MODE=on`. So if Claude Code ever *did* honour `env.OCTO_FREEZE_MODE == 'on'`, the freeze hook would stop running unless the user had manually exported that variable, and `/octo:freeze` would silently stop being enforced. The condition as written is the inverse of the script's own contract.

## Suggested fix

Delete the two group-level `if` keys; the scripts already do the gating, and there is no `if` syntax that can express "env var is set" today. Optionally extend `tests/smoke/test-fleet-dispatch-guard.sh` (which already asserts every block has a `matcher`) to assert that no group object carries `if`, and that every hook-level `if` matches `^[A-Za-z]+\(.*\)$`.

## Verification

With the two keys deleted the warning is gone from the next session on (Claude Code reads hooks.json before SessionStart hooks run), all 42 hook entries and 5 hook-level `if` filters remain, and `tests/smoke/test-fleet-dispatch-guard.sh`, `test-safety-hooks.sh` and `test-packaging-integrity.sh` pass unchanged. Measured on 11.4.1 with the keys stripped locally.

nyldn:
## Description

`hooks/hooks.json` had two `if` keys on *matcher-group* objects using
env-expression syntax:

```json
{ "matcher": "Edit|Write", "if": "env.OCTO_FREEZE_MODE == 'on'", ... }
{ "matcher": "Bash",       "if": "env.OCTOPUS_WEBHOOK_URL", ... }
```

Per Claude Code's hooks reference, `if` is valid only on an individual hook
object inside the `hooks` array, and only accepts permission-rule syntax
(`Bash(git *)`, `Edit(*.ts)`) — never env expressions. Claude Code silently
ignores both keys and prints a warning at every session start:

```
octo: hooks.json: unknown keys "if" in hooks.PreToolUse[6], "if" in hooks.PostToolUse[3] ignored
```

Because the keys are ignored, `freeze-check.sh` and `telemetry-webhook.sh`
run on every Edit/Write/Bash call respectively. Both scripts already gate
themselves internally (`freeze-check.sh` checks `OCTO_FREEZE_MODE`,
`telemetry-webhook.sh` checks `OCTOPUS_WEBHOOK_URL`), so behavior stayed
correct — this was cosmetic plus two extra no-op process spawns.

Worth calling out separately: `freeze-check.sh`'s condition as written
(`env.OCTO_FREEZE_MODE == 'on'`) is the inverse of the script's own contract
(`[[ "${OCTO_FREEZE_MODE:-on}" == "off" ]] && exit 0` — unset means *on*).
Had Claude Code ever honored that group-level condition, `/octo:freeze`
enforcement would have silently stopped firing unless a user manually
exported `OCTO_FREEZE_MODE=on`. Deleting the key (rather than "fixing" its
condition) is correct here since no `if` syntax exists today that can
express "env var is set" at the group level.

## Root cause

- `hooks/hooks.json:82` — group-level `if` on the `Edit|Write` PreToolUse block
- `hooks/hooks.json:126` — group-level `if` on the `Bash` PostToolUse block

## Fix

Deleted both group-level `if` keys. The five hook-level `if`s elsewhere in
the file (`Bash(codex *)`, `Bash(*orchestrate.sh*)`, etc.) were already
correct and are untouched. The gating scripts already handle their own
conditions, so no behavior changes.

Also extended `tests/smoke/test-fleet-dispatch-guard.sh` with
`test_hooks_json_if_schema`, which asserts no matcher-group object carries
an `if` key and every hook-level `if` matches `^[A-Za-z]+\(.*\)$`. Verified
this test fails against the pre-fix file and passes against the fix.

## Testing

```bash
jq empty hooks/hooks.json                                  # valid JSON
bash -n scripts/orchestrate.sh                              # syntax OK
bash tests/smoke/test-fleet-dispatch-guard.sh                # 5/5 pass (incl. new check)
bash tests/smoke/test-safety-hooks.sh                        # 25/25 pass
bash tests/smoke/test-packaging-integrity.sh                 # 10/10 pass
```

Confirmed the new regression check fails when run against the original
(pre-fix) `hooks/hooks.json` and passes against this branch's version.

Not run: `make test-unit` / `make test-integration` in full — this change
is a two-line JSON deletion plus a test-only addition with no code path
changes, so the smoke tests above are the meaningful surface. Happy to run
the full suite if requested.

No CHANGELOG entry in this PR — per the repo's fix→release split, that's
added at release time listing the merged fix PRs.

## Related Issues
Closes #1039

https://claude.ai/code/session_01CUF9FuCz5G5MLTmMzQvmga

---
_Generated by [Claude Code](https://claude.ai/code/session_01CUF9FuCz5G5MLTmMzQvmga)_

## Summary by CodeRabbit

- **Behavior Changes**
  - Freeze-check automation now runs for every edit or write event, regardless of freeze mode.
  - Telemetry webhook automation now runs for every matching Bash command, regardless of webhook URL configuration.

- **Tests**
  - Added validation to ensure hook conditions use the supported configuration format and are placed at the correct level.
  - Smoke tests now detect unsupported condition placement, invalid formatting, and malformed hook configuration before deployment.


nyldn:
Re the Octopus review NIT on `test_hooks_json_if_schema`'s regex not validating tool *identity* (would still pass a typo like `Bsah(codex *)`): agreed it's a real gap, but I'm not fixing it in this PR. The proposed remedy — constraining the prefix to an enumerated tool-name allowlist — would need to include every built-in tool plus arbitrary MCP tool names (e.g. `mcp__github__get_me`, `mcp__github__list_issues`), which grows and changes independently of this repo. A stale allowlist would start failing CI on legitimate new tools, which is a worse failure mode than the shape-only check it replaces. The shape check still does its actual job: catching a group-level `if` or a non-permission-rule value (env expressions, missing parens, etc.), which is what issue #1039 was about. Leaving the tool-name-typo case uncaught rather than trade it for a maintenance-burden false-failure risk.

---

nyldn:
Re the latest Octopus review (5 findings) — going through each:

**Two [NORMAL] findings claiming a regression (freeze/webhook hooks now fire unconditionally):** verified against the code, these are false positives. Both removed keys were the invalid *matcher-group*-level `if` (the whole subject of #1039) — Claude Code already silently ignored them before this PR, which is the bug being fixed, not a behavior this PR removes. Both scripts fully self-gate independent of that key:
- `hooks/telemetry-webhook.sh:19-25` exits 0 immediately if `OCTOPUS_WEBHOOK_URL` is unset — no network call.
- `hooks/freeze-check.sh:21` and `:32-38` exit 0 immediately unless `/octo:freeze` has written its session state file, independent of `OCTO_FREEZE_MODE`.

Runtime behavior is identical before and after this PR; only the dead, misleading JSON keys (and the startup warning they caused) are gone. This is also explained in the original PR description.

**[NIT] hook-level `if` regex assumes single `Tool(pattern)` form:** skipping — the finding itself flags "low confidence, no local evidence" for compound `if` syntax, and I found none either. Not worth speculatively loosening the check.

**[NIT] `test_hooks_json_if_schema` doesn't verify removed conditions were re-homed:** moot given the above — there's no equivalent hook-level condition to migrate to, since the group-level conditions were never functional.

**[PRE-EXISTING] `test_hooks_json_matchers` never iterates hook blocks (same `data.items()` bug I already found and fixed in my own new test):** confirmed real, and I attempted the same one-line fix here. It exposed a bigger issue: most hook event types (`SessionStart`, `UserPromptSubmit`, `PreCompact`, etc.) legitimately have no `matcher` — only tool-scoped events (`PreToolUse`/`PostToolUse`) use one. Fixing the traversal bug naively makes the check "every block needs a matcher," which is false and would break CI on valid config. Reverted; scoping that check correctly (matcher required only for tool-scoped event types) is real work deserving its own PR, not a drive-by inside this one.

---

nyldn:
Re CodeRabbit's latest 2 findings: fixed the `block.get("hooks", [])` non-list crash (same class as the earlier container-type gaps — pushed in 97171d1a). Skipping the `"if": null` vs. absent-key distinction: both cases mean "no condition to enforce" for what this check actually validates (that a present `if` is a valid permission-rule string), so treating them identically doesn't lose any real coverage. `hooks.json` is maintainer-edited config, not adversarial input, and I'd rather stop adding type-safety layers to a smoke test than keep chasing edge cases with no practical difference in outcome.

---

nyldn:
Superseded by #1037, which is now merged. Both PRs remove the same two ignored matcher-level conditions. The merged fix preserves all 42 hook handlers and five valid handler-level filters, with regression coverage for unsupported matcher-group fields.

This PR's additional validator has an unresolved null-value case, so it was not included. Closing the duplicate without merging it.

PR #1046, opened 2026-09-18
fix(tangle): normalize structured decomposition markdown before fallback

Jhacarreiro:
Tangle currently treats provider decomposition output as unusable unless every subtask already matches its compact one-line wire format. A provider can return a semantically complete structured Markdown decomposition (for example, `### N. [CODING]` headings with multiline `Reads:`, `Files:`, `Creates:` and task prose) and still be rejected as semantic-invalid, causing unnecessary provider fallback.

This PR adds a deterministic local normalizer before provider fallback:
- accepts equivalent structured Markdown headings for CODING/REASONING subtasks
- extracts concrete `Reads:`/`Files:`/`Creates:` paths without inventing write authority
- derives `Task:` prose from known action sections while excluding acceptance/input/output/blocking metadata
- keeps valid wire-format output unchanged
- fails closed for CODING items without concrete `Files:`/`Creates:` scopes
- materializes normalized wire format only after the existing wire validator accepts it

The fallback chain therefore stops at the first provider when local syntax repair succeeds; semantic/scope adequacy, contextual-read validation, overlap handling and adaptive/strict write policy still run afterwards.

Validation:
- new normalization suite: 9/9
- write-scope safety: 34/34
- contextual reads: 40/40
- post-execution scope contract: 16/16
- syntax smoke: 5/5
- decomposition adequacy: 14 consecutive cases passed before the bounded 40s diagnostic timeout, with no failure observed
- the normalizer was also exercised against a real rich-Markdown decomposition shape and produced 5 parseable subtasks / 4 coding subtasks that pass the downstream parallel-scope validator when evaluated from the target repository context.

Public diff contains no deployment-specific paths, credentials, runtime IDs, or private project identifiers.

## Summary by CodeRabbit

- **New Features**
  - Added deterministic conversion of provider-generated decomposition text into a compact, validated task format.
  - Preserves valid wire-format output and recognized read, write, and create scopes.
  - Rejects incomplete coding tasks without concrete file scope and continues fallback handling when normalization fails.
  - Added documentation describing decomposition normalization behavior.

- **Tests**
  - Added coverage for normalization, validation, fallback handling, scope preservation, and valid-format pass-through.


Jhacarreiro:
ClawSweeper 🐠 reef update

Thanks for the contribution here. ClawSweeper was able to repair this branch in place, which keeps the original PR as the clean canonical path.

Validation: bash tests/unit/test-tangle-decomposition-normalization.sh; bash tests/unit/test-configurable-fallback-chains.sh; bash -n scripts/lib/workflows.sh tests/unit/test-tangle-decomposition-normalization.sh; git diff --check; make sync-check
This keeps attribution in the original current: commits, review context, and changelog notes all stay visible.

fish notes: reasoning high; reviewed against 6ac93957135f.

PR #1049, opened 2026-09-18
fix(doctor): resolve PLUGIN_DIR and WORKSPACE_DIR defaults

### LINKED ISSUE #1048: doctor reports 161 false failures: PLUGIN_DIR is never set in scripts/doctor.sh (assigned but not exported in orchestrate.sh, then exec'd)
### Plugin Version

11.5.0 (code identical at `main` HEAD as of 2026-09-18)

### Claude Code Version

2.1.277

### What happened?

A full `doctor` run reports **161 failures** that are all false. The real total is 0.

```
Summary: 32 passed, 12 warning(s), 161 failure(s)
```

Breakdown by category (`doctor --json`):

| category | false failures |
|---|---|
| skills | 117 (63 skills + 54 commands) |
| hooks  | 42 |
| state  | 1 |
| agents | 1 |

The messages point at paths that begin at the filesystem root, which is the tell:

```
✗ Command file missing: plan.md
  /commands/plan.md
✗ Skill file missing: octopus-code-review
  /skills/octopus-code-review
✗ agents/config.yaml not found
  Expected at: /agents/config.yaml
✗ Workspace directory missing
```

**Cause:** `scripts/doctor.sh` never sets `PLUGIN_DIR`, and nothing exports it into the process.

`scripts/orchestrate.sh:46` computes it:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
PLUGIN_DIR="$(dirname "$SCRIPT_DIR")"
```

but it is assigned, not exported. The early-dispatch block then hands `doctor` to a **fresh process** (`scripts/orchestrate.sh:62`):

```bash
doctor) exec bash "${SCRIPT_DIR}/doctor.sh" "${_octo_early_tail[@]}" ;;
```

and `scripts/doctor.sh` is only:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
source "${SCRIPT_DIR}/lib/doctor.sh"
do_doctor "$@"
```

So `PLUGIN_DIR` is empty in every consumer in `lib/doctor.sh`, and each check resolves `"${PLUGIN_DIR}/commands/plan.md"` to `/commands/plan.md`. Same story for `WORKSPACE_DIR` at `lib/doctor.sh:855` (the one `state` failure) — `lib/doctor.sh` defaults it to `${HOME}/.claude-octopus` at lines 148 and 634, but not at 855.

Setting `CLAUDE_PLUGIN_ROOT` does **not** help — no code path derives `PLUGIN_DIR` from it.

The `installation` category is unaffected, so the health pass `/octo:setup` runs (`doctor installation --json` + `cache-check --json`) stays clean and hides this. It only shows up on a full run.

### What did you expect?

`doctor` resolves its own plugin root and reports 0 failures on a healthy install.

### Steps to reproduce

```bash
OCTO_ROOT=~/.claude/plugins/cache/nyldn-plugins/octo/11.5.0
CLAUDE_PLUGIN_ROOT="$OCTO_ROOT" bash "$OCTO_ROOT/scripts/orchestrate.sh" doctor
# => Summary: 32 passed, 12 warning(s), 161 failure(s)
```

Same install, with the two variables provided:

```bash
PLUGIN_DIR="$OCTO_ROOT" WORKSPACE_DIR="$HOME/.claude-octopus" \
  bash "$OCTO_ROOT/scripts/doctor.sh"
# => Summary: 39 passed, 13 warning(s), 0 failure(s)
```

### Suggested fix

One line in `scripts/doctor.sh`, which fixes both the `orchestrate.sh` path and direct invocation:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
PLUGIN_DIR="${PLUGIN_DIR:-$(dirname "$SCRIPT_DIR")}"   # add
source "${SCRIPT_DIR}/lib/doctor.sh"
```

Exporting `PLUGIN_DIR` in `orchestrate.sh` would fix the dispatched path only, and `doctor.sh` is runnable on its own. `lib/doctor.sh:855` could also take the same `${WORKSPACE_DIR:-${HOME}/.claude-octopus}` default already used at lines 148 and 634.

The other four early-dispatch targets (`capabilities.sh`, `cache-check.sh`, `security-audit.sh`, `repair.sh`) are `exec`'d the same way and may want the same guard; I only verified `doctor.sh`.

### Footnote

`.github/ISSUE_TEMPLATE/bug_report.yml` asks reporters to "Run `/octo:doctor`" for the version, but no such command ships — `commands/` has 54 files and no `doctor.md`, and the manifest lists none. Doctor is script-only. The template probably wants `/octo:setup` or `package.json`.

nyldn:
## Description

`doctor` reports 161 false failures on a healthy install. Root cause: `scripts/orchestrate.sh:46` computes `PLUGIN_DIR` but never exports it, and the early-dispatch block `exec`'s `doctor.sh` into a fresh process (`scripts/orchestrate.sh:62`), so every path-based check in `lib/doctor.sh` resolves `"${PLUGIN_DIR}/..."` against an empty variable, i.e. `/`. `doctor.sh` is also runnable standalone, where `PLUGIN_DIR` was never set at all. Separately, `lib/doctor.sh:855` (the workspace-writable check) reads `$WORKSPACE_DIR` with no default, unlike the same variable at `lib/doctor.sh:148` and `:634`, producing the one `state`-category false failure.

## Type of Change
- [x] Bug fix

## Fix

- `scripts/doctor.sh`: default `PLUGIN_DIR="${PLUGIN_DIR:-$(dirname "$SCRIPT_DIR")}"`, so both the `orchestrate.sh doctor` dispatch path and a standalone `bash scripts/doctor.sh` invocation resolve it.
- `scripts/lib/doctor.sh:855`: give the workspace-writable check the same `"${WORKSPACE_DIR:-${HOME}/.claude-octopus}"` default already used elsewhere in the file.

Not in scope: the issue also flags `capabilities.sh`, `cache-check.sh`, `security-audit.sh`, and `repair.sh` as other `exec`-dispatched targets that *may* have the same gap, but the reporter only verified `doctor.sh`. Of those, `capabilities.sh`, `cache-check.sh`, and `repair.sh` do source `lib/lifecycle.sh`, which reads `${PLUGIN_DIR:-}` — worth a follow-up, but left out here to keep this fix to the confirmed, reproduced defect.

## Testing

Reproduced the reported repro against a copy of this tree with `PLUGIN_DIR`/`WORKSPACE_DIR` unset and `CLAUDE_PLUGIN_ROOT` set:

```
# before: Summary: 32 passed, 12 warning(s), 161 failure(s)
# after:  Summary: 26 passed, 7 warning(s), 3 failure(s)
```

The remaining 3 failures (no provider authenticated, workspace directory missing, probe cache not writable) are genuine for that throwaway sandbox — not path-resolution bugs.

```bash
bash -n scripts/*.sh scripts/lib/*.sh   # all pass
bash tests/unit/test-doctor-v10.sh              # 30/30
bash tests/unit/test-windows-doctor-compat.sh   # 5/5
bash tests/unit/test-skill-doctor-stable-link.sh # 6/6
```

Did not run the full `make test-unit`/`make test-integration` matrix (332 unit suites) — disproportionate for a two-line default-value fix isolated to `doctor.sh`/`lib/doctor.sh`, and none of the other suites touch these code paths.

## Related Issues
Closes #1048

https://claude.ai/code/session_013anKgNUpwgXNx9s27jG8xp

---
_Generated by [Claude Code](https://claude.ai/code/session_013anKgNUpwgXNx9s27jG8xp)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace checks now use configured plugin data, custom workspace, or the default workspace location in the correct order.
  * Workspace validation correctly detects stale results and reports write access for plugin data storage.
  * Doctor diagnostics locate the plugin directory from the script location when no override is provided.

* **Configuration**
  * Added support for overriding the plugin directory with `PLUGIN_DIR`.
  * Workspace paths beginning with `~` are expanded to the home directory.


nyldn:
Pushed 9504a4d addressing the Claude Octopus Code Review findings:

- **Stale-results check still used raw `WORKSPACE_DIR`** — fixed. Hoisted `workspace_dir` to the top of `doctor_check_state()` and reused it for both the stale-results check (line 841) and workspace-writable check.
- **Workspace fallback ignored `CLAUDE_PLUGIN_DATA` precedence** — fixed. The default is now `${CLAUDE_PLUGIN_DATA:-${WORKSPACE_DIR:-${HOME}/.claude-octopus}}`, matching the precedence already used at `lib/doctor.sh:634`.
- **Local fallback diverges from canonical resolution** (nit) — partially addressed: matched the file's own existing `CLAUDE_PLUGIN_DATA`-first pattern rather than introducing a new one. Didn't switch to `state-manager.sh`'s `resolve_octopus_workspace()` — that file runs setup side effects at source time, so pulling it into doctor.sh's lightweight check path is a bigger change than this fix's scope.
- **No regression test** (nit) — fixed. Added a test that fails against the pre-fix code (unbound `WORKSPACE_DIR` under `set -u`) and passes after.

---

nyldn:
The "Unit Tests" and "Integration Tests" failures reported on `1b40749d` (the first commit) aren't real test failures — both jobs' logs show `Unit matrix result: cancelled` / `Integration Tests did not run: blocked by Unit Tests (failure), not an integration failure`. That run (35386506411) was superseded and cancelled when I pushed the second commit (`9504a4d1`) addressing the review findings, which is expected GitHub Actions concurrency behavior, not a defect in this PR. CI for the current head (`9504a4d1`) is what's authoritative; watching it now.

---

nyldn:
Pushed 2e7b9db addressing the second review round:

- **Workspace resolution dropped `CLAUDE_OCTOPUS_WORKSPACE`** — fixed. Precedence is now `CLAUDE_PLUGIN_DATA` > `CLAUDE_OCTOPUS_WORKSPACE` > `WORKSPACE_DIR` > `${HOME}/.claude-octopus`, matching `resolve_octopus_workspace()`.
- **New test didn't cover the real entry point (`scripts/doctor.sh`)** — fixed. Added a test that runs `bash scripts/doctor.sh --json config` as a subprocess with `PLUGIN_DIR` unset (the actual orchestrate.sh early-dispatch condition from #1048), and asserts `plugin-build`/`install-source` resolve correctly instead of falling back to the broken pre-fix output (`Build SHA unavailable`, empty `detail`).
- **No test for the `CLAUDE_OCTOPUS_WORKSPACE` tier** (nit) — fixed, alongside the precedence fix above.

Verified each new test independently fails against the specific pre-fix line it targets (isolated via `git show <parent-commit>:<file>` swaps) and passes after.

Not addressed: `lib/doctor.sh:634`'s pre-existing `_cb_dir` default has the same `CLAUDE_OCTOPUS_WORKSPACE` gap — untouched by this PR's diff and outside #1048's confirmed defect, so left for a separate fix.

---

nyldn:
Pushed 33542ee addressing CodeRabbit's changes-requested review:

- **🟠 Major: `doctor_check_v10_state_health` didn't use the resolved workspace** — fixed. It now accepts the already-resolved `workspace_dir` as an optional first argument (falling back to the same `CLAUDE_PLUGIN_DATA` > `CLAUDE_OCTOPUS_WORKSPACE` > `WORKSPACE_DIR` > default chain standalone), and `doctor_check_state` passes its resolved value in. Verified `probe-cache-writable` now resolves under `CLAUDE_PLUGIN_DATA` instead of falling back to `${HOME}/.claude-octopus`.
- **🟡 Test precedence with a competing workspace** — fixed. Both precedence tests now set `WORKSPACE_DIR` to a distinct, real directory (not just unset it) and assert the higher-precedence variable wins and the competing directory never appears in results.
- **🟡 Restore inherited environment variables** — fixed. Moved each test's env overrides and `doctor_check_state` call into a `$( ... )` subshell, so they can't leak into later cases (confirmed by direct check: vars are unset again immediately after the subshell exits).

Not addressed: the "Docstring Coverage" pre-merge check (50%, threshold 80%) — this codebase doesn't use docstrings for bash functions anywhere in `lib/doctor.sh`, so adding them here would be inconsistent with the file's own convention rather than a real gap.

Full suite: 33/33 passing.

---

nyldn:
Pushed b12ecd3 addressing the latest Claude Octopus Code Review:

- **"New workspace_dir resolution isn't propagated to `doctor_check_v10_state_health`"** — already fixed in 33542ee (this review ran against 2e7b9db0, the commit before that fix).
- **NIT: comment overstated parity with `resolve_octopus_workspace()`** — fixed. `resolve_octopus_workspace()` has three tiers and never references `WORKSPACE_DIR`; reworded the comment to say the `WORKSPACE_DIR` tier is doctor's own addition, not part of the canonical resolver's contract.

No behavior change in this push. Suite: 33/33.

---

nyldn:
Both findings from the latest review are explicitly marked `[PRE-EXISTING]` by the reviewer itself — `doctor_check_recurrence` (line ~1649) and the `_cb_dir` circuit-breaker resolution (line ~640) have the same `WORKSPACE_DIR`-only/`CLAUDE_OCTOPUS_WORKSPACE`-omitting gaps this PR fixed elsewhere, but in functions this diff doesn't touch. Consistent with what this PR has said since the first review round (the `_cb_dir` gap was flagged as "Perspective B" commentary two rounds ago and left out then too): keeping this fix scoped to the confirmed #1048 defect and the two functions (`doctor_check_state`, `doctor_check_v10_state_health`) this diff actually changes, rather than sweeping the whole file for the same pattern. Filing a follow-up issue to track both locations so they aren't lost.

---

nyldn:
Pushed 4e4c846 addressing CodeRabbit's fourth review round:

- **🟡 Normalize leading `~` in doctor workspace resolution** — fixed. Extracted the precedence chain into a shared `_doctor_resolve_workspace_dir()` helper that also expands a leading `~` in `CLAUDE_OCTOPUS_WORKSPACE` (matching `resolve_octopus_workspace()`'s handling), used by both `doctor_check_state` and `doctor_check_v10_state_health`'s no-argument fallback. Verified `CLAUDE_OCTOPUS_WORKSPACE=~/some-dir` now resolves to `${HOME}/some-dir`.
- **🟡 Test `errexit` trap** — fixed, and this one's worth flagging as a real catch: `doctor_env_output="$(...)"` followed by reading `$?` on the next line doesn't work under `set -euo pipefail` — a nonzero exit from a command substitution inside a plain assignment triggers `errexit` *immediately*, before the next line runs, so any `doctor.sh` run reporting a failure-category result (exit 1) would have silently aborted the rest of the test file. Reproduced the abort in isolation (`foo="$(bash -c "exit 1")"` under `set -e` exits before printing anything) and confirmed the `if`/`else` form survives it with the correct `$doctor_env_rc`.

Full suite: 33/33.

---

nyldn:
Pushed 36621f4 addressing the Claude Octopus Code Review round (the two earlier runs on this commit hit their own session/quota limits — "Do not merge based on this output" — and are unrelated to this PR's code; the retry succeeded and found real findings):

- **NIT: `_doctor_resolve_workspace_dir` omits the `HOME`-unset fallback** — fixed. `resolve_octopus_workspace()` computes `home_base="${HOME:-$PWD}"` first; the doctor helper used raw `${HOME}` for both the default path and tilde expansion, so with `HOME` unset it silently degraded to `/.claude-octopus` (filesystem root) instead of falling back to `$PWD`. Verified the fix resolves to `$PWD/.claude-octopus` with `HOME` unset.
- **NIT: precedence duplicated across two files** — acknowledged, not changed. Same tradeoff explained in earlier rounds: sourcing `state-manager.sh` into doctor's lightweight check path pulls in its source-time setup side effects, which is a bigger change than this fix's scope. The helper is now centralized within `doctor.sh` itself (one function, two call sites) with a comment explaining the intentional divergence (the extra `WORKSPACE_DIR` tier).
- **Both `[PRE-EXISTING]` findings** (`doctor_check_recurrence` raw `WORKSPACE_DIR`, `_cb_dir` circuit-breaker missing `CLAUDE_OCTOPUS_WORKSPACE`) — same two call sites already filed as #1050 after the previous review round; not duplicating that issue.

Full suite: 33/33. CI is green on `4e4c8469` (Smoke/Unit/Integration Tests, Portability Lint, Symlinked Path all passed) and mergeable state is clean.

---

nyldn:
Pushed 28e9fd3 addressing CodeRabbit's latest finding:

- **🟡 Derive the probe-cache path from `workspace` directly** — fixed. `doctor_check_v10_state_health` computed its resolved `$workspace` correctly, but the probe-cache-writable check still preferred `octo_probe_cache_dir()` (`scripts/lib/session.sh`) when that function was available, and it checks raw `WORKSPACE_DIR` before falling back to `resolve_octopus_workspace()`. So when `WORKSPACE_DIR` is set to something other than what `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE` resolves to, the probe-cache check could inspect a different directory than the run-record/PID checks in the same function. Dropped that lookup; `cache_dir` is now derived from `$workspace` directly, same as its siblings. Verified with `session.sh` sourced and a competing `WORKSPACE_DIR` set — `probe-cache-writable` now agrees with the rest of the function.

Full suite: 33/33.

CodeRabbit auto-paused further reviews on this branch (too many commits in quick succession) — happy to trigger `@coderabbitai review` again if useful, but CI (Smoke/Unit/Integration Tests, the three required checks) is green and mergeable state was clean before this push.

---

nyldn:
Pushed 998c3f8 addressing this review round. Five findings, one pushback:

- **[NORMAL] `_doctor_resolve_workspace_dir` adds a `WORKSPACE_DIR` tier `resolve_octopus_workspace()` doesn't have** — investigated, not changed. I don't think removing the tier is correct: `orchestrate.sh:297` always assigns `WORKSPACE_DIR="$(resolve_octopus_workspace)"`, so in every real orchestrate.sh-initiated invocation `WORKSPACE_DIR` already agrees with `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE` — there's no real path where they diverge in production. But the pre-existing test at `tests/unit/test-doctor-v10.sh` (the "state health reports writable cache..." case, not part of this diff) explicitly sets `WORKSPACE_DIR` alone and expects `doctor_check_v10_state_health`'s no-arg fallback to honor it — that's *existing*, established behavior this PR is building on, not something it introduced. Dropping the tier to match `resolve_octopus_workspace()` exactly would silently break that test. Keeping `WORKSPACE_DIR` as the lowest-priority tier (below the two documented overrides) preserves that existing contract while still fixing the actual bug (#1048's `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE` gap).
- **[NORMAL] Pre-existing test doesn't isolate `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE`** — fixed, and this is the sharp edge of the finding above: that same pre-existing test *would* silently probe the wrong directory if either var were set ambiently. Reproduced it (`CLAUDE_PLUGIN_DATA=/tmp/x` in the environment flips `probe-cache-writable` to `fail`) and fixed it by unsetting both before the call, matching the isolation pattern the two new tests already use.
- **[NIT] Resolver skips `orchestrate.sh`'s `validate_workspace_path()` gate** — acknowledged, not changed. Same tradeoff as the "don't source `state-manager.sh`" calls in earlier rounds: pulling in path validation duplicates more of that file's logic into doctor's lightweight check path. Impact is bounded as the finding itself notes — doctor is read-only and the env var is self-set by the same user/host that would set it for `orchestrate.sh` anyway.
- **Both `[PRE-EXISTING]`** findings (circuit-breaker `_cb_dir`, `doctor_check_recurrence`) — same two call sites already filed as #1050.

Full suite: 33/33.

---

nyldn:
On the `octo_probe_cache_dir()` finding: I looked into this rather than reverting, because it directly contradicts CodeRabbit's finding from the previous round (which asked me to remove that same delegation, and which I did in 28e9fd3).

Checked where `octo_probe_cache_dir()` is actually called for real caching: `scripts/lib/workflows.sh:769` and `scripts/lib/semantic-cache.sh` (multiple call sites) — all from within an already-running `orchestrate.sh` workflow dispatch (probe/tangle/etc). By that point, `orchestrate.sh:297` has already run `WORKSPACE_DIR="$(resolve_octopus_workspace)"`, so `WORKSPACE_DIR` is *always* derived from the exact same `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE` inputs `_doctor_resolve_workspace_dir()` reads. In every real invocation path, both resolvers land on the same directory — there's no supported way in this codebase to get `WORKSPACE_DIR` set to something independent of those two variables. The scenario the finding describes ("`WORKSPACE_DIR` explicitly set to a value different from `CLAUDE_PLUGIN_DATA`/`CLAUDE_OCTOPUS_WORKSPACE`") only arises in the synthetic test setup my own regression test constructs to prove precedence — not in production usage.

Separately, in the standalone `doctor.sh` early-dispatch path this PR is actually about (`orchestrate.sh:62`), `session.sh` is never sourced, so `octo_probe_cache_dir` doesn't even exist there — the fallback `${workspace}/.cache/probe-results` is what always ran in that path, before and after this change.

Keeping the current code (deriving `cache_dir` from `$workspace` directly): it's correct for every real call path, and reverting to the `octo_probe_cache_dir()` delegation would reopen the actual, demonstrated bug from the previous round (internal inconsistency between `probe-cache-writable` and its sibling checks in the same function under a competing `WORKSPACE_DIR`).

The duplicated-resolver NIT is the same tradeoff acknowledged in prior rounds.

---

